PR #145, opened 2020-06-07
Designing approach for generating report for scanners

preetkaran20:


preetkaran20:
As we are not going with this approach so closing this PR.

PR #153, opened 2020-07-13
Reporting Vulnerability

preetkaran20:


src/main/java/org/sasanlabs/internal/utility/annotations/VulnerabilityLevel.java
```diff
@@ -35,6 +35,33 @@
...
+     */
+    RequestParameterLocation location() default RequestParameterLocation.QUERY_PARAM;
+
+    /**
+     * This information can be used by Scanners to know the name of the key whose value is read by
+     * the endpoint/vulnerability level for performing the operation.
+     *
+     * @return name of the parameter which holds the value
+     */
+    String parameterName() default "";
```
preetkaran20, line 52: Assumption is only one parameter will correspond to one Level. Not sure if there is any issue. Will tackle if so in future.

PR #197, opened 2020-08-30
File Upload Vulnerability

preetkaran20:


src/main/resources/static/sasan.html
```diff
@@ -1,3 +1,6 @@
 <div id="content"> 
     UserInterface for this Vulnerability level is still in-progress. Please use API interface for learning/testing this vulnerability. 
+    <script>
```
preetkaran20, line 3: Need to remove this.

preetkaran20:
This is a WIP for File Upload Vulnerability

src/main/java/org/sasanlabs/service/vulnerability/fileupload/UnrestrictedFileUpload.java
```diff
@@ -10,31 +18,149 @@
...
 public class UnrestrictedFileUpload {
+    private final Path root;
 
-    /** https://bezkoder.com/spring-boot-file-upload/ READ it */
+    private static final String STATIC_FILE_LOCATION = "upload";
+    private static final String BASE_PATH = "static";
+    private static final Random RANDOM = new Random(new Date().getTime());
+
+    public UnrestrictedFileUpload() throws IOException, URISyntaxException {
+        URI uploadDirectoryURI =
```
hemantgs, line 45: Should we have file paths in `application.properties`
 and read from there ? 
Given `VulnerableAppRestController` is already a `@RestController`

src/main/java/org/sasanlabs/service/vulnerability/fileupload/UnrestrictedFileUpload.java
```diff
@@ -10,31 +18,149 @@
...
+                new URI(
+                        Thread.currentThread()
+                                        .getContextClassLoader()
+                                        .getResource(BASE_PATH)
+                                        .toURI()
+                                + FrameworkConstants.SLASH
+                                + STATIC_FILE_LOCATION);
+        root = Paths.get(uploadDirectoryURI);
+        if (!root.toFile().exists()) {
```
hemantgs, line 54: We could probably think of using something like
IOUtils/FileUtils to handle file operations , off-load some of the headaches

src/main/java/org/sasanlabs/service/vulnerability/fileupload/UnrestrictedFileUpload.java
preetkaran20: hmm makes sense. Sure will move this to application.properties

src/main/java/org/sasanlabs/service/vulnerability/fileupload/UnrestrictedFileUpload.java
preetkaran20: Actually i think Directory handling's best way is to use NIO Api's, FileUtils can help with file operation. Other part of URI handling, i am not sure what is the alternative to get the path to resources folder easily.

src/main/java/org/sasanlabs/service/vulnerability/fileupload/UnrestrictedFileUpload.java
preetkaran20: I was thinking again about it, i am not sure whether we need to add directory to application.properties as i think this  might not change in future and in case if we think it changes or being used at multiple places we can either move to a common constant or move to application.properties. WDYT ?

hemantgs: Hmm , Let's say someone want to host this in a cloud server and the feel like they need to test this with a huge file of their own and they mount a volume that holds the file , in that case it might be easier for them to just change the path to the file in their own overidden app.properties , else they might have to tinker with the code again.
What do you think

src/main/java/org/sasanlabs/service/vulnerability/fileupload/UnrestrictedFileUpload.java
hemantgs: Yeah you're right , this should be fine :+1:

src/main/java/org/sasanlabs/service/vulnerability/fileupload/UnrestrictedFileUpload.java
preetkaran20: great example however the purpose of this vulnerability level is that user is only able to upload images but he is able to do many things and as a vulnerability it might not has that functionality of testing huge files.

I completely agree with your point that say NFS is mounted to some folder and they can change that folder in application.properties but i think being a vulnerable app this might never be the case and this is the vulnerability if user is able to upload for very big sizes as we want to restrict it to just images.

hemantgs: Hmm I see , if we think that case is not going to come up , I think it should be fine 👍

preetkaran20:
Hi @hemantgs ,

Please review this PR.

thanks,
Karan

src/main/resources/static/templates/JWTVulnerability/keys/private_key.pem
```diff
@@ -1,28 +1,32 @@
+Bag Attributes
```
hemantgs, line 1: Is this key being used anywhere ?

src/main/resources/static/templates/JWTVulnerability/keys/public_crt.pem
```diff
@@ -0,0 +1,27 @@
+Bag Attributes
```
hemantgs, line 1: the cert too

hemantgs (commented):
apart from the two comments , others LGTM !

src/main/resources/static/templates/JWTVulnerability/keys/private_key.pem
preetkaran20: Yes they are used in JWT Vulnerability. Actually not directly but in case we want to test ZAP fuzzer we use these.

src/main/resources/static/templates/JWTVulnerability/keys/public_crt.pem
preetkaran20: same here.

preetkaran20:
Merging this as review being done by @hemantgs

src/main/resources/static/templates/JWTVulnerability/keys/public_crt.pem
hemantgs: hopefully the key and cert are not in active use anywhere :smile:

preetkaran20: haha yes yes they are not used anywhere else :D

PR #216, opened 2020-10-02
issues-211-change level to construction injection

fengyuanyang:
1. Using construction dependency injection instead of field injection
2. remove Autowired annotation from construction since it's not necessary in spring.


preetkaran20 (approved):
@fengyuanyang  thanks a lot for the change. LGTM.

thanks,
Karan

src/main/java/org/sasanlabs/controller/VulnerableAppRestController.java
```diff
@@ -27,7 +26,6 @@
 
     private int port;
 
-    @Autowired
```
preetkaran20, line 30: Great !!!

Issue #212, opened 2020-10-02
Adding Unit Test for Vulnerabilities


preetkaran20:
**Is your feature request related to a problem? Please describe.**
Currently we are not having any Unit tests and hence regression is quite tough to handle. This task is an Uber level task to add unit tests and it will be divided into multiple parts to handle it.

**Describe the solution you'd like**
Adding unit tests.

preetkaran20:
Unit test case for even one method is welcomed. 

thanks,
Karan

AlessandroArosio:
I can have a look during this weekend

fengyuanyang:
I am working on Commoninjection unit tests.
There are some issues i need to conquer. 
I will send pr later

preetkaran20:
Unit test case for command injection is done. PR https://github.com/SasanLabs/VulnerableApp/pull/221

preetkaran20:
@fengyuanyang please let me know if you want to work on other tests else we can unassign this task from you and let someone else pick other tests.

fengyuanyang:
Hi @preetkaran20 
You can unassign me , i am not available for now to do other tests.

PR #221, opened 2020-10-03
Command injection unit test

fengyuanyang:
#212 
unit test for commoninjection level1 and level2


src/test/java/org/sasanlabs/service/vulnerability/commandInjection/CommandInjectionTest.java
```diff
@@ -0,0 +1,76 @@
...
+        Assert.isTrue(
+                responseEntity
+                        .getBody()
+                        .getContent()
+                        .startsWith("PING 127.0.0.1 (127.0.0.1): 56 data bytes"),
+                "must get response");
+        Assert.isTrue(HttpStatus.OK == responseEntity.getStatusCode(), "must equal");
+        Assert.isTrue(responseEntity.getBody().getIsValid(), "must equal");
+    }
+
```
hemantgs, line 32: @fengyuanyang I doubt this test would run ,
So the `getVulnerablePayloadLevel1()` method is actually a `RequestMapping` and the `CommandInjection`  class is a `RestController` basically what that means is the class and the method together expose a REST endpoint , 
So the test should focus on checking whether this endpoint can be reached when it runs in a test server. So that would require using a `TestRunner` 
I would recommend giving https://spring.io/guides/gs/testing-web/ a read .

src/test/java/org/sasanlabs/service/vulnerability/commandInjection/CommandInjectionTest.java
```diff
@@ -0,0 +1,76 @@
...
+public class CommandInjectionTest {
+
+    CommandInjection commandInjection = new CommandInjection();
+    private String wrongUrl = "localhsost";
+
+    @Test
+    public void getVulnerablePayloadLevel1_Localhost_ExpectPingSuccess() throws IOException {
+        ResponseEntity<GenericVulnerabilityResponseBean<String>> responseEntity =
+                commandInjection.getVulnerablePayloadLevel1("127.0.0.1");
+        Assert.isTrue(
```
hemantgs, line 23: We should try to use AssertJ assertions , they provide a fluent API for checks
Refer https://assertj.github.io/doc/

src/test/java/org/sasanlabs/service/vulnerability/commandInjection/CommandInjectionTest.java
```diff
@@ -0,0 +1,76 @@
...
+                        .startsWith("PING 127.0.0.1 (127.0.0.1): 56 data bytes"),
+                "must get response");
+        Assert.isTrue(HttpStatus.OK == responseEntity.getStatusCode(), "must equal");
+        Assert.isTrue(responseEntity.getBody().getIsValid(), "must equal");
+    }
+
+    @Test
+    public void getVulnerablePayloadLevel1_WrongHost_ExpectPingSuccess() throws IOException {
+        ResponseEntity<GenericVulnerabilityResponseBean<String>> responseEntity =
+                commandInjection.getVulnerablePayloadLevel1(wrongUrl);
```
hemantgs, line 36: Same as above

src/test/java/org/sasanlabs/service/vulnerability/commandInjection/CommandInjectionTest.java
```diff
@@ -0,0 +1,76 @@
...
+        Assert.isTrue(HttpStatus.OK == responseEntity.getStatusCode(), "must equal");
+        Assert.isTrue(responseEntity.getBody().getIsValid(), "must equal");
+    }
+
+    @Test
+    public void getVulnerablePayloadLevel2_WrongHost_Return_OKAnd()
+            throws ServiceApplicationException, IOException {
+        RequestEntity entity = new RequestEntity(HttpMethod.GET, URI.create("localhost"));
+        ResponseEntity<GenericVulnerabilityResponseBean<String>> responseEntity =
+                commandInjection.getVulnerablePayloadLevel2("localhost", entity);
```
hemantgs, line 50: Same as above

src/test/java/org/sasanlabs/service/vulnerability/commandInjection/CommandInjectionTest.java
```diff
@@ -0,0 +1,76 @@
...
+        Assert.isTrue(HttpStatus.OK == responseEntity.getStatusCode(), "must equal");
+        Assert.isTrue(responseEntity.getBody().getIsValid(), "must equal");
+    }
+
+    @Test
+    public void getVulnerablePayloadLevel2_WrongHost_Return_NotResolve()
+            throws ServiceApplicationException, IOException {
+        RequestEntity entity = new RequestEntity(HttpMethod.GET, URI.create(wrongUrl));
+        ResponseEntity<GenericVulnerabilityResponseBean<String>> responseEntity =
+                commandInjection.getVulnerablePayloadLevel2(wrongUrl, entity);
```
hemantgs, line 67: Same as above

hemantgs (changes requested):
@fengyuanyang Thanks for taking the time to open this PR ,please have a look at the comments.

src/test/java/org/sasanlabs/service/vulnerability/commandInjection/CommandInjectionTest.java
preetkaran20: I think we need not to spawn a server as per my understanding lets only stick to business logic testing and mock all these OS calls.

build.gradle
```diff
@@ -60,7 +60,9 @@ jib {
 dependencies {
     implementation group: 'org.springframework.boot', name: 'spring-boot-starter-web'
 
-    testImplementation group: 'junit', name: 'junit'
+    testImplementation('org.springframework.boot:spring-boot-starter-test') {
+        exclude group: 'org.junit.vintage', module: 'junit-vintage-engine'
```
preetkaran20, line 64: Not need to have a testing framework. Let's go with plane Mockito or what ever comes with Spring.

preetkaran20 (changes requested):
Thanks for the PR. Let's think why we need unit test ? For checking business logic and making sure after some change it doesn't cause regression. So in Command Injection class we are having logics in each method like if request contains %26 then allow or if OS is windows execute a different command.

There is no need to actually calling Ping functionality mock that part.

src/test/java/org/sasanlabs/service/vulnerability/commandInjection/CommandInjectionTest.java
hemantgs: IMHO RestController tests should be SprintBootTests only , their sole purpose would be to check if the method as an endpoint is reachable , 
other method calls like `getResponseFromPingCommand()` can be mocked for sanity.

fengyuanyang: I would like to mock all these OS calls

src/test/java/org/sasanlabs/service/vulnerability/commandInjection/CommandInjectionTest.java
```diff
@@ -0,0 +1,114 @@
...
+ */
+public class CommandInjectionTest {
+
+    private CommandInjection commandInjection;
+
+    private String localhost = "localhost";
+    Supplier<Boolean> trueCondition = () -> true;
+    Supplier<Boolean> falseCondition = () -> false;
+    String successReturnStr = "PING SUCCESS";
+
```
preetkaran20, line 34: can be made private static final.

build.gradle
fengyuanyang: that's a good point , aren't you going to do integration test for spring boot ?

src/main/java/org/sasanlabs/service/vulnerability/commandInjection/CommandInjection.java
```diff
@@ -35,7 +36,7 @@
     private static final Pattern IP_ADDRESS_PATTERN =
             Pattern.compile("\\b((25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)(\\.|$)){4}\\b");
 
-    private StringBuilder getResponseFromPingCommand(String ipAddress, Supplier<Boolean> predicate)
+    public StringBuilder getResponseFromPingCommand(String ipAddress, Supplier<Boolean> predicate)
```
preetkaran20, line 39: can be made package private.

fengyuanyang (commented):
sure, i will change to private static final later, having dinner now . ha

src/main/java/org/sasanlabs/service/vulnerability/commandInjection/CommandInjection.java
fengyuanyang: sure, i will change to private static final later, having dinner now . ha

build.gradle
preetkaran20: we will and at that time we might need to add some dependencies and we are not sure which library we should pick at that time. for now lets not worry about that.

src/main/java/org/sasanlabs/service/vulnerability/commandInjection/CommandInjection.java
```diff
@@ -204,20 +209,18 @@ private StringBuilder getResponseFromPingCommand(String ipAddress, Supplier<Bool
     public ResponseEntity<GenericVulnerabilityResponseBean<String>> getVulnerablePayloadLevel6(
             @RequestParam(IP_ADDRESS) String ipAddress)
             throws ServiceApplicationException, IOException {
-
+        Supplier<Boolean> condition =
+                getSupplier(
+                        StringUtils.isNotBlank(ipAddress)
+                                && (IP_ADDRESS_PATTERN.matcher(ipAddress).matches()
+                                        || ipAddress.contentEquals("localhost")));
```
preetkaran20, line 216: lets make localhost a constant in constants file (we already have one org.sasanlabs.vulnerability.utils.Constants.java) and make both the testcase and here a reference to it.

src/test/java/org/sasanlabs/service/vulnerability/commandInjection/CommandInjectionTest.java
```diff
@@ -0,0 +1,114 @@
...
+    @Test
+    public void getVulnerablePayloadLevel1_Localhost_ExpectPingSuccess() throws IOException {
+        // Act
+        ResponseEntity<GenericVulnerabilityResponseBean<String>> responseEntity =
+                commandInjection.getVulnerablePayloadLevel1(localhost);
+
+        // Assert
+        Assert.isTrue(
+                successReturnStr.equals(responseEntity.getBody().getContent()),
+                "must get response");
```
preetkaran20, line 60: message is not descriptive and can cause problems later. please add descriptive message.

src/test/java/org/sasanlabs/service/vulnerability/commandInjection/CommandInjectionTest.java
```diff
@@ -0,0 +1,114 @@
...
+        Assert.isTrue(HttpStatus.OK == responseEntity.getStatusCode(), "must equal");
+        Assert.isTrue(responseEntity.getBody().getIsValid(), "must equal");
```
preetkaran20, line 62: Same as above.

src/test/java/org/sasanlabs/service/vulnerability/commandInjection/CommandInjectionTest.java
preetkaran20: Assert.equals can also be used there.

src/test/java/org/sasanlabs/service/vulnerability/commandInjection/CommandInjectionTest.java
```diff
@@ -0,0 +1,114 @@
...
+
+        verify(commandInjection).getVulnerablePayloadLevel2(eq(localhost), eq(entity));
+        verify(commandInjection).getResponseFromPingCommand(eq(localhost), eq(trueCondition));
+    }
+
+    @Test
+    public void getVulnerablePayloadLevel2_WrongHost_Return_NotResolve()
```
preetkaran20, line 100: I think these test cases are quite similar and as business logic we should test is:
1. ``` Supplier<Boolean> condition =
                () ->
                        ipAddress != null
                                && !SEMICOLON_SPACE_LOGICAL_AND_PATTERN
                                        .matcher(requestEntity.getUrl().toString())
                                        .find();
 i.e. predicate condition
2. Same response and "\n".

So i think you can remove this method. and for each level you can add test cases for Predicate conditions.

src/test/java/org/sasanlabs/service/vulnerability/commandInjection/CommandInjectionTest.java
```diff
@@ -0,0 +1,114 @@
...
+        commandInjection = Mockito.spy(new CommandInjection());
+
+        StringBuilder stringBuilder = new StringBuilder().append(successReturnStr);
+
+        // Real localhost ping is not able to operate in CI, Mock instead
+        doReturn(stringBuilder)
+                .when(commandInjection)
+                .getResponseFromPingCommand(eq(localhost), any(Supplier.class));
+
+        doReturn(trueCondition).when(commandInjection).getSupplier(true);
```
preetkaran20, line 46: please correct the formatting. Extra lines are not required.

src/test/java/org/sasanlabs/service/vulnerability/commandInjection/CommandInjectionTest.java
```diff
@@ -0,0 +1,114 @@
...
+
+        doReturn(falseCondition).when(commandInjection).getSupplier(false);
```
preetkaran20, line 48: I think these lines are not required and i think nor is there any requirement for the getSupplier method. Please correct me if i am wrong.

src/main/java/org/sasanlabs/service/vulnerability/commandInjection/CommandInjection.java
```diff
@@ -204,20 +209,18 @@ private StringBuilder getResponseFromPingCommand(String ipAddress, Supplier<Bool
...
-                                                                        .matches()
-                                                                || ipAddress.contentEquals(
-                                                                        "localhost")))
-                                .toString(),
-                        true),
+                        this.getResponseFromPingCommand(ipAddress, condition).toString(), true),
                 HttpStatus.OK);
     }
+
+    public Supplier<Boolean> getSupplier(boolean condition) {
```
preetkaran20, line 223: I think this is not required.

src/test/java/org/sasanlabs/service/vulnerability/commandInjection/CommandInjectionTest.java
```diff
@@ -0,0 +1,114 @@
+package org.sasanlabs.service.vulnerability.commandInjection;
+
+import static org.mockito.ArgumentMatchers.*;
```
preetkaran20, line 3: How many classes/constants are imported ? can we get rid of ".*"

build.gradle
```diff
@@ -60,7 +60,10 @@ jib {
 dependencies {
     implementation group: 'org.springframework.boot', name: 'spring-boot-starter-web'
 
-    testImplementation group: 'junit', name: 'junit'
+    testCompile "org.mockito:mockito-core:3.+"
+
```
preetkaran20, line 64: add junit5 comment.

src/test/java/org/sasanlabs/service/vulnerability/commandInjection/CommandInjectionTest.java
fengyuanyang: That's a great idea, assertJ did provides better checks

src/main/java/org/sasanlabs/service/vulnerability/commandInjection/CommandInjection.java
fengyuanyang: oh , i change to public here because i need to mock this method, in order to verify this method is called with Supplier (false or true ) , otherwise there is no way to verify getResponseFromPingCommand is called with expected predicate value.

src/test/java/org/sasanlabs/service/vulnerability/commandInjection/CommandInjectionTest.java
fengyuanyang: localhost has been moved to Constants

src/main/java/org/sasanlabs/service/vulnerability/commandInjection/CommandInjection.java
preetkaran20: Got that point but i think making package-private/Default can help achieve the same and also restrict access outside the package.

fengyuanyang: default works

src/main/java/org/sasanlabs/service/vulnerability/commandInjection/CommandInjection.java
```diff
@@ -204,20 +209,22 @@ private StringBuilder getResponseFromPingCommand(String ipAddress, Supplier<Bool
...
-                        true),
+                        this.getResponseFromPingCommand(ipAddress, condition).toString(), true),
                 HttpStatus.OK);
     }
+
+    /* In order to test getResponseFromPingCommand in unit test
+     * and verify with mockito to check condition is true or false
+     * getSupplier is extracted to an independent method
+     */
+    Supplier<Boolean> getSupplier(boolean condition) {
```
preetkaran20, line 227: I think this is not required. While checking the condition in mockito you need not to compare the 2 suppliers instead you can check condition.get() to be true or false.

src/test/java/org/sasanlabs/service/vulnerability/commandInjection/CommandInjectionTest.java
```diff
@@ -0,0 +1,125 @@
...
+        ResponseEntity<GenericVulnerabilityResponseBean<String>> responseEntity =
+                commandInjection.getVulnerablePayloadLevel1(Strings.EMPTY);
+
+        // Assert
+        assertThat(responseEntity.getBody().getContent()).isEqualTo(Strings.EMPTY);
+        assertThat(responseEntity.getStatusCode()).isEqualTo(HttpStatus.OK);
+        assertThat(responseEntity.getBody().getIsValid()).isEqualTo(true);
+
+        verify(commandInjection).getVulnerablePayloadLevel1(eq(Strings.EMPTY));
+        verify(commandInjection).getResponseFromPingCommand(eq(Strings.EMPTY), eq(falseCondition));
```
preetkaran20, line 75:
```suggestion
        verify(commandInjection).getResponseFromPingCommand(eq(Strings.EMPTY), (right) -> !right.get());
```

I think this or similar to this should work. Please check once.

src/test/java/org/sasanlabs/service/vulnerability/commandInjection/CommandInjectionTest.java
```diff
@@ -0,0 +1,125 @@
...
+        ResponseEntity<GenericVulnerabilityResponseBean<String>> responseEntity =
+                commandInjection.getVulnerablePayloadLevel2(Strings.EMPTY, entity);
+
+        // Assert
+        assertThat(responseEntity.getBody().getContent()).isEqualTo(Strings.EMPTY);
+        assertThat(responseEntity.getStatusCode()).isEqualTo(HttpStatus.OK);
+        assertThat(responseEntity.getBody().getIsValid()).isEqualTo(true);
+
+        verify(commandInjection).getVulnerablePayloadLevel2(eq(Strings.EMPTY), eq(entity));
+        verify(commandInjection).getResponseFromPingCommand(eq(Strings.EMPTY), eq(falseCondition));
```
preetkaran20, line 106: Please club common code inside a method.

src/test/java/org/sasanlabs/service/vulnerability/commandInjection/CommandInjectionTest.java
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    @Test
+    public void getVulnerablePayloadLevel2_LocalhostAndSEMICOLON_ExpectNoResponseContent()
+            throws ServiceApplicationException, IOException {
+        RequestEntity entity =
+                new RequestEntity(HttpMethod.GET, URI.create("localhost;localhost2"));
```
preetkaran20, line 113: Please create a proper URI. for this you need to run this vulnerability and check what is being passed so that we can have a proper setup, similar to what we have when application runs.

src/main/java/org/sasanlabs/service/vulnerability/commandInjection/CommandInjection.java
```diff
@@ -93,13 +93,16 @@ StringBuilder getResponseFromPingCommand(String ipAddress, boolean predict)
...
+        Supplier<Boolean> condition =
+                () ->
+                        StringUtils.isNotBlank(ipAddress)
                                 && !SEMICOLON_SPACE_LOGICAL_AND_PATTERN
                                         .matcher(requestEntity.getUrl().toString())
                                         .find();
         return new ResponseEntity<GenericVulnerabilityResponseBean<String>>(
                 new GenericVulnerabilityResponseBean<String>(
-                        this.getResponseFromPingCommand(ipAddress, condition.get()).toString(), true),
+                        this.getResponseFromPingCommand(ipAddress, condition.get()).toString(),
```
preetkaran20, line 104: i am not sure if this change is required. just for "eq" matcher you are doing these changes. can you please write some argument matcher in test case ?

preetkaran20:
@hemantgs  please help in reviewing this PR.

thanks,
Karan

src/main/java/org/sasanlabs/service/vulnerability/commandInjection/CommandInjection.java
```diff
@@ -204,19 +209,15 @@ private StringBuilder getResponseFromPingCommand(String ipAddress, Supplier<Bool
     public ResponseEntity<GenericVulnerabilityResponseBean<String>> getVulnerablePayloadLevel6(
             @RequestParam(IP_ADDRESS) String ipAddress)
             throws ServiceApplicationException, IOException {
+        Supplier<Boolean> condition =
```
preetkaran20, line 212: can we name it as validator ?

src/main/java/org/sasanlabs/service/vulnerability/commandInjection/CommandInjection.java
```diff
@@ -35,11 +36,10 @@
     private static final Pattern IP_ADDRESS_PATTERN =
             Pattern.compile("\\b((25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)(\\.|$)){4}\\b");
 
-    private StringBuilder getResponseFromPingCommand(String ipAddress, Supplier<Boolean> predicate)
-            throws IOException {
+    StringBuilder getResponseFromPingCommand(String ipAddress, boolean predict) throws IOException {
         boolean isWindows = System.getProperty("os.name").toLowerCase().startsWith("windows");
         StringBuilder stringBuilder = new StringBuilder();
-        if (predicate.get()) {
+        if (predict) {
```
preetkaran20, line 42: can we rename it as "isValid" ?

src/main/java/org/sasanlabs/service/vulnerability/commandInjection/CommandInjection.java
preetkaran20: at other locations too.

src/test/java/org/sasanlabs/service/vulnerability/commandInjection/CommandInjectionTest.java
```diff
@@ -0,0 +1,119 @@
...
+import org.springframework.http.RequestEntity;
+import org.springframework.http.ResponseEntity;
+
+/*
+ * @author Feng-Yuan Yang coolesaltw@hotmail.com
+ */
+public class CommandInjectionTest {
+
+    private static CommandInjection commandInjection;
+    private static final String successReturnStr = "PING SUCCESS";
```
preetkaran20, line 28: can we rename it as mockedSuccessPingResponse or something like that?

src/test/java/org/sasanlabs/service/vulnerability/commandInjection/CommandInjectionTest.java
```diff
@@ -0,0 +1,119 @@
...
+import org.springframework.http.HttpStatus;
+import org.springframework.http.RequestEntity;
+import org.springframework.http.ResponseEntity;
+
+/*
+ * @author Feng-Yuan Yang coolesaltw@hotmail.com
+ */
+public class CommandInjectionTest {
+
+    private static CommandInjection commandInjection;
```
preetkaran20, line 27: No need for this variable to be static. Please remove static from this declaration.

src/test/java/org/sasanlabs/service/vulnerability/commandInjection/CommandInjectionTest.java
```diff
@@ -0,0 +1,119 @@
...
+    private static final String successReturnStr = "PING SUCCESS";
+
+    @BeforeEach
+    void setUp() throws IOException {
+        commandInjection = Mockito.spy(new CommandInjection());
+
+        StringBuilder stringBuilder = new StringBuilder().append(successReturnStr);
```
preetkaran20, line 34: rename it to something similar to mockedSuccessPingResponse

src/test/java/org/sasanlabs/service/vulnerability/commandInjection/CommandInjectionTest.java
```diff
@@ -0,0 +1,119 @@
...
+
+    @Test
+    public void getVulnerablePayloadLevel2_NoIp_ExpectNoResponseContent()
+            throws ServiceApplicationException, IOException {
+        RequestEntity entity = new RequestEntity(HttpMethod.GET, URI.create(Strings.EMPTY));
+        ResponseEntity<GenericVulnerabilityResponseBean<String>> responseEntity =
+                commandInjection.getVulnerablePayloadLevel2(Strings.EMPTY, entity);
+
+        // Assert
+        assertStatusAndValid(responseEntity);
```
preetkaran20, line 91: instead of this method i think what you can do is, pass the boolean flag to a private method and when boolean is false response is same across all false conditions and when it is true response is same across all true conditions.

src/test/java/org/sasanlabs/service/vulnerability/commandInjection/CommandInjectionTest.java
```diff
@@ -0,0 +1,119 @@
...
+                new RequestEntity(HttpMethod.GET, URI.create("http://localhost:8080/;"));
+        ResponseEntity<GenericVulnerabilityResponseBean<String>> responseEntity =
+                commandInjection.getVulnerablePayloadLevel2(LOCALHOST, entity);
+
+        // Assert
+        assertStatusAndValid(responseEntity);
+        assertThat(responseEntity.getBody().getContent()).isEqualTo(Strings.EMPTY);
+
+        verify(commandInjection).getVulnerablePayloadLevel2(eq(LOCALHOST), eq(entity));
+        verify(commandInjection).getResponseFromPingCommand(eq(LOCALHOST), eq(false));
```
preetkaran20, line 111: These 5 lines can be wrapped inside a method and based on condition as false, responseEntity.getBody().getContent() is empty and based on condition as true response is successStr.

preetkaran20: so that this common code can be used at all the tests.

preetkaran20 (changes requested):
Common code can be refactored. Add some more tests covering the condition.

src/test/java/org/sasanlabs/service/vulnerability/commandInjection/CommandInjectionTest.java
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    private void assertStatusAndValid(
```
hemantgs, line 114: Please rename to `assertStatusAndValidate`

hemantgs (commented):
@fengyuanyang @preetkaran20 
This class would be fully tested if there is another Test class , that does an integration test of the vulnerable endpoints.
It can be done as part of this PR or a separate PR also.
Other than that with few minor nits , LGTM

src/test/java/org/sasanlabs/service/vulnerability/commandInjection/CommandInjectionTest.java
```diff
@@ -0,0 +1,117 @@
...
+        assertStatusAndValidate(responseEntity, Strings.EMPTY, Strings.EMPTY);
+
+        verify(commandInjection).getVulnerablePayloadLevel2(eq(Strings.EMPTY), eq(entity));
+    }
+
+    @Test
+    public void getVulnerablePayloadLevel2_LocalhostAndSEMICOLON_ExpectNoResponseContent()
+            throws ServiceApplicationException, IOException {
+        RequestEntity entity =
+                new RequestEntity(HttpMethod.GET, URI.create("http://localhost:8080/;"));
```
preetkaran20, line 95: this seems like not a correct url. can you please start the application and try to inject a payload so that we can use the exact payload here.

preetkaran20 (approved):
LGTM, However wanted the correct URL containing the semicolon and also if possible please add more test cases which tests all the clauses in the validator.

preetkaran20:
Hi @fengyuanyang ,

Are you planning to add more test cases to this PR or shall we proceed with merging this PR ?

thanks,
Karan

fengyuanyang (commented):
> Hi @fengyuanyang ,
> 
> Are you planning to add more test cases to this PR or shall we proceed with merging this PR ?
> 
> thanks,
> Karan

I can add more test case later this weekend.

src/test/java/org/sasanlabs/service/vulnerability/commandInjection/CommandInjectionTest.java
```diff
@@ -0,0 +1,119 @@
...
+        // Assert
+        assertStatusAndValid(responseEntity);
+        assertThat(responseEntity.getBody().getContent()).isEqualTo(Strings.EMPTY);
+
+        verify(commandInjection).getVulnerablePayloadLevel2(eq(Strings.EMPTY), eq(entity));
+        verify(commandInjection).getResponseFromPingCommand(eq(Strings.EMPTY), eq(false));
+    }
+
+    @Test
+    public void getVulnerablePayloadLevel2_LocalhostAndSEMICOLON_ExpectNoResponseContent()
```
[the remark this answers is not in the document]
preetkaran20: so the validator is having other conditions i.e. there is a regex pattern ```[;& ]``` 
i was asking you to validate everything in this pattern using unit test case.

preetkaran20:
@fengyuanyang 
https://github.com/SasanLabs/VulnerableApp/pull/221#discussion_r499328749 for this query please start the vulnerableApp you will get the UI, go to command injection vulnerability and then give some input to input box, there will be one url fired to the backend, use  that url as that is a correct url for command injection.

src/test/java/org/sasanlabs/service/vulnerability/commandInjection/CommandInjectionTest.java
preetkaran20: please start the vulnerableApp you will get the UI, go to command injection vulnerability and then give some input to input box, there will be one url fired to the backend, use that url as that is a correct url for command injection.

src/test/java/org/sasanlabs/service/vulnerability/commandInjection/CommandInjectionTest.java
preetkaran20: Please ignore for now.

src/test/java/org/sasanlabs/service/vulnerability/commandInjection/CommandInjectionTest.java
```diff
@@ -0,0 +1,592 @@
...
+import org.springframework.http.RequestEntity;
+import org.springframework.http.ResponseEntity;
+
+/*
+ * @author Feng-Yuan Yang coolesaltw@hotmail.com
+ */
+public class CommandInjectionTest {
+
+    private CommandInjection commandInjection;
+    private static final String ipAddress = "10.29.28.27";
```
preetkaran20, line 29: I think the simplest way is to have 127.0.0.1 which will work for localhost as well as level 6.
can you please update https://github.com/SasanLabs/VulnerableApp/pull/221/files#diff-d7958c1340593be6c9fb8b7ec2d7935aR11 with 127.0.0.1 address.

src/test/java/org/sasanlabs/service/vulnerability/commandInjection/CommandInjectionTest.java
```diff
@@ -0,0 +1,592 @@
...
+        commandInjection = Mockito.spy(new CommandInjection());
+
+        StringBuilder stringBuilder = new StringBuilder().append(mockedSuccessPingResponse);
+
+        // Real localhost ping is not able to operate in CI, Mock instead
+        doReturn(stringBuilder)
+                .when(commandInjection)
+                .getResponseFromPingCommand(eq(LOCALHOST), eq(true));
+
+        doReturn(stringBuilder)
```
preetkaran20, line 43: if done the above mentioned changes this is not required.

src/test/java/org/sasanlabs/service/vulnerability/commandInjection/CommandInjectionTest.java
```diff
@@ -0,0 +1,592 @@
...
+        // Assert
+        assertStatusAndValidate(responseEntity, Strings.EMPTY, Strings.EMPTY);
+
+        verify(commandInjection).getVulnerablePayloadLevel2(eq(Strings.EMPTY), eq(entity));
+    }
+
+    @Test
+    public void getVulnerablePayloadLevel2_LocalhostAndSEMICOLON_ExpectNoResponseContent()
+            throws ServiceApplicationException, IOException {
+        RequestEntity entity =
```
preetkaran20, line 99: Great !!! thnks

src/test/java/org/sasanlabs/service/vulnerability/commandInjection/CommandInjectionTest.java
```diff
@@ -0,0 +1,592 @@
...
+                commandInjection.getVulnerablePayloadLevel2(LOCALHOST, entity);
+
+        // Assert
+        assertStatusAndValidate(responseEntity, LOCALHOST, Strings.EMPTY);
+
+        verify(commandInjection).getVulnerablePayloadLevel2(eq(LOCALHOST), eq(entity));
+    }
+
+    @Test
+    public void getVulnerablePayloadLevel2_LocalhostAnd20_ExpectPingSuccess()
```
preetkaran20, line 131: 20 can be replaced with `cat_command`

src/test/java/org/sasanlabs/service/vulnerability/commandInjection/CommandInjectionTest.java
```diff
@@ -0,0 +1,592 @@
...
+                commandInjection.getVulnerablePayloadLevel2(LOCALHOST, entity);
+
+        // Assert
+        assertStatusAndValidate(responseEntity, LOCALHOST, mockedSuccessPingResponse);
+
+        verify(commandInjection).getVulnerablePayloadLevel2(eq(LOCALHOST), eq(entity));
+    }
+
+    @Test
+    public void getVulnerablePayloadLevel2_LocalhostAnd26_ExpectPingSuccess()
```
preetkaran20, line 148: can be named `localhostAndUrlEncoded&`  or  `localhostAndUrlEncodedand` instead of `LocalhostAnd26`

src/test/java/org/sasanlabs/service/vulnerability/commandInjection/CommandInjectionTest.java
```diff
@@ -0,0 +1,592 @@
...
+    }
+
+    @Test
+    public void getVulnerablePayloadLevel2_LocalhostAnd3B_ExpectPingSuccess()
+            throws ServiceApplicationException, IOException {
+        RequestEntity entity =
+                new RequestEntity(
+                        HttpMethod.GET,
+                        URI.create(
+                                "http://localhost:9090/CommandInjection/LEVEL_2?ipaddress=localhost%3B"));
```
preetkaran20, line 171: similar to above names of the test cases can be changed.

src/test/java/org/sasanlabs/service/vulnerability/commandInjection/CommandInjectionTest.java
```diff
@@ -0,0 +1,592 @@
...
+    }
+
+    @Test
+    public void getVulnerablePayloadLevel3_LocalhostPlusAnd_ExpectNoResponseContent()
+            throws ServiceApplicationException, IOException {
+        RequestEntity entity =
+                new RequestEntity(
+                        HttpMethod.GET,
+                        URI.create(
+                                "http://localhost:9090/CommandInjection/LEVEL_3?ipaddress=localhost&"));
```
preetkaran20, line 235: you can write a method which returns various response entities, call that method get the response entity as same line is repeated multiple times.

preetkaran20 (commented):
LGTM just a few minor things. Thanks a lot for detailed test case.

src/test/java/org/sasanlabs/service/vulnerability/commandInjection/CommandInjectionTest.java
fengyuanyang: This mock and return is required, otherwise level 6 will get into do SH command

src/test/java/org/sasanlabs/service/vulnerability/commandInjection/CommandInjectionTest.java
```diff
@@ -26,7 +26,7 @@
 public class CommandInjectionTest {
 
     private CommandInjection commandInjection;
-    private static final String ipAddress = "10.29.28.27";
+    private static final String ipAddress = "127.0.0.1";
     private static final String mockedSuccessPingResponse = "PING SUCCESS";
```
preetkaran20: can you make both of these constant variable names as LOCALHOST_ADDRESS and MOCKED_SUCCESS_PING_RESPONSE as this is the convention of static final variables

src/test/java/org/sasanlabs/service/vulnerability/commandInjection/CommandInjectionTest.java
preetkaran20: aah got it because level 6 checks for both localhost and some ip address and you have added check for both.

thanks,
Karan

preetkaran20 (approved):

preetkaran20:
Hi @fengyuanyang ,

it looks good to me. Can we merge it now ?

thanks,
Karan

preetkaran20:
Thanks a lot @fengyuanyang for the changes.

fengyuanyang:
> Hi @fengyuanyang ,
> 
> it looks good to me. Can we merge it now ?
> 
> thanks,
> Karan

Sure, thanks for your help and suggestions:)

PR #229, opened 2020-10-11
Applying SpotlessApply

preetkaran20:
As @Nimanita has faced issues with Spotless, i am suspecting that may be it was failing because of version issues of spotless.
Updated the version.

thanks,
Karan


Nimanita:
Finally all checks have passed.Thank you for it

PR #248, opened 2020-11-11
Back button functionality

ShareACoke1:
Added back button functionality to vulnerableApp.js, some changes were made in index.html too.

### DIFF
--- a/src/main/resources/static/vulnerableApp.js
+++ b/src/main/resources/static/vulnerableApp.js
@@ -321,3 +321,24 @@ function _addingEventListenerToShowHideHelpButton(vulnerableAppEndPointData) {
     document.getElementById("aboutContainer").scrollIntoView(true);
   });
 })();
+
+
+document.getElementById("learnAndPracticeBtn").addEventListener("click", pushToVulnerabilities);
+let previousUrl;
+
+function pushToVulnerabilities() {
+  let b = document.getElementById("learnAndPracticeBtn");
+  previousUrl = './learnVulnerabilities';
+  let id = b.id;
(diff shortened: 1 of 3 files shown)


src/main/resources/static/vulnerableApp.js
```diff
@@ -321,3 +321,24 @@ function _addingEventListenerToShowHideHelpButton(vulnerableAppEndPointData) {
...
   });
 })();
+
+
+document.getElementById("learnAndPracticeBtn").addEventListener("click", pushToVulnerabilities);
+let previousUrl;
+
+function pushToVulnerabilities() {
+  let b = document.getElementById("learnAndPracticeBtn");
+  previousUrl = './learnVulnerabilities';
```
preetkaran20, line 331: we can have URL's with the element id's. WDYT ?

ShareACoke1: I think it's a good idea.

preetkaran20: also 
Hi @ShareACoke1 ,

I had looked at the code, actually if we can do the same with opening the URL directly in browser then it would be great. I was thinking that we can parse the url, find the id's and click on the elements with those id's. this way we can achieve the full flow.

WDYT ?
Also i was planning to look at the router logics in frameworks so that we can do the same.

thanks,
Karan

preetkaran20:
Hi @ShareACoke1 ,

Are you stuck at something ? please let me know and we can discuss in detail.

thanks,
Karan

AdeenIlyas:
## AI Review Summary

PR #248: Found 1 issues across 0 security and 1 complexity.

PR #278, opened 2021-04-26
Introduce 'SECURE' vulnerability variant handled both on API and UI

### LINKED ISSUE #258: Adding a way to include some indicator for FalsePositives
Currently we can add the vulnerabilities in the VulnerableApp but there is no way to have some false positives "Secure" implementations such that we can better judge the scanners.

**Additional context**
we have a level which is called "secure" but we might need something like "secure1" , "secure2" etc. so that those can be secure implementations and "/scanner" endpoint will return those and then in Owasp ZAP evaluation script we can include it.

Owasp ZAP Integration with VulnerableApp: https://github.com/zapbot/zap-mgmt-scripts/tree/master/vulnerableApp

nowakkamil:
This pull request closes #258.
This change is in line with: https://github.com/SasanLabs/VulnerableApp/issues/258#issuecomment-826163238.


src/main/java/org/sasanlabs/beans/LevelResponseBean.java
```diff
@@ -3,16 +3,24 @@
...
 import java.util.List;
+import java.util.function.ToIntFunction;
 import org.sasanlabs.internal.utility.LevelConstants;
+import org.sasanlabs.internal.utility.SecureConstants;
+import org.sasanlabs.internal.utility.VariantConstants;
 import org.sasanlabs.internal.utility.annotations.RequestParameterLocation;
 import org.springframework.web.bind.annotation.RequestMethod;
 
-/** @author KSASAN preetkaran20@gmail.com */
+/**
```
preetkaran20, line 13: can you please run `gradle spotlessApply` ?

src/main/java/org/sasanlabs/internal/utility/SecureConstants.java
```diff
@@ -0,0 +1,30 @@
+package org.sasanlabs.internal.utility;
+
+/**
+ * Represents the Level of the Vulnerability Type that is considered secure.
+ */
+public interface SecureConstants {
```
preetkaran20, line 6: As discussed, we don't need SecureConstants as we are relying on Level and Variant(Type of Secure/Unsecure)

src/main/java/org/sasanlabs/internal/utility/VariantConstants.java
```diff
@@ -0,0 +1,6 @@
+package org.sasanlabs.internal.utility;
+
+public class VariantConstants {
```
preetkaran20, line 3: can we create an Enum ? Also please add Author tag and description of Enum.

src/main/java/org/sasanlabs/internal/utility/annotations/VulnerableAppRequestMapping.java
```diff
@@ -22,6 +24,13 @@
     @AliasFor(annotation = RequestMapping.class)
     String value();
 
+    /**
+     * Specify whether the implementation can be considered secure, as in, non-exploitable.
+     *
+     * @return variant
+     */
+    String variant() default VariantConstants.UNSECURE;
```
preetkaran20, line 32: Please go ahead with Enum.

src/main/java/org/sasanlabs/service/vulnerability/xss/reflected/UrlParamBasedImgTagAttrInjection.java
```diff
@@ -204,7 +206,8 @@ public UrlParamBasedImgTagAttrInjection() {
             description =
                     "XSS_QUOTES_AND_WITH_HTML_ESCAPE_PLUS_FILTERING_ON_INPUT_SRC_ATTRIBUTE_IMG_TAG")
     @VulnerableAppRequestMapping(
-            value = LevelConstants.SECURE,
+            value = SecureConstants.SECURE_1,
```
preetkaran20, line 209: make value = LevelConstants.LEVEL_*

src/main/resources/static/vulnerableApp.js
```diff
@@ -70,75 +70,86 @@ function _callbackForInnerMasterOnClickEvent(
...
-      if (isFirst) {
-        column.click();
-        isFirst = false;
-      }
-      innerMaster.appendChild(column);
+function getVariant(detailedInformation) {
+  return detailedInformation["Variant"] === "SECURE";
+}
+
+function getSvgVariant(isSecure) {
```
preetkaran20, line 77: we can rename it to `_getSvgElementForVariant` or some other name which depicts that it is returning the javascript element. "_" is to depict that it is a private utility.

src/main/resources/static/vulnerableApp.js
```diff
@@ -70,75 +70,86 @@ function _callbackForInnerMasterOnClickEvent(
...
+  let svg = document.createElement("img");
+  svg.classList.add("vector");
+  svg.setAttribute("src", "vectors/" + (isSecure ? "secure" : "unsecure") + ".svg");
```
preetkaran20, line 80: can we go with if else ?

src/main/resources/static/vulnerableApp.js
```diff
@@ -70,75 +70,86 @@ function _callbackForInnerMasterOnClickEvent(
...
-          key,
-          vulnerabilitySelected
-        )
-      );
-      if (isFirst) {
-        column.click();
-        isFirst = false;
-      }
-      innerMaster.appendChild(column);
+function getVariant(detailedInformation) {
```
preetkaran20, line 73: start the name with _ to depict that it is a private utility method.

src/main/resources/static/vulnerableApp.js
```diff
@@ -70,75 +70,86 @@ function _callbackForInnerMasterOnClickEvent(
...
+    );
+    if (isFirst) {
+      column.click();
+      isFirst = false;
+    }
+    innerMaster.appendChild(column);
   }
 }
 
+function handleElementAutoSelection(vulnerableAppEndPointData, id = 0) {
```
preetkaran20, line 128: please add a description of method and also variable id.

src/main/resources/static/vulnerableApp.js
```diff
@@ -70,75 +70,86 @@ function _callbackForInnerMasterOnClickEvent(
...
+  if (!vulnerableAppEndPointData.length) {
+    return;
+  }
+
+  if (id) {
```
preetkaran20, line 133: as id is an integer so please add a proper condition like `id === 0`

preetkaran20 (changes requested):
Please add some screenshots/screen video depicting the change so that i can know what exactly this pr has changed in UI

src/main/java/org/sasanlabs/beans/LevelResponseBean.java
```diff
@@ -101,7 +117,12 @@ public void setAttackVectorResponseBeans(
 
     @Override
     public int compareTo(LevelResponseBean levelResponseBean) {
-        return LevelConstants.getOrdinal(this.level)
-                - LevelConstants.getOrdinal(levelResponseBean.getLevel());
+        ToIntFunction<String> variantOrdinal = level ->
```
preetkaran20, line 120: As we are not using SecureConstants so this change might not be required. please check.

src/main/java/org/sasanlabs/internal/utility/annotations/VulnerableAppRequestMapping.java
preetkaran20: can we name it as SecurityVariant or SecureVariant ?

src/main/java/org/sasanlabs/beans/LevelResponseBean.java
nowakkamil: Resolved in 422b8d3

src/main/resources/static/vulnerableApp.js
nowakkamil: Resolved in 422b8d3

src/main/resources/static/vulnerableApp.js
nowakkamil: Resolved in 422b8d3

src/main/resources/static/vulnerableApp.js
nowakkamil: Resolved in 422b8d3

src/main/resources/static/vulnerableApp.js
nowakkamil: Resolved in 422b8d3

src/main/resources/static/vulnerableApp.js
nowakkamil: Resolved in 422b8d3

src/main/java/org/sasanlabs/internal/utility/VariantConstants.java
nowakkamil: Unfortunately, it's not possible due to Java limitations. I tried the ```Enum``` approach initially, but the annotation requires providing compile-time constant value, so I had to abandon this idea. The current implementation is similar to what we already have for ```LevelConstants```.

src/main/java/org/sasanlabs/service/vulnerability/xss/reflected/UrlParamBasedImgTagAttrInjection.java
nowakkamil: Doing that results in a run-time exception thrown by having the same value in the controller scope. This behaviour prevents ambiguity. Introducing duplicates would mean that identical endpoint would have to be handled via different methods. Additonally, this is connected to https://github.com/SasanLabs/VulnerableApp/pull/278#discussion_r623483099

src/main/java/org/sasanlabs/internal/utility/SecureConstants.java
nowakkamil: Please take a look at https://github.com/SasanLabs/VulnerableApp/pull/278#discussion_r623483099 and https://github.com/SasanLabs/VulnerableApp/pull/278#discussion_r623490100

src/main/java/org/sasanlabs/internal/utility/annotations/VulnerableAppRequestMapping.java
nowakkamil: I've got a mild preference for the current approach as I believe this _security_ part might be inferred by looking at sibling and parent properties. However, if you feel like your approach could avoid confusion in the future, I'm happy to make an update. Please let me know your final decision.

src/main/java/org/sasanlabs/beans/LevelResponseBean.java
nowakkamil: Please take a look at https://github.com/SasanLabs/VulnerableApp/pull/278#discussion_r623483099 and https://github.com/SasanLabs/VulnerableApp/pull/278#discussion_r623490100

nowakkamil:
The screenshot as per @preetkaran20 [request](https://github.com/SasanLabs/VulnerableApp/pull/278#pullrequestreview-645315758)
image

src/main/java/org/sasanlabs/internal/utility/VariantConstants.java
preetkaran20: Enums are allowed in annotations and they are static constants only. Look at following line https://github.com/SasanLabs/VulnerableApp/blob/ed98eb0923893ea822761cca3e7d0969b0eb340d/src/main/java/org/sasanlabs/internal/utility/annotations/VulnerableAppRequestMapping.java#L48 for more information.

src/main/java/org/sasanlabs/internal/utility/annotations/VulnerableAppRequestMapping.java
preetkaran20: then lets go with variant only but add a javadoc explaining the use of this field

src/main/java/org/sasanlabs/service/vulnerability/xss/reflected/UrlParamBasedImgTagAttrInjection.java
preetkaran20: no no wait, * was the next level number i.e. `Level_7`. please try this and it will work.

src/main/java/org/sasanlabs/internal/utility/SecureConstants.java
preetkaran20: please have a look at https://github.com/SasanLabs/VulnerableApp/pull/278/files#r623531869

preetkaran20 (changes requested):
added few comments. Also as there are lots of UI changes, please do some testing such that it doesn't break any existing flows.

src/main/java/org/sasanlabs/internal/utility/VariantConstants.java
nowakkamil: I see your point. I tried with different Enum (containing String), but now the simplified approach works. Resolved in 7eca817

src/main/java/org/sasanlabs/service/vulnerability/xss/reflected/UrlParamBasedImgTagAttrInjection.java
nowakkamil: It will, but wasn't this endpoint supposed to be marked as _secure_?

src/main/java/org/sasanlabs/internal/utility/SecureConstants.java
nowakkamil: Resolved in 3211f26

src/main/java/org/sasanlabs/service/vulnerability/xss/reflected/UrlParamBasedImgTagAttrInjection.java
nowakkamil: As recently discussed, a single level will be unique in the controller scope, so I have changed it per your suggesion. Resolved in 3211f26

src/main/java/org/sasanlabs/beans/LevelResponseBean.java
nowakkamil: Resolved in 3211f26

src/main/java/org/sasanlabs/beans/LevelResponseBean.java
```diff
@@ -101,7 +115,13 @@ public void setAttackVectorResponseBeans(
 
     @Override
     public int compareTo(LevelResponseBean levelResponseBean) {
-        return LevelConstants.getOrdinal(this.level)
-                - LevelConstants.getOrdinal(levelResponseBean.getLevel());
+        ToIntFunction<String> variantOrdinal =
```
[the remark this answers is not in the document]
nowakkamil: Resolved in 3211f26

src/main/resources/static/vulnerableApp.js
```diff
@@ -70,75 +70,95 @@ function _callbackForInnerMasterOnClickEvent(
...
-          key,
-          vulnerabilitySelected
-        )
-      );
-      if (isFirst) {
-        column.click();
-        isFirst = false;
-      }
-      innerMaster.appendChild(column);
+function _getVariant(detailedInformation) {
```
preetkaran20, line 73: name can be `_isSecureVariant`

preetkaran20 (approved):

src/main/resources/static/vulnerableApp.js
preetkaran20: @nowakamil

nowakkamil: Resolved in 0c0d74c

nowakkamil:
@preetkaran20, please review the final version and possibly merge if everything seems sorted out from your perspective. I've added a tooltip as previously discussed and tested the UI. As a result, no issues have been found. The following screenshot depicts the latest version
image

preetkaran20:
> @preetkaran20, please review the final version and possibly merge if everything seems sorted out from your perspective. I've added a tooltip as previously discussed and tested the UI. As a result, no issues have been found. The following screenshot depicts the latest version
> image

Can we have image color as green and hover over level as black and tooltip text as white and background color as black. Also tooltip text looks a lot bigger, can we have it as "Secure Implementation"

nowakkamil:
I've updated the PR by making the tooltip font a little bit smaller and changing its text. Below are two tooltip variants:
1. Unsecure
image

2. Secure
image


Regarding making the secure level text black, I think it is more congruent to have the vector and text in the same green colour. It also looks better, in my opinion.

preetkaran20:
> I've updated the PR by making the tooltip font a little bit smaller and changing its text. Below are two tooltip variants:
> 
> 1. Unsecure
>    image
> 2. Secure
>    image
> 
> Regarding making the secure level text black, I think it is more congruent to have the vector and text in the same green colour. It also looks better, in my opinion.

Yeah makes sense.

preetkaran20:
Before merging, i will take the update from your fork for little testing. In the mean time please go ahead with the VulnerableApp-Nginx changes as discussed earlier over the call.

preetkaran20:
@nowakkamil I have verified the PR and it looks good. Merging it.

Thanks a lot for the contribution. Now we are left with the next task which is making the same changes in VulnerableApp-Nginx.

thanks and regards,
Karan

PR #281, opened 2021-05-16
Adding VulnerableApp as it is broken

preetkaran20:


preetkaran20:
This is in accordance with https://github.com/SasanLabs/VulnerableApp-facade/issues/19

Issue #286, opened 2021-05-25
Add Local File Inclusion Vulnerability


preetkaran20:
**Is your feature request related to a problem? Please describe.**
Currently, we have 2 levels for LFI vulnerability but because it is a very common vulnerability and has a lot of variants hence, it requires to be relooked. 

**Describe the solution you'd like**
Add levels to LFI Vulnerability, create a good use-case for this vulnerability by looking at other vulnerable applications like Juice shop, DVWA, etc.

preetkaran20:
Acceptance criteria:
1. Use-case finding
2. Backend API creation for Vulnerability.
3. UserInterface/Template creation for the Vulnerability.
4. Compatible with https://github.com/SasanLabs/VulnerableApp-facade schema design.

Ivan12273:
@preetkaran20 Hi, can I give it a try? I was reading the code and I saw how can I add a new vulnerability, but I'm not sure where those 2 levels of vulnerability are, I imagine that they are in "UnrestrictedFileUpload" but I don't know which ones.

preetkaran20:
Hi @Ivan12273,

Sorry for the confusion, actually RFI has 2 levels implemented but for LFI we don't have any vulnerability level implemented.
Vulnerabilities implemented are: https://github.com/SasanLabs/VulnerableApp/tree/master/src/main/java/org/sasanlabs/service/vulnerability

thanks,
Karan

Ivan12273:
@preetkaran20 Oh I see, thanks, by the way, I was testing the RFI vulnerabilities using the url param like this:

http://localhost:9090/VulnerableApp/?url=_my-param_

But I don't get any kind of response, I am doing something wrong?

preetkaran20:
Hi @Ivan12273 ,

Actually RFI is not fully implemented, however you can go to following URL:
http://localhost:9090/VulnerableApp/RemoteFileInclusion/LEVEL_1?url=https://www.google.com

The logic is: `http://localhost:9090/VulnerableApp/<VulnerabilityName>/<Level>`

thanks,
Karan

preetkaran20:
[...]

PR #293, opened 2021-07-01
Add Local File Inclusion Vulnerability

Ivan01925:
Hi!, I was working on the issue #286 I added the Local File Inclusion to the project:

image

It has two level:
1. Direct url param injection
2. Direct url param injection with null byte character 

image

It works similar to Remote File Inclusion, when the user find the folder with the password file it will return the following:

image

When the users clicks on the hyperlink they will go to the next level:

image

The users can see the secret data if they successfully performs the vulnerability:

image

I also made some fixes, these hyperlinks were sending you to broken links when clicking on them, so I added "/ VulnerableApp" to fix them.

image
image

And the GitHub icon wasn't on the project so I added it too:

image

Greetings, I hope these changes are ok :)


src/main/resources/static/index.html
```diff
@@ -86,8 +86,8 @@ <h2>How can Vulnerability Scanning Tools use VulnerableApp?</h2>
 				<br/>
 				<b>Following are the endpoints exposed: </b>
 				<ol>
-				<li><a href="/scanner">Scanner Endpoint</a></li>
-				<li><a href="/sitemap.xml">SiteMap Endpoint</a></li>
+				<li><a href="/VulnerableApp/scanner">Scanner Endpoint</a></li>
```
preetkaran20, line 89: Thanks a lot this fix.

src/main/resources/static/index.html
```diff
@@ -22,7 +22,7 @@
 		<!-- TODO Add active to mode when clicked-->
 		<div class="navbar navbar-item">
 			<div class="navbar-item-menu active" id="home">
-				<a href="/">Home</a>
+				<a href="/VulnerableApp">Home</a>
```
preetkaran20, line 25: Thanks a lot for the fix.

preetkaran20 (changes requested):
Hi @Ivan12273,

Thanks for the PR, however, I have few suggestions:

1. we should create a UI where users can give input and that input causes LFI. Something similar to what we have for Command Injection or other vulnerability levels. 
2. For creating a UI, we need a particular theme such that users related to real-world LFI vulnerabilities. For this, have a look at other vulnerable applications such as DVWA.
3. Please add more Vulnerability levels associated with Blacklist, wrongly implemented whitelisting, different protocol bypasses, and 1 or 2 secure implementations such that we can better judge a scanner.


thanks,
Karan

src/main/resources/static/index.html
```diff
@@ -31,7 +31,7 @@
 				Mode
 			</div>
 			<div class="navbar-item-menu">
-				<a href="https://github.com/SasanLabs/VulnerableApp" target="_blank"><img src="/images/GitHub-Mark-32px.png" /></a>
+				<a href="https://github.com/SasanLabs/VulnerableApp" target="_blank"><img src="images/GitHub-Mark-32px.png" /></a>
```
preetkaran20, line 34: Amazing

Ivan12273:
> Hi @Ivan12273,
> 
> Thanks for the PR, however, I have few suggestions:
> 
> 1. we should create a UI where users can give input and that input causes LFI. Something similar to what we have for Command Injection or other vulnerability levels.
> 2. For creating a UI, we need a particular theme such that users related to real-world LFI vulnerabilities. For this, have a look at other vulnerable applications such as DVWA.
> 3. Please add more Vulnerability levels associated with Blacklist, wrongly implemented whitelisting, different protocol bypasses, and 1 or 2 secure implementations such that we can better judge a scanner.
> 
> thanks,
> Karan

Hello @preetkaran20 then it would be fine if I add an input where the user write the url? 
I see that if I try to write an url param like this: _http://localhost:9090/VulnerableApp/?file=**something**_ the app sends me to the home view, so I was thinking on an input as the command injection one but longer where if the user types a path with a different file, the application shows the content below the input.

preetkaran20:
> > Hi @Ivan12273,
> > Thanks for the PR, however, I have few suggestions:
> > 
> > 1. we should create a UI where users can give input and that input causes LFI. Something similar to what we have for Command Injection or other vulnerability levels.
> > 2. For creating a UI, we need a particular theme such that users related to real-world LFI vulnerabilities. For this, have a look at other vulnerable applications such as DVWA.
> > 3. Please add more Vulnerability levels associated with Blacklist, wrongly implemented whitelisting, different protocol bypasses, and 1 or 2 secure implementations such that we can better judge a scanner.
> > 
> > thanks,
> > Karan
> 
> Hello @preetkaran20 then it would be fine if I add an input where the user write the url?
> I see that if I try to write an url param like this: _[http://localhost:9090/VulnerableApp?file=[something]](http://localhost:9090/VulnerableApp?file=%5Bsomething%5D)_ the app sends me to the home view, so I was thinking on an input as the command injection one but longer where if the user types a path with a different file, the application shows the content below the input.

Didn't get your question fully, however, I would suggest not take URL as input because that doesn't seem like a real-world use case, check some applications like DVWA, etc to see how are they taking the input.

We are fine with showing the output below the input as that is there with all the vulnerabilities.

Ivan12273:
Hello @preetkaran20 I did something similar as DVWA, this is how it works:

I added a button for verify the URL:
image

For the first level if the user writes a path like this http://localhost:9090/VulnerableApp/?file=secretFiles/passwords.txt this is going to happen: 
image

In other cases the app will show:
image

I was only able to add another blacklist vulnerability and a secure implementations, sorry, I hope that's okay.

Best regards!

preetkaran20:
> Hello @preetkaran20 I did something similar as DVWA, this is how it works:
> 
> I added a button for verify the URL:
> image
> 
> For the first level if the user writes a path like this http://localhost:9090/VulnerableApp/?file=secretFiles/passwords.txt this is going to happen:
> image
> 
> In other cases the app will show:
> image
> 
> I was only able to add another blacklist vulnerability and a secure implementations, sorry, I hope that's okay.
> 
> Best regards!

@Ivan12273,
 one question, Does the user need to add a query param for input? I think we should add an input box. Adding the input to query param of a single page application doesn't seems right.

Ivan12273:
> > Hello @preetkaran20 I did something similar as DVWA, this is how it works:
> > I added a button for verify the URL:
> > image
> > For the first level if the user writes a path like this http://localhost:9090/VulnerableApp/?file=secretFiles/passwords.txt this is going to happen:
> > image
> > In other cases the app will show:
> > image
> > I was only able to add another blacklist vulnerability and a secure implementations, sorry, I hope that's okay.
> > Best regards!
> 
> @Ivan12273,
> one question, Does the user need to add a query param for input? I think we should add an input box. Adding the input to query param of a single page application doesn't seems right.

Alright, I will change it for a input box.

Ivan12273:
@preetkaran20 Hi again, I changed it for the input box :) 

image

preetkaran20:
> @preetkaran20 Hi again, I changed it for the input box :)
> 
> image



> @preetkaran20 Hi again, I changed it for the input box :)
> 
> image

I don't think input as a URL to an application for LFI is right. It might be the case for RFI or SSRF. For LFI we just need file name only.

As we see in DVWA, a PHP file is included based on an input parameter. My suggestion is to think of a use case first. Please go through some use cases from multiple places like https://www.neuralegion.com/blog/local-file-inclusion-lfi/ or https://www.netsparker.com/blog/web-security/local-file-inclusion-vulnerability/. 

An example use-case at the top of my mind is:
Say we want to show a page with images and the image name goes as an input to the application. Now if that image name is modified then that can cause LFI.

Code:
```
<img src="http://localhost:8080/LFI/?filename=image1.png" />
<img src="http://localhost:8080/LFI/?filename=image2.png" />
```

Now in this, user can change filename as "/etc/passwd" and that can cause LFI.

You are free to ignore my suggestion usecase.

src/main/java/org/sasanlabs/service/vulnerability/lfi/secretFiles/passwords.txt
```diff
@@ -0,0 +1,6 @@
+<h2>TOP SECRET PASSWORDS</h2>
+
```
preetkaran20, line 2: I think we don't need a special file, use /etc/passwd and equivalent for linux.

src/main/resources/static/templates/LocalFileInclusion/LEVEL_1/LFI.html
```diff
@@ -0,0 +1,20 @@
...
+    <div>
+        Local file inclusion (also known as LFI) is the process of including files,
+        that are already locally present on the server, through the exploiting of
+        vulnerable inclusion procedures implemented in the application.
+        This vulnerability occurs, for example, when a page receives, as input,
+        the path to the file that has to be included and this input is not properly
+        sanitized, allowing directory traversal characters (such as dot-dot-slash)
+        to be injected.
+        <br /> <br />
+        Try to find the password file inside the secretFiles folder using the URL param: <i>file</i>
```
preetkaran20, line 11: This line sounds like a challenge, which doesn't align with our vulnerable applications mission. Consider this application as a general application which has flaws and let scanners explore and find them.

src/main/java/org/sasanlabs/service/vulnerability/lfi/LFIVulnerability.java
```diff
@@ -0,0 +1,170 @@
...
+                        new File(
+                                "src/main/java/org/sasanlabs/service/vulnerability/lfi/"
+                                        + queryParameterURL);
+                Scanner reader = new Scanner(file);
+                String data = "";
+                while (reader.hasNextLine()) {
+                    data = data + reader.nextLine();
+                }
+                payload.append(data);
+                reader.close();
```
preetkaran20, line 56: Please close the reader in finally block or better use try with resourcr

src/main/java/org/sasanlabs/service/vulnerability/lfi/LFIVulnerability.java
```diff
@@ -0,0 +1,170 @@
...
+    }
+
+    @AttackVector(
+            vulnerabilityExposed = VulnerabilitySubType.LFI,
+            description = "LFI_URL_DIRECT_INJECTION_WITH_VALIDATION_ON_FILE")
+    @VulnerableAppRequestMapping(
+            value = LevelConstants.LEVEL_2,
+            descriptionLabel = "LFI_URL_PARAM_BASED_INJECTION_WITH_VALIDATION_ON_FILE",
+            htmlTemplate = "LEVEL_1/LFI")
+    public ResponseEntity<String> getVulnerablePayloadLevelUnsecureLevel2(
```
preetkaran20, line 72: Please try queryparam by name injection instead of map.

preetkaran20 (changes requested):

src/main/java/org/sasanlabs/service/vulnerability/lfi/LFIVulnerability.java
```diff
@@ -0,0 +1,170 @@
...
+            @RequestParam Map<String, String> queryParams) {
+        StringBuilder payload = new StringBuilder();
+        String queryParameterURL = queryParams.get(URL_PARAM_KEY);
+        if (queryParameterURL != null) {
+            try {
+                File file =
+                        new File(
+                                "src/main/java/org/sasanlabs/service/vulnerability/lfi/"
+                                        + queryParameterURL);
+                Scanner reader = new Scanner(file);
```
preetkaran20, line 50: Is scanner the best way to read file?

Ivan12273:
> > @preetkaran20 Hi again, I changed it for the input box :)
> > image
> 
> > @preetkaran20 Hi again, I changed it for the input box :)
> > image
> 
> I don't think input as a URL to an application for LFI is right. It might be the case for RFI or SSRF. For LFI we just need file name only.
> 
> As we see in DVWA, a PHP file is included based on an input parameter. My suggestion is to think of a use case first. Please go through some use cases from multiple places like https://www.neuralegion.com/blog/local-file-inclusion-lfi/ or https://www.netsparker.com/blog/web-security/local-file-inclusion-vulnerability/.
> 
> An example use-case at the top of my mind is:
> Say we want to show a page with images and the image name goes as an input to the application. Now if that image name is modified then that can cause LFI.
> 
> Code:
> 
> ```
> 
> 
> ```
> 
> Now in this, user can change filename as "/etc/passwd" and that can cause LFI.
> 
> You are free to ignore my suggested example usecase.
> 
> Earlier Input box suggestions was little wrong from my side as there is nothing much we can do with input box.

Alright!, then you suggest an input box only for the param, I will take this usecase, it is fine?

src/main/resources/static/templates/LocalFileInclusion/LEVEL_1/LFI.html
Ivan12273: Yes, I will change it.

src/main/java/org/sasanlabs/service/vulnerability/lfi/LFIVulnerability.java
Ivan12273: Alright!

src/main/java/org/sasanlabs/service/vulnerability/lfi/LFIVulnerability.java
Ivan12273: Not sure, do you have a suggestion?

preetkaran20:
> Alright!, then you suggest an input box only for the param, I will take this usecase, it is fine?

ok, can you please check if this use-case is feasible or not? You need to do a little more research on this.

src/main/java/org/sasanlabs/service/vulnerability/lfi/LFIVulnerability.java
preetkaran20: May be buffered reader but not sure if there are any new libraries which can be more better.

Ivan12273:
> > Alright!, then you suggest an input box only for the param, I will take this usecase, it is fine?
> 
> ok, can you please check if this use-case is feasible or not? You need to do a little more research on this.

Hi @preetkaran20 sorry for the delay, I did a research about this use-case, is feasible, it would be the same idea as this scenario but with an input for the images:

image

It occurred to me an use case where the user can change the file name of an image, for example: an user can select one image with a select input, once selected he can write in an input text the new name of the image, when the user press submit the app will show the image with a successful message. This will be the "normal" case.

In the other case, the user write a new name with a path like: "/etc/passwd", and when the user press submit the app will show the passwd info.

What do you think about this approach?

preetkaran20:
> > > Alright!, then you suggest an input box only for the param, I will take this usecase, it is fine?
> > 
> > 
> > ok, can you please check if this use-case is feasible or not? You need to do a little more research on this.
> 
> Hi @preetkaran20 sorry for the delay, I did a research about this use-case, is feasible, it would be the same idea as this scenario but with an input for the images:
> 
> image
> 
> It occurred to me an use case where the user can change the file name of an image, for example: an user can select one image with a select input, once selected he can write in an input text the new name of the image, when the user press submit the app will show the image with a successful message. This will be the "normal" case.
> 
> In the other case, the user write a new name with a path like: "/etc/passwd", and when the user press submit the app will show the passwd info.
> 
> What do you think about this approach?

Yeah, this use case looks good however it is not very clear to me. Can you please elaborate more? 
My thought was someone same as the example use case, i.e. we just create a gallery of images and each image is shown like  ``. If the user opens this URL directly in a new tab, the user will get the image and if the user tweaks the param value to `http://localhost:.../LFI?param=/etc/password` he should be able to see the contents of the directory.

Also if possible let's connect over a call to discuss? I work in IST time zone from 10 AM IST to 10 PM IST. Please let me know your preferred timings.

Ivan12273:
> > > > Alright!, then you suggest an input box only for the param, I will take this usecase, it is fine?
> > > 
> > > 
> > > ok, can you please check if this use-case is feasible or not? You need to do a little more research on this.
> > 
> > 
> > Hi @preetkaran20 sorry for the delay, I did a research about this use-case, is feasible, it would be the same idea as this scenario but with an input for the images:
> > image
> > It occurred to me an use case where the user can change the file name of an image, for example: an user can select one image with a select input, once selected he can write in an input text the new name of the image, when the user press submit the app will show the image with a successful message. This will be the "normal" case.
> > In the other case, the user write a new name with a path like: "/etc/passwd", and when the user press submit the app will show the passwd info.
> > What do you think about this approach?
> 
> Yeah, this use case looks good however it is not very clear to me. Can you please elaborate more?
> My thought was somewhat same as the example use case, i.e. we just create a gallery of images and each image is shown like ``. If the user opens this URL directly in a new tab, the user will get the image and if the user tweaks the param value to `http://localhost:.../LFI?param=/etc/password` he should be able to see the contents of the directory.
> 
> Also if possible let's connect over a call to discuss? I work in IST time zone from 10 AM IST to 10 PM IST. Please let me know your preferred timings.

Hello @preetkaran20 I am so sorry, it seems that for now I will not have enough time to continue with this, I will probably be available in the future but at the moment it is complicated.
If you want I can close the PR or leave it open.
Sorry again I whish I could helped more :(

preetkaran20:
Hi @Ivan12273,

No problem. No need to close the PR, we will pick this issue from here only. Thanks a lot for the contribution. 

thanks,
Karan

Issue #310, opened 2021-09-25
Issues with Hint Message in Error Based SQL Injection and union Based SQL Injections


preetkaran20:
**Describe the bug**
Error-based SQL Injection and Union-based SQL Injection has hints which have messages without space between 2 words in multiple levels. 

![image](https://user-images.githubusercontent.com/13570884/134773900-e5cde155-7cb6-4b4c-bd46-e9a46f234d83.png)

**How to fix**
Correct the string in messages files: https://github.com/SasanLabs/VulnerableApp/tree/master/src/main/resources/i18n.

priyanka010392:
Hello @preetkaran20, could you please assign it to me? I would like to work on it during Hacktober fest

preetkaran20:
Closing this issue as this is fixed by the #326

PR #326, opened 2021-09-30
Fix for Hint Message in Error Based SQL Injection and union Based SQL Injections

priyanka010392:
Fixes the[ issue](https://github.com/SasanLabs/VulnerableApp/issues/310)


preetkaran20:
Thank you @priyanka010392.

--
Karan

Issue #329, opened 2021-10-10
Adding CSRF vulnerability


preetkaran20:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

**Would you like to help fix this issue?**
Just say if you'd like to try and help fix this issue. We'll provide any advice and guidance you need.

AlexisOsmont:
Hello, I represent my group, we are in a Master's degree in cybersecurity and our final project is to contribute to some security issues on opensource projects, i would like to contribute to this issue, is it still open/available ? Thank u.

preetkaran20:
Hi @AlexisOsmont ,

Thanks for looking to contribute to this issue and Yes, this issue is open and available. 
In the start I would suggest to curate the list of vulnerability levels you are looking to add and let's discuss on them. (A small doc would be really helpful.)


Regards,
Karan

AlexisOsmont:
Hi again @preetkaran20 , 

we've seen that there is already an unmerged branch would you like us to work on this to finish the work based on your comments ? if the answer is yes may you give us the changes that you want please ? 

Regards,
@AlexisOsmont, @zgzorx.

preetkaran20:
Hi @AlexisOsmont @Zgzorx,

The unfinished PR is: https://github.com/SasanLabs/VulnerableApp/pull/364 and https://github.com/SasanLabs/VulnerableApp/pull/357 PR has all the comments.

The major conversation is at https://github.com/SasanLabs/VulnerableApp/pull/357/files/6cb5db77aa19f550b30567e7a8247ebc18797c8f#r820378181. 
You can follow from there onwards and please let me know if you need more information.

thanks,
Karan

Zgzorx:
Hi !

We managed to set the http_only cookie on level 3, we would like to be sure : does it need to be set on higher levels or not ?

Regards

preetkaran20:
@Zgzorx yes, Http_only should be set for all the higher levels.
I would also suggest to include other cookie flags that owasp recommends to make it more secure for higher levels if possible.

Issue #288, opened 2021-05-25
Add SSRF Vulnerability


preetkaran20:
**Is your feature request related to a problem? Please describe.**
SSRF is a very important vulnerability and it is currently not present in Owasp VulnerableApp. A good tutorial video on SSRF is https://www.youtube.com/watch?v=AsPpxqIlTDU

**Describe the solution you'd like**
Try finding the use-cases from other vulnerable applications and then add those to the VulnerableApp.

preetkaran20:
This Issue doesn't include the blind SSRF. 
Acceptance criteria:

1. Use-case finding
2. Backend API creation for Vulnerability.
3. UserInterface/Template creation for the Vulnerability.
4. Compatible with https://github.com/SasanLabs/VulnerableApp-facade schema design.

lfga98:
Hi, I would like to work on this issue but I was wondering if we could add 3 levels.

**Level 1: Basic SSRF** 
-   I've seen vulnerable applications where you enter and URL and an image is displayed, we could do something like that. This is level 1 so the vulnerability is very easy to exploit any URL it's going to work. I'll try to find a way to make **file:///etc/passwd** work.
- Input: A URL. http://localhost:80 or file:///etc/passwd
- Output: The requested file or an output showing a scanned port response

**Level 2: Filtered input SSRF**

- The same idea showing an image from a URL but now having filters in the backend that aren't going to block hostnames (127.0.0.1, localhost, etc) the idea here is to show users that a filter like that could be bypassed with an alternative IP representation.

**Level 3: Blind SSRF** 
- I don't have a well-defined idea on this one, but maybe we could show this by entering a command injection URL. I'm going to try to find a better way.

Taking this approach we will be covering basic, filtered, and maybe blind SSRF.

What do you think about this approach?

preetkaran20:
Hi @lfga98,

Thanks for taking up this issue and thinking about it. I think we can divide SSRF into 2 categories:
1. SSRF 2. Blind SSRF.
we are having a different issue for blind SSRF.

Now coming to the number of levels, We need to look at multiple vulnerable applications like https://portswigger.net/web-security/ssrf and others so that we can introduce multiple levels under one category.

Yeah, filtering can be a good example and we can add more to it to make it more complex in further levels.

thanks,
Karan

preetkaran20:
@lfga98 along with the level you added above, some other levels which can be added are: 
1. Port scanning 
2. Cloud Metadata service

[...]

PR #357, opened 2022-03-03
Adding DoS, SSRF and CSRF vulnerabilities

NMV01:
# This document gives the technical details of our contribution to VulnerableApp by adding DoS, SSRF and CSRF vulnerabilities

## Summary 
As mentioned by email with @preetkaran20 , we worked on adding various attacks. We have added some new levels for :
- DoS
- SSRF ([issue #288](https://github.com/SasanLabs/VulnerableApp/issues/288 "https://github.com/SasanLabs/VulnerableApp/issues/288"))
- CSRF ([issue #329](https://github.com/SasanLabs/VulnerableApp/issues/329))

### DoS

The **Denial of Service (DoS)** attack is focused on making a resource (site, application, server) unavailable for the purpose it was designed. 

We have added 2 levels for performing a DoS attack while uploading files. These levels are added to the *UnrestrictedFileUpload* class (levels 10 and 11). In order to exploit this vulnerability, we give the right to users to upload
many files. 

For *level 10*, the user needs to upload a large number of files or just a very large file in order to cause the environment to run out of available memory and causing DoS. We handled the exception of exceeding the maximum size configured, so the user can't upload files if the maximum configured size is exceeded.

When the environment runs out of memory, the user will get an alert : `something else other than 200/401 was returned`. This is normal as the device runs out of memory. At this level, the user can perform other vulnerabilities described at LEVEL 1 of File Upload Vulnerability.

For *level 11*, we have added some tests in order to make DoS attack impossible. A size limit for files uploaded is set (100000 kB) and the number of files that can be uploaded at the same time is set to 50. An exception will be raised if not more space available. We handled the exception of exceeding the maximum size configured, so the user can't upload files if the maximum configured size is exceeded. At this level, the user can perform other  vulnerabilities described at LEVEL 1 of File Upload Vulnerability.

### SSRF

In a **Server-Side Request Forgery (SSRF)** attack, the attacker can abuse functionality on the server to read or update internal resources.

We have added 2 more levels in order to implement SSRF by bypassing techniques based on IP addresses. These levels are added to the *SSRFVulnerability* class (levels 2 and 3).

For *level 2*, the user needs to bypass the protections that prevent `127.0.0.1` or `localhost`. To do so, the user can send a 
request to `127.1` and confirm that he receives a response.

For *level 3*, marked as secure, we have added an IP blacklist in order to block IPs that a user can use for performing an SSRF by a bypass.

### CSRF

**Cross-Site Request Forgery (CSRF)** is an attack that forces an end user to execute unwanted actions on a web application in which they’re currently authenticated.

The implementation of this attack is complex because it requires another connected user to execute action. So we decided to simulate this attack with three levels :
1. CSRF with get parameter
2. CSRF with post parameter
3. [SECURE] CSRF with get parameter, secured by random token
4. [SECURE] CSRF with post parameter, secured by random token

The goal of the attacker is to change a user's password by sending him a bad link.

*Contribution by [Project Opensource](https://github.com/Project-Opensource) .*


preetkaran20:
> Hi @preetkaran20 , We are trying to add our PR but we have problems with passing build test and we don't know why. Do we need to do something ? Thanks

run `./gradlew spotlessApply`

src/main/java/org/sasanlabs/service/vulnerability/csrf/CSRFVulnerability.java
```diff
@@ -0,0 +1,130 @@
...
+ *
+ * @author xxx
+ */
+@VulnerableAppRestController(
+        descriptionLabel = "CSRF_VULNERABILITY",
+        value = "CSRFVulnerability")
+public class CSRFVulnerability {
+
+    private static final String PTTRN_PGE = "resetpassword\\.php(\\?.*=.*)?";
+    private static final String PTTRN_GET = "resetpassword\\.php\\?(.*=.*&)?pwd=.*";
```
preetkaran20, line 33: php?

preetkaran20:
Tomorrow I will review the PR surely. 

Thanks, 
Karan

src/main/resources/attackvectors/UnrestrictedFileUploadPayload.properties
```diff
@@ -35,4 +35,13 @@ UNRESTRICTED_FILE_UPLOAD_PAYLOAD_LEVEL_7=Hackers can execute code in a user's ma
...
-  This level is vulnerable for path traversal, so if you choose a filename like <code>../index.html</code> then the file will be stored in another directory than provided.
\ No newline at end of file
+  This level is vulnerable for path traversal, so if you choose a filename like <code>../index.html</code> then the file will be stored in another directory than provided.
+UNRESTRICTED_FILE_UPLOAD_PAYLOAD_LEVEL_10=The Denial of Service (DoS) attack is focused on making a resource (site, application, server) \
+  unavailable for the purpose it was designed. Hackers can upload a large number of files or a very large file, in the proper format to the server, in order to cause a denial of service. Sometimes the hacker can inject and execute arbitrary code while performing a DoS attack in order to access critical information or execute commands on the server.<br/>\
+<b>How to exploit this level?</b><br/>\
+  Upload a large number of files or just a very large file in order to cause the environment to run out of available memory (be careful with not exceeding the configured max size). You will get an alert : <code>something else other than 200/401 was returned</code><br/>\
```
preetkaran20, line 42: we can correct `something else other than 200/401 was returned</code><br/>` such that for 413 error code we will show right message in the UI instead of user knowing this information.

preetkaran20 (changes requested):
Hi @NMV01 ,

I really liked the way you have understood the design of the project and was able to code very fast.
However, I have a few concerns:
1. CSRF vulnerability is not using DB and doesn't depict it correctly. I would also suggest to take only the value as input from user not the key. e.g. input box for username and inputbox for password and append the key like pwd=<value from inputbox> so that it can be easier and more usable. Have a look at DVWA for more information (https://www.youtube.com/watch?v=Nfb9E8MJv6k)
2. For DOS, I would suggest to only take input as a single large file, not multiple files. 
3. For SSRF, the current implementation is wrong, and i would suggest to first agree on the usecase first before starting implementation. you can start a discussion thread or schedule a meeting to discuss.

thanks,
Karan

src/main/java/org/sasanlabs/service/vulnerability/fileupload/UnrestrictedFileUpload.java
```diff
@@ -9,8 +9,7 @@
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.nio.file.StandardCopyOption;
-import java.util.Date;
-import java.util.Random;
+import java.util.*;
```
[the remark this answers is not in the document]
NMV01: Ok, we will fix this.

src/main/java/org/sasanlabs/service/vulnerability/ssrf/SSRFVulnerability.java
```diff
@@ -1,17 +1,27 @@
 package org.sasanlabs.service.vulnerability.ssrf;
 
 import com.nimbusds.jose.util.StandardCharset;
+import java.io.BufferedReader;
+import java.io.File;
+import java.io.IOException;
```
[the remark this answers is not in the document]
NMV01: Ok, we will fix this.

src/main/java/org/sasanlabs/service/vulnerability/fileupload/UnrestrictedFileUpload.java
```diff
@@ -379,4 +380,78 @@ Path getContentDispositionRoot() {
                 true,
                 false);
     }
+
+    @AttackVector(
```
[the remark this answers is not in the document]
NMV01: Ok, we will add this.

src/main/java/org/sasanlabs/service/vulnerability/fileupload/UnrestrictedFileUpload.java
```diff
@@ -379,4 +380,78 @@ Path getContentDispositionRoot() {
...
+                VulnerabilityType.REFLECTED_XSS,
+                VulnerabilityType.PATH_TRAVERSAL
+            },
+            description = "UNRESTRICTED_MULTIPLE_FILES_UPLOAD_DOS",
+            payload = "UNRESTRICTED_FILE_UPLOAD_PAYLOAD_LEVEL_10")
+    @VulnerableAppRequestMapping(
+            value = LevelConstants.LEVEL_10,
+            htmlTemplate = "LEVEL_10/MultipleFilesUpload",
+            requestMethod = RequestMethod.POST)
+    public ResponseEntity<GenericVulnerabilityResponseBean<String>> getVulnerablePayloadLevel10(
```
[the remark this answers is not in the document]
gled02: We gave the possibility to exploit DoS by uploading multiple files after our discussion by e-mail. We can modify this by letting a single file to upload, as a possibility to a user, but the DoS will be limited. 
It's there a problem if we upload multiple files to perform a DoS attack?

src/main/java/org/sasanlabs/service/vulnerability/ssrf/SSRFVulnerability.java
```diff
@@ -49,4 +61,87 @@
                         false),
                 HttpStatus.BAD_REQUEST);
     }
+
+    @AttackVector(
```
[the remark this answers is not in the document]
gled02: We didn't know that the actual implementation of SSRF was wrong, so we added our levels based at the present level. If the re-implementation doesn't need a long time to be done, we can discuss for changing it. Our contribution is limited in time. Can you give us some usecases that you think to implement?

src/main/resources/attackvectors/UnrestrictedFileUploadPayload.properties
gled02: Yes, we thought to correct this error, but when we tried to do this, we got the same message for the same error (500) at other levels of other vulnerabilities.

src/main/java/org/sasanlabs/service/vulnerability/fileupload/UnrestrictedFileUpload.java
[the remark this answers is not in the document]
preetkaran20: I think there was some confusion in our mail discussion, sorry for that.


The reason is consistency in all the levels and ease for scanners/users as they just need a single large file. How DoS will be limited? My suggestion is to use spring boot property for this level which allows upload of a very large file (check: https://stackoverflow.com/questions/40271484/spring-upload-file-size-limit). you can even enable DOS at other levels as well using this property but restrict in only one level which is secure.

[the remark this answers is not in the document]
gled02: DoS will be limited in terms of possible ways to exploit it. We have visited the link before, and the configuration of spring.servlet.multipart.max-file-size and spring.servlet.multipart.max-request-size, needs to be done at application.properties. So this configuration will be applied to all vulnerabilities, am I right? We can put this configuration in place, but in this case we need to add just a level that will block DoS. It is correct?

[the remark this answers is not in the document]
preetkaran20: yes, we need to find out a way such that property applies to all the other parts of VulnerableApplication except the FileUpload Vulnerability. and yes, in FileUpload let's block it for one or 2 levels only and all other levels will have DOS enabled.

src/main/resources/attackvectors/UnrestrictedFileUploadPayload.properties
preetkaran20: For other level's there is no upload functionality so 413 is not possible.

src/main/java/org/sasanlabs/service/vulnerability/ssrf/SSRFVulnerability.java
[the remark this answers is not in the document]
preetkaran20: yeah, this was added by mistake and hence we are stuck with the wrong SSRF implementation. Thanks for the interest in implementing it.

Sure, I am thinking of a use case where our application will call some endpoint which will list all the projects by SasanLabs but that URL will be controlled from the user interface(UI will only have a button which when pressed shows the projects by SasanLabs (gist page may be) but internally that button will invoke an SSRF endpoint which internally will invoke the provided url in query param)

There can be easily 4-5 levels, 
1. No restriction
2. user is disallowed from file:// extension
2. user is disallowed from localhost
3. user is disallowed from any variant of localhost
4. user is disallowed from metadataservice (you can ignore this one)
5. secure level where whitelist is made
6. secure level where user doesn't control the url.

let me know if have other idea's, as this is the only thing coming to my mind right now.

src/main/java/org/sasanlabs/service/vulnerability/csrf/CSRFVulnerability.java
```diff
@@ -0,0 +1,133 @@
...
+import org.springframework.http.HttpStatus;
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.bind.annotation.RequestParam;
+
+/**
+ * This class contains vulnerabilities related to CSRF.
+ *
+ * @author xxx
+ */
+@VulnerableAppRestController(descriptionLabel = "CSRF_VULNERABILITY", value = "CSRFVulnerability")
```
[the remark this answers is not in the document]
preetkaran20: perfect.

Yes, I agree that is some kind of an issue with CSRF but will cookies and Ajax salt (generation after user login), we can depict the levels of password change. 
Level1: Get call
Level2: Post call (to break the myth)
Level3: Token-based call but unsecured, just a count.
Level4: Token-based call but very random.

we can depict this to the human users by creating a webpage to construct the requests which cause the attack whose link we will mention below the actual practice vulnerability section (or in hints section) where we will add details like the following webpage is malicious and the user accidentally opened that then it will change their password without their knowledge which is not possible in level4. Or we can add some GitHub links to code snippets and ask users to open those HTML snippets in their browser and that will change their passwords.


I am going to this idea from https://youtu.be/hW2ONyxAySY?t=314 video from Troyhunt.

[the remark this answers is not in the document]
preetkaran20: for this implementation ensure that the cookie is only for a level and doesn't interfere with other levels.

src/main/java/org/sasanlabs/service/vulnerability/fileupload/UnrestrictedFileUpload.java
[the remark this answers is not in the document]
gled02: Normally, we can perform DoS to existing levels of FileUpload, by uploading a large file, as verifications on file size aren't present. We added level 10 for uploading multiple files, but we can remove it. 
Finally, we just need to add a level to block DoS (as level 11 that we added, but for one file uploaded), by checking if the file size is bigger than x kB. I don't see other ways to fix the size of files uploaded, without changing the property for all the application.

src/main/java/org/sasanlabs/service/vulnerability/csrf/CSRFVulnerability.java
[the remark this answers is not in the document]
NMV01: I'm not sure that I understand everything.
According to your message (and the video), the malicious request is generated by visiting a web page and not directly in a link (as we have already implemented) ?
So how the request generation page that you proposed, works exactly?
Once the url / page is visited, how to verify that the malicious request is correctly executed and the attack is performed?
If we want to really perform the attack, we must have another session (different than that of the current user), but I don't see how to implement this.
For me, the easiest solution is a test of inputs to verify that it produces the correct request (with regex for example).
In this case, I can change the presentation / context of the user's page to generate the request, but I don't see how we can execute the attack in this application.

src/main/java/org/sasanlabs/service/vulnerability/ssrf/SSRFVulnerability.java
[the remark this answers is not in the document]
gled02: I don't think I really understood the UI interface with a button to press in order to show the projects by SasanLabs. I think that a user needs to enter an URL and this URL needs to be verified in different levels. So, I don't really understand the problem with the current level of SSRF, as we need to enter an URL and when we enter an URL like file://etc/passwd, we have the right information. 

For levels, the two levels that we added can:
- level 2 -> block localhost and 127.0.0.1
- level 3 -> block all IPs present in a black list

src/main/resources/attackvectors/UnrestrictedFileUploadPayload.properties
gled02: Ok, we will correct this error

src/main/java/org/sasanlabs/service/vulnerability/ssrf/SSRFVulnerability.java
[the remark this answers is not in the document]
preetkaran20: ok so the current level is not right because of the following issues:
1. It fetches the image and then sends it to UI as a base64 encoded string using data URI which is currently only working with PNG, not with JPEG images or other types.
2. The UI code has some odd check that data should start with "i" which seems unmaintainable in the longer run and doesn't seems like a real world application works this way, atleast i have not seen any application like that.

Also blacklist is not a secured way for handling such usecases, it is very complex to implement the secured level for the current implementation.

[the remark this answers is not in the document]
preetkaran20: My suggestion was:
Have a simple button in the UI which loads the list of Projects by SasanLabs or on the page load (html) just go and fetch the projects by SasanLabs but pass the url from user interface. the backend level will take that url and calls it and returns the response from that url.

For level1 to level6 (explained above) will work on taking user input from UI where as level7 will not take the input at all and works directly. However, level6 and level7 both are secure. this way we can test the scanners with false-positive usecases as well.

src/main/java/org/sasanlabs/service/vulnerability/csrf/CSRFVulnerability.java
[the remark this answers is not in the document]
preetkaran20: Good questions,

1. No, The malicious request can be just a URI for GET call and will be an HTML page for POST calls.
2. The request generation/HTML page will be for post call and it will be similar to the way explained in the video.
3. The user password will be changed to something else and that will give indication that attack is successful.
4. Nope, for CSRF the session is the same just that user is not aware that a malicious website or a link it has clicked, by mistake has done a change to his session.

This may sound like there is no challenge for users here, but we can explain user about it in hints section. Have a look at https://www.youtube.com/watch?v=Nfb9E8MJv6k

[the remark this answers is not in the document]
preetkaran20: So this is how the entire process will work:

We will give a login page (on login we set the cookie) and then a password reset page to user where user can entire the new password and ok clicking change, it will change the password.
For level1: we will use GET call, in Hints we can explain to user that if user is tricked into clicking the URL then it can cause password change. 

For level2: it will be a POST call, in hints section, we will give a html page which will explain in Hints section that if that HTML is executed from a malicious page or if user is tricked into opening directly in browser it will change password without user knowing it.

similarly other levels.

One important thing is VulnerableApp's main target is scanner not human users so scanner should be able to figure out that ajax salt is not there so CSRF is possible.

src/main/java/org/sasanlabs/service/vulnerability/ssrf/SSRFVulnerability.java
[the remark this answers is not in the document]
gled02: Ok, I will start to re-implement SSRF by taking into consideration your suggestions. We'll keep you posted about our progress.

src/main/java/org/sasanlabs/service/vulnerability/csrf/CSRFVulnerability.java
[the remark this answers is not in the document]
NMV01: Ok i will see what i can do. I will keep you informed.

src/main/resources/attackvectors/UnrestrictedFileUploadPayload.properties
gled02: Hi @preetkaran20,

I'm trying to resolve the 413 error code, but FileSizeLimitExceededException leads to error 500 instead of 413. So, I don't know how to fix this problem. Thanks.

gled02: The only way i have in mind to fix the problem, is to implement a function similar to genericResponseHandler in fileupload.js, in order to catch 500 error code and then show a message in UI. In this way, other levels will not be affected.

src/main/java/org/sasanlabs/service/vulnerability/fileupload/UnrestrictedFileUpload.java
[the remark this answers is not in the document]
gled02: Hi, 

We implemented DoS vulnerability, as we discussed. All levels are vulnerable to DoS, except level 9 where we block DoS, by checking if the file size is bigger than 100000 kB. Only one file can be uploaded for each level.

src/main/java/org/sasanlabs/service/vulnerability/csrf/CSRFVulnerability.java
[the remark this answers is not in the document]
NMV01: Hello,
A new version of CSRF has been pushed.
Here are the levels :
1. simple GET
2. simple POST
3. GET with fixed token
4. POST with fixed token
5. GET with random token
6. POST with random token

I didn't implement "Setting Http only cookie" because I don't think I have the necessary skills (to be sure not to impact the program globally) and time.

For each incorrect input (wrong login/password, wrong action token), the user is automatically logged out.

Links (GET) and pages (POST) are created to perform and test the attacks (I can send it to you if you want). They are functional. Level 5 is secure. Level 6 is also considered secure (it may be a possibility to exploit it, but it's very difficult).

src/main/java/org/sasanlabs/service/vulnerability/ssrf/SSRFVulnerability.java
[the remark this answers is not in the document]
gled02: We use a file for testing that we placed in /VulnerableApp. I can push this file if you want to see the results of the tests.

NMV01:
Hi @preetkaran20 , 
We are trying to finalise our university project. Could you please check our PR, so we can put in place the necessary modifications during the remaining time.
Thanks

preetkaran20:
> Hi @preetkaran20 , We are trying to finalise our university project. Could you please check our PR, so we can put in place the necessary modifications during the remaining time. Thanks

Hi @NMV01 ,

I am going through the code and by tomorrow, I will complete the review.

thanks,
Karan

src/main/java/org/sasanlabs/service/vulnerability/csrf/CSRFVulnerability.java
```diff
@@ -0,0 +1,218 @@
...
+            }
+        } else if (validatorNP.get()) {
+            if (connect && securityOk) {
+                passlive = usrNewPass;
+                rtn = "<div id=\"lgin\">[Logged in]</div> Password changed correctly.";
+            } else {
+                rtn =
+                        (connect
+                                ? "Password change refused."
+                                : "<div id=\"lgou\">[Not Logged in]</div> You are not logged in. Login to change password.");
```
preetkaran20, line 64: Hi @NMV01 ,  If we don't use cookies then I think it will not be the right way to achieve CSRF.

For example, you are setting connect flag in the server after login so even an attacker calls the same endpoint with a change password request, it will change the password for the actual user.

I think this is not the right way to implement the CSRF attack. Changing cookie should not be very complex, we already have a code that sets the cookie at (https://github.com/SasanLabs/VulnerableApp/blob/master/src/main/java/org/sasanlabs/service/vulnerability/jwt/JWTVulnerability.java#L170-L173)

preetkaran20: I would also suggest that once the user logs in, please hide the login form and have a logout button which have logout functionalilty and on logout, show the login form.

preetkaran20: Also please don't share the password between different levels so that scanner can handle it.

src/main/java/org/sasanlabs/service/vulnerability/csrf/CSRFVulnerability.java
```diff
@@ -0,0 +1,218 @@
...
+        if (usrLiveToken.equals("secretrequest")) {
+            scrt = Integer.toString(rand.nextInt(1000000000));
+            return new ResponseEntity<GenericVulnerabilityResponseBean<String>>(
+                    new GenericVulnerabilityResponseBean<String>(scrt, true), HttpStatus.OK);
+        } else {
+            return new ResponseEntity<GenericVulnerabilityResponseBean<String>>(
+                    new GenericVulnerabilityResponseBean<String>(
+                            checkInputsToken(usrLiveToken, scrt, usrId, usrPass, usrNewPass), true),
+                    HttpStatus.OK);
+        }
```
preetkaran20, line 216: I think you are doing 2 calls on clicking reset password, one for checking the token and the other for changing the password. I would suggest having one single call which does both things.

src/main/java/org/sasanlabs/service/vulnerability/csrf/CSRFVulnerability.java
```diff
@@ -0,0 +1,218 @@
...
+            throws IOException {
+        if (usrLiveToken.equals("secretrequest")) {
+            scrt = Integer.toString(rand.nextInt(1000000000));
+            return new ResponseEntity<GenericVulnerabilityResponseBean<String>>(
+                    new GenericVulnerabilityResponseBean<String>(scrt, true), HttpStatus.OK);
+        } else {
+            return new ResponseEntity<GenericVulnerabilityResponseBean<String>>(
+                    new GenericVulnerabilityResponseBean<String>(
+                            checkInputsToken(usrLiveToken, scrt, usrId, usrPass, usrNewPass), true),
+                    HttpStatus.OK);
```
preetkaran20, line 188: Same as below comment.

src/main/java/org/sasanlabs/service/vulnerability/csrf/CSRFVulnerability.java
```diff
@@ -0,0 +1,218 @@
...
+
+    private static final String LOGIN = "mylogin";
+    private String passlive = "mypassword";
+    private Boolean connect = false;
+
+    private static final String RTK_PARAM = "rtkparam";
+    private static final String FIXED_TOKEN = "7877864849";
+    private String scrt = "";
+
+    private static final Random rand = new Random();
```
preetkaran20, line 41: use secure random from crypto package for better security.

src/main/java/org/sasanlabs/service/vulnerability/csrf/CSRFVulnerability.java
```diff
@@ -0,0 +1,218 @@
...
+public class CSRFVulnerability {
+
+    private static final String RTN_BADINPUT = "Bad input!";
+
+    private static final String USR_LOGIN = "id";
+    private static final String USR_PASS = "pass";
+    private static final String USR_NEWPASS = "newpass";
+
+    private static final String LOGIN = "mylogin";
+    private String passlive = "mypassword";
```
preetkaran20, line 34: Use SHA256 + hmac with salt (random), a hash of the password instead of a plain text password. Why? because we also support [SonarSource](https://sonarsource.com/). 
I would suggest adding one more level where put the MD5 hash so that we can expect SonarSource to find the vulnerability.

src/main/java/org/sasanlabs/service/vulnerability/csrf/CSRFVulnerability.java
```diff
@@ -0,0 +1,218 @@
...
+    private String checkInputs(
+            String usrId, String usrPass, String usrNewPass, Boolean securityOk) {
+        Supplier<Boolean> validatorId = () -> StringUtils.isNotBlank(usrId + usrPass);
+        Supplier<Boolean> validatorNP = () -> StringUtils.isNotBlank(usrNewPass);
+        String rtn;
+        if (validatorId.get()) {
+            if (usrId.equals(LOGIN) && usrPass.equals(passlive)) {
+                connect = true;
+                rtn = "<div id=\"lgin\">[Logged in]</div> Hello " + usrId + ".";
+            } else {
```
preetkaran20, line 52: set the cookie which makes it a complete test.

src/main/java/org/sasanlabs/service/vulnerability/csrf/CSRFVulnerability.java
preetkaran20, line 64: Use message.properties such that we can achieve multiple locales.

src/main/resources/static/templates/CSRFVulnerability/LEVEL_1/CSRFVulnerability.html
```diff
@@ -0,0 +1,30 @@
+<div id="CSRFVulnerability">
```
preetkaran20, line 1: you can reuse the HTML templates across levels so in case you are using the same js/CSS/HTML pages then I would suggest using the same templates. it will help reduce the duplicacy.

src/main/resources/static/templates/CSRFVulnerability/LEVEL_5/CSRFVulnerability.css
```diff
@@ -0,0 +1,63 @@
...
+#note {
+    font-size: 15px;
+    margin: 10px;
+}
+
+#lgin {
+    color: #09C700
+}
+
+#lgou {
```
preetkaran20, line 28: I would suggest giving full names so that it is easy to understand the functionality.

src/main/java/org/sasanlabs/service/vulnerability/csrf/CSRFVulnerability.java
```diff
@@ -0,0 +1,218 @@
...
+
+    private static final String USR_LOGIN = "id";
+    private static final String USR_PASS = "pass";
+    private static final String USR_NEWPASS = "newpass";
+
+    private static final String LOGIN = "mylogin";
+    private String passlive = "mypassword";
+    private Boolean connect = false;
+
+    private static final String RTK_PARAM = "rtkparam";
```
preetkaran20, line 37: Use the full names as variable names so that it is easier to understand functionality.

src/main/java/org/sasanlabs/service/vulnerability/csrf/CSRFVulnerability.java
```diff
@@ -0,0 +1,218 @@
...
+    private static final String FIXED_TOKEN = "7877864849";
+    private String scrt = "";
```
preetkaran20, line 39: same as above.

src/main/java/org/sasanlabs/service/vulnerability/csrf/CSRFVulnerability.java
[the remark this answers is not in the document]
preetkaran20: Http Only is pretty simple, we are already doing the same at: https://github.com/SasanLabs/VulnerableApp/blob/master/src/main/java/org/sasanlabs/service/vulnerability/jwt/JWTVulnerability.java#L205-L207

src/main/java/org/sasanlabs/service/vulnerability/csrf/CSRFVulnerability.java
```diff
@@ -0,0 +1,218 @@
...
+
+    private static final Random rand = new Random();
+
+    private String checkInputs(
+            String usrId, String usrPass, String usrNewPass, Boolean securityOk) {
+        Supplier<Boolean> validatorId = () -> StringUtils.isNotBlank(usrId + usrPass);
+        Supplier<Boolean> validatorNP = () -> StringUtils.isNotBlank(usrNewPass);
```
preetkaran20, line 46: Suggestion, to make supplier as static and final such that it will not take heap memory again and again.

src/main/java/org/sasanlabs/service/vulnerability/csrf/CSRFVulnerability.java
```diff
@@ -0,0 +1,218 @@
...
+        String rtn;
```
preetkaran20, line 47: use full names

src/main/java/org/sasanlabs/service/vulnerability/ssrf/SSRFVulnerability.java
```diff
@@ -1,52 +1,221 @@
 package org.sasanlabs.service.vulnerability.ssrf;
 
-import com.nimbusds.jose.util.StandardCharset;
-import java.io.InputStream;
+import java.io.BufferedReader;
```
preetkaran20, line 3: The file permissions should remain same. please revert the permissions to 100644.

src/main/resources/i18n/messages_en_US.properties
```diff
@@ -111,16 +111,17 @@ Important Links:<br/>\
...
-UNRESTRICTED_FILE_UPLOAD_NO_VALIDATION_FILE_NAME=There is no validation on uploaded file's name.
-UNRESTRICTED_FILE_UPLOAD_IF_NOT_HTML_FILE_EXTENSION=All file extensions are allowed except .html extensions.
-UNRESTRICTED_FILE_UPLOAD_IF_NOT_HTML_NOT_HTM_FILE_EXTENSION=All file extensions are allowed except .html and .htm extensions.
-UNRESTRICTED_FILE_UPLOAD_IF_NOT_HTML_NOT_HTM_FILE_EXTENSION_CASE_INSENSITIVE=All file extensions are allowed except case insensitive .html, .htm extensions.
-UNRESTRICTED_FILE_UPLOAD_IF_FILE_NAME_NOT_CONTAINS_.PNG_OR_.JPEG_CASE_INSENSITIVE=Only file name is allowed if it contains case insensitive .jpeg or .png.
-UNRESTRICTED_FILE_UPLOAD_IF_FILE_NAME_NOT_ENDS_WITH_.PNG_OR_.JPEG_CASE_INSENSITIVE_AND_FILE_NAMES_CONSIDERED_BEFORE_NULL_BYTE=Only file name is allowed if it ends with case insensitive .jpeg or .png and it is considered before Null Bytes only.
-UNRESTRICTED_FILE_UPLOAD_IF_FILE_NAME_NOT_ENDS_WITH_.PNG_OR_.JPEG_CASE_INSENSITIVE=Only file name is allowed if it ends with case insensitive .jpeg or .png.
+UNRESTRICTED_FILE_UPLOAD_NO_VALIDATION_FILE_NAME=There is no validation on uploaded file's name. You can perform a DoS attack.
```
preetkaran20, line 118: Better text would be "This level is also Vulnerable to DoS attack"

src/main/java/org/sasanlabs/service/vulnerability/ssrf/SSRFVulnerability.java
```diff
@@ -1,52 +1,221 @@
...
+    private static final String[] URL_WHITE_LIST = {
+        "https://github.com/SasanLabs/",
+        "https://github.com/SasanLabs/VulnerableApp",
+        "https://github.com/SasanLabs/owasp-zap-jwt-addon",
+        "https://github.com/SasanLabs/owasp-zap-fileupload-addon",
+        "https://github.com/SasanLabs/VulnerableApp-php",
+        "https://github.com/SasanLabs/VulnerableApp-jsp"
+    };
+
+    boolean isUrlValid(String url) {
```
preetkaran20, line 49: add Private visibility to the method.

src/main/java/org/sasanlabs/service/vulnerability/ssrf/SSRFVulnerability.java
```diff
@@ -1,52 +1,221 @@
...
+        try {
+            URL obj = new URL(url);
+            obj.toURI();
+            return true;
+        } catch (MalformedURLException | URISyntaxException e) {
+            return false;
+        }
+    }
+
+    StringBuilder writeResult(Reader in) throws IOException {
```
preetkaran20, line 59: please add `private`

src/main/java/org/sasanlabs/service/vulnerability/ssrf/SSRFVulnerability.java
```diff
@@ -1,52 +1,221 @@
...
+        StringBuilder content = new StringBuilder();
+        BufferedReader bufferedReader = new BufferedReader(in);
+        String line;
+        while ((line = bufferedReader.readLine()) != null) {
+            content.append(line).append("\n");
+        }
+        bufferedReader.close();
+        return content;
+    }
```
preetkaran20, line 68: use `try with resource` else an exception will cause DoS

src/main/java/org/sasanlabs/service/vulnerability/ssrf/SSRFVulnerability.java
```diff
@@ -1,52 +1,221 @@
...
+
+    ResponseEntity<GenericVulnerabilityResponseBean<StringBuilder>>
```
preetkaran20, line 70: Add private

src/main/java/org/sasanlabs/service/vulnerability/ssrf/SSRFVulnerability.java
```diff
@@ -1,52 +1,221 @@
...
-import org.springframework.util.StreamUtils;
 import org.springframework.web.bind.annotation.RequestParam;
 
 @VulnerableAppRestController(descriptionLabel = "SSRF_VULNERABILITY", value = "SSRFVulnerability")
 public class SSRFVulnerability {
 
-    private static final String IMAGE_URL = "imageurl";
-    private static final transient Logger LOGGER = LogManager.getLogger(SSRFVulnerability.class);
+    private static final String FILE_URL = "fileurl";
+    private static final String FILE_PATH = "static/files/SSRF.txt";
```
preetkaran20, line 33: There is no need to make the challenge here as our target audience is scanner. I would still suggest to have just one button in UI or even no button in UI and as we load the level, we do an Api call to backend which will load the name of projects by SasanLabs.
Every level will send the url from where to load the projects from user input except the last one which is super secure.

preetkaran20: I would still suggest to have just one button in UI or even no button in UI and as we load the level, we do an Api call to backend which will load the name of projects by SasanLabs.
Every level will send the url from where to load the projects from user input except the last one which is super secure.

preetkaran20: so from your side nothing will change in backend except in front end you will be sending a fixed url always.

src/main/java/org/sasanlabs/service/vulnerability/ssrf/SSRFVulnerability.java
```diff
@@ -1,52 +1,221 @@
...
-        }
-        return new ResponseEntity<>(
-                new GenericVulnerabilityResponseBean<>(
-                        ("Failed to fetch image from URL " + urlImage)
-                                .getBytes(StandardCharset.UTF_8),
-                        false),
-                HttpStatus.BAD_REQUEST);
+                    new GenericVulnerabilityResponseBean<>(
+                            new StringBuilder("URL not valid"), false),
+                    HttpStatus.OK);
```
preetkaran20, line 218: No need for checking if url is valid or not, just compare with whitelist and if it doesn't match return not a valid url.

src/main/resources/static/templates/UnrestrictedFileUpload/LEVEL_1/FileUpload.js
```diff
@@ -1,13 +1,46 @@
...
+
 function addingEventListenerToUploadImage() {
   document.getElementById("upload").addEventListener("click", function () {
-    var form = document.getElementById("file");
-    var file = form.files[0];
-    var formData = new FormData();
-    formData.append("file", file);
-    let url = getUrlForVulnerabilityLevel();
-    doPostAjaxCall(uploadImage, url, true, formData);
+    const form = document.getElementById("file");
```
preetkaran20, line 22: I think we can allow every call from UI and let's not have a check in  the UI. thoughts?

src/main/resources/static/templates/UnrestrictedFileUpload/LEVEL_1/FileUpload.js
```diff
@@ -1,13 +1,46 @@
+function genericResponseHandler2(xmlHttpRequest, callBack, isJson) {
```
preetkaran20, line 1: Good check.

src/main/resources/static/templates/UnrestrictedFileUpload/LEVEL_1/FileUpload.js
```diff
@@ -1,13 +1,46 @@
...
   document.getElementById("upload").addEventListener("click", function () {
-    var form = document.getElementById("file");
-    var file = form.files[0];
-    var formData = new FormData();
-    formData.append("file", file);
-    let url = getUrlForVulnerabilityLevel();
-    doPostAjaxCall(uploadImage, url, true, formData);
+    const form = document.getElementById("file");
+    const file = form.files[0];
+    if (file.size > 1048576) {
```
preetkaran20, line 24: I think we can allow every call from UI and let's not have a check in the UI. thoughts?

src/main/resources/static/templates/UnrestrictedFileUpload/LEVEL_1/FileUpload.js
```diff
@@ -1,13 +1,46 @@
...
+      document.getElementById("uploaded_file_info").innerHTML =
+        "Max upload size exceeded exception : " +
```
preetkaran20, line 26: one more thing, i don't see any spring property update, does that means you it still blocks 10MB file?

src/main/resources/static/templates/SSRFVulnerability/LEVEL_1/SSRF.js
```diff
@@ -1,41 +1,16 @@
...
-      var img = document.createElement("img");
-      img.src = "data:image/png;base64," + data.content;
-      img.style.width = "500px";
-      img.style.height = "auto";
-      document.getElementById("imageResponse").append(img);
-    } else {
-      document.getElementById("imageResponse").append(atob(data.content));
-    }
-  }
+  document.getElementById("fileResponse").innerHTML = data.content;
```
preetkaran20, line 15: Good code.

src/main/resources/static/templates/SSRFVulnerability/LEVEL_1/SSRF.html
```diff
@@ -1,15 +1,15 @@
...
-        <button id="showBtn">Show Image</button>
+        <input type="text" id="urlInput" placeholder="File URL"/>
+        <button id="showBtn">Load File</button>
     </div>
 
     <div>
-        <div id="imageResponse" ></div>
+        <div id="fileResponse" ></div>
     </div>
 </div>
```
preetkaran20, line 14: we will just have a simple html page showing list of projects by SasanLabs (let's create a github gist for that) and that html page for each of the levels will send a call to backend and backend will call that gist page.
The 2nd last level will validate the url and last one will not take url as input. (both of the are secure)

src/main/resources/i18n/messages_en_US.properties
```diff
@@ -280,4 +281,25 @@ Important Links on SSRF : \
  <li> <a href="https://owasp.org/www-community/attacks/Server_Side_Request_Forgery" target="_blank">Owasp SSRF</a> \
  <li> <a href="https://www.youtube.com/watch?v=AsPpxqIlTDU" target="_blank">SSRF (Server Side Request Forgery) by Musab Khan</a>\
  </ol> 
-IMAGE_URL_PASSED_TO_REQUEST=The URL that the user enters is directly passed to a request.
\ No newline at end of file
+FILE_URL_WITHOUT_CHECK=No check for user input data. The application is not prepending any data string.
+FILE_URL_IF_NOT_FILE_PROTOCOL=The file will be loaded from URL, only if the URL does not contain the file:// protocol.
```
preetkaran20, line 285: file:///  (3 slashes)

src/main/java/org/sasanlabs/service/vulnerability/ssrf/SSRFVulnerability.java
```diff
@@ -1,52 +1,221 @@
...
 
 @VulnerableAppRestController(descriptionLabel = "SSRF_VULNERABILITY", value = "SSRFVulnerability")
 public class SSRFVulnerability {
 
-    private static final String IMAGE_URL = "imageurl";
-    private static final transient Logger LOGGER = LogManager.getLogger(SSRFVulnerability.class);
+    private static final String FILE_URL = "fileurl";
+    private static final String FILE_PATH = "static/files/SSRF.txt";
+    private static final String FILE_NAME = "SSRF.txt";
+    private static final String FILE_PROTOCOL = "file://";
```
preetkaran20, line 35: file:///

src/main/java/org/sasanlabs/service/vulnerability/ssrf/SSRFVulnerability.java
```diff
@@ -1,52 +1,221 @@
...
+    public ResponseEntity<GenericVulnerabilityResponseBean<StringBuilder>>
+            getVulnerablePayloadLevel2(@RequestParam(FILE_URL) String url)
+                    throws IOException, URISyntaxException {
+        if (isUrlValid(url)) {
+            if (!url.startsWith(FILE_PROTOCOL)) {
+                return getGenericVulnerabilityResponseWhenURL(url);
+            } else {
+                return new ResponseEntity<>(
+                        new GenericVulnerabilityResponseBean<>(
+                                new StringBuilder("file:// protocol is not allowed"), false),
```
preetkaran20, line 152: I would suggest, creating a single object as private static final and reuse it to have smaller footprint.

files/SSRF.txt
```diff
@@ -0,0 +1 @@
+Your attack was successful!!
```
preetkaran20, line 1: we don't need this, instead create a github gist page (I will create for you which has all the project names by SasanLabs)

src/main/java/org/sasanlabs/service/vulnerability/csrf/CSRFVulnerability.java
```diff
@@ -0,0 +1,218 @@
...
+import org.sasanlabs.vulnerability.types.VulnerabilityType;
+import org.springframework.http.HttpStatus;
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.bind.annotation.RequestMethod;
+import org.springframework.web.bind.annotation.RequestParam;
+
+/**
+ * This class contains vulnerabilities related to CSRF.
+ *
+ * @author xxx
```
preetkaran20, line 22: Please add your Email and Github user name.

src/main/resources/static/templates/CSRFVulnerability/LEVEL_5/CSRFVulnerability.js
```diff
@@ -0,0 +1,50 @@
...
+}
+addingEventListenerToResetPassButton();
+
+function CallbackSecret(data) {
+  let url = getUrlForVulnerabilityLevel();
+  doGetAjaxCall(
+    CallbackResponse,
+    url +
+      "?id=" +
+      "&pass=" +
```
preetkaran20, line 39: No need to send the query params if they are not used.

src/main/resources/static/templates/CSRFVulnerability/LEVEL_6/CSRFVulnerability.js
```diff
@@ -0,0 +1,41 @@
+function addingEventListenerToSendButton() {
+  document.getElementById("sendBtn").addEventListener("click", function () {
+    let url = getUrlForVulnerabilityLevel();
+    let dataPost = new FormData();
+    dataPost.append("id", document.getElementById("id").value);
+    dataPost.append("pass", document.getElementById("pass").value);
+    dataPost.append("newpass", "");
+    dataPost.append("rtkparam", "");
```
preetkaran20, line 8: No need to send the parameters which are not used. (you can configured json ignore on the java pojo class.)

src/main/java/org/sasanlabs/service/vulnerability/ssrf/SSRFVulnerability.java
```diff
@@ -1,52 +1,221 @@
...
 import org.sasanlabs.service.vulnerability.bean.GenericVulnerabilityResponseBean;
 import org.sasanlabs.vulnerability.types.VulnerabilityType;
+import org.springframework.core.io.ClassPathResource;
 import org.springframework.http.HttpStatus;
 import org.springframework.http.ResponseEntity;
-import org.springframework.util.StreamUtils;
 import org.springframework.web.bind.annotation.RequestParam;
 
 @VulnerableAppRestController(descriptionLabel = "SSRF_VULNERABILITY", value = "SSRFVulnerability")
 public class SSRFVulnerability {
```
preetkaran20, line 30: add your author details.

src/main/resources/static/files/SSRF.txt
```diff
@@ -0,0 +1 @@
+Your attack was successful!!
```
preetkaran20, line 1: we don't need this file.

src/main/java/org/sasanlabs/vulnerability/types/VulnerabilityType.java
```diff
@@ -32,10 +32,18 @@
 
     // SSRF Vulnerability
     SIMPLE_SSRF(918, 15),
+    SSRF_WITH_INPUT_FILTERS(918, 15),
+    SSRF_WHITE_LIST(918, 15),
```
preetkaran20, line 36: Let's have just Simple_SSRF, No need to create 2 new Vulnerability types.

src/main/java/org/sasanlabs/service/vulnerability/ssrf/SSRFVulnerability.java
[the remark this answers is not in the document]
preetkaran20: My suggestion in the UI side:

we will just have a simple html page showing list of projects by SasanLabs (let's create a github gist for that) and that html page for each of the levels will send a call to backend and backend will call that gist page.
The 2nd last level will validate the url and last one will not take url as input. (both of them are secure)

Why i am not in favour of this design is, because the final goal is not to have a file viewer as you added the whitelist logic.

preetkaran20 (changes requested):

preetkaran20 (changes requested):
Hi @NMV01 @gled02,

Overall, this is amazing progress, I have added a few comments and once they are done, I think we should be able to complete the PR by this week.

thanks,
Karan

gled02:
Hi @preetkaran20,

Thank you for your review. 
Despite our wish to advance the contribution, we don’t have much time because, as we mentioned before, we are finalising our university project. So we can't do all the modifications mentioned, such as re-implementing a vulnerability entirely.

We can do the following modifications:

1. CSRF

We don't know if our implementation of CSRF is the right one as you mentioned that this is not the right way to implement the CSRF attack.  So if our implementation is not the right one, we can't re-implement it entirely. The changes demanded on the code will be useless, if our implementation is not the correct one and will not be added to the application. 

2. SSRF

We decided to implement SSRF based on one of the suggestion you gave us. We can't change the UI part because the backend will be affected and we need more time to do this. But, we can modify parts of the code like making methods private, changing file permissions, deleting useless files etc.

To conclude, we are able to improve the existing implementation based on your suggestions, but due to the lack of time we can't re-implement all entirely. If you think that there are parts of our code that you don't want to be part of PR, we can remove them. We would be grateful if our PR is accepted. Then, the future contributors can improve what we started. 

Thanks

preetkaran20:
> Hi @preetkaran20,
> 
> Thank you for your review. Despite our wish to advance the contribution, we don’t have much time because, as we mentioned before, we are finalising our university project. So we can't do all the modifications mentioned, such as re-implementing a vulnerability entirely.
> 
> We can do the following modifications:
> 
> 1. CSRF
> 
> We don't know if our implementation of CSRF is the right one as you mentioned that this is not the right way to implement the CSRF attack. So if our implementation is not the right one, we can't re-implement it entirely. The changes demanded on the code will be useless, if our implementation is not the correct one and will not be added to the application.
> 
> 2. SSRF
> 
> We decided to implement SSRF based on one of the suggestion you gave us. We can't change the UI part because the backend will be affected and we need more time to do this. But, we can modify parts of the code like making methods private, changing file permissions, deleting useless files etc.
> 
> To conclude, we are able to improve the existing implementation based on your suggestions, but due to the lack of time we can't re-implement all entirely. If you think that there are parts of our code that you don't want to be part of PR, we can remove them. We would be grateful if our PR is accepted. Then, the future contributors can improve what we started.
> 
> Thanks

Hi @gled02 ,

For CSRF, I think it will take some time to complete the implementation but for SSRF, I think you are almost done and we can merge the PR very soon.

If you still think it will take more time finalizing then my suggestion is to raise 2 different PRs, 1. with CSRF and the other with SSRF. we can merge in different branches and I will rectify and make it proper by next month.

Let me know how can we proceed further?

thanks,
Karan

src/main/resources/static/templates/UnrestrictedFileUpload/LEVEL_1/FileUpload.js
gled02: We didn't add any spring property because this will affect the other levels and the entire applications. We just added the following : test file.size > 1048576, in order to handle the exception of Maximum Size Configured.

preetkaran20:
> Hi @preetkaran20,
> 
> We created 2 new PR, one for CSRF (#364) and one for SSRF & DoS (#365). We are really sorry, we don't have the time to change our code for SSRF. However, we made some changes (author name, private methods, file permission, etc.) in order to improve the quality of our code based on your suggestions.
> 
> Thanks,

@gled02 @NMV01 , you guys have done an amazing job, just that we are not able to merge the PRs in official branches.

PR #363, opened 2022-03-14
Nimbus jose jwt library upgrade due to vulnerabilities

SampathKumarAmex:
- Nimbus jose jwt library upgrade due to vulnerabilities

https://mvnrepository.com/artifact/com.nimbusds/nimbus-jose-jwt/8.3

https://github.com/advisories/GHSA-g5vf-v6wf-7w2r
https://github.com/advisories/GHSA-269g-pwp5-87pp

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -113,7 +113,7 @@ dependencies {
     implementation group: 'org.json', name: 'json', version: '20190722'
 
 	//https://mvnrepository.com/artifact/com.nimbusds/nimbus-jose-jwt
-    implementation group: 'com.nimbusds', name: 'nimbus-jose-jwt', version: '8.3'
+    implementation group: 'com.nimbusds', name: 'nimbus-jose-jwt', version: '9.12'
 
     // https://mvnrepository.com/artifact/commons-io/commons-io
     implementation group: 'commons-io', name: 'commons-io', version: '2.7'


build.gradle
```diff
@@ -113,7 +113,7 @@ dependencies {
     implementation group: 'org.json', name: 'json', version: '20190722'
 
 	//https://mvnrepository.com/artifact/com.nimbusds/nimbus-jose-jwt
-    implementation group: 'com.nimbusds', name: 'nimbus-jose-jwt', version: '8.3'
+    implementation group: 'com.nimbusds', name: 'nimbus-jose-jwt', version: '9.12'
```
preetkaran20, line 116: @SampathKumarAmex this should not break our existing vulnerable levels. Can you please check vulnerable level using this library and see if that is working as expected

SampathKumarAmex: Could you please provide more info on how to test that? Do you have any documentation covering that part? 

It's good to have some test for this if feasible.

preetkaran20: @SampathKumarAmex we need to start VulnerableApp, go to UI, and test that each level doesn't break.

preetkaran20: yeah, we are lacking unit tests and other things right now.

PR #372, opened 2022-07-14
fixes major code smells in GenericUtils.java

hks1:
1. Adds a private constructor
2. Adds missing 'Deprecated' annotation

### DIFF
--- a/src/main/java/org/sasanlabs/internal/utility/GenericUtils.java
+++ b/src/main/java/org/sasanlabs/internal/utility/GenericUtils.java
@@ -9,11 +9,14 @@ public class GenericUtils {
 
     public static final String LOCALHOST = "127.0.0.1";
 
+    private GenericUtils() {}
+
     /**
      * @deprecated
      * @param payload
      * @return
      */
+    @Deprecated
     public static String wrapPayloadInGenericVulnerableAppTemplate(String payload) {
         String generalPayload =
                 "<html><title>Security Testing</title><body><h1>Vulnerable Application </h1> %s </body></html>";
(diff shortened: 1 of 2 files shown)


src/main/java/org/sasanlabs/internal/utility/JSONSerializationUtils.java
```diff
@@ -17,6 +17,8 @@ public class JSONSerializationUtils {
 
     private static final ObjectMapper MAPPER = new ObjectMapper();
 
+    private static JSONSerializationUtils() {}
```
preetkaran20, line 20: static constructor?

src/main/java/org/sasanlabs/internal/utility/GenericUtils.java
```diff
@@ -9,11 +9,14 @@ public class GenericUtils {
 
     public static final String LOCALHOST = "127.0.0.1";
 
+    private GenericUtils() {}
```
preetkaran20, line 12: can you please mark this class as final as well?

src/main/java/org/sasanlabs/internal/utility/JSONSerializationUtils.java
preetkaran20, line 20: please mark this class as final.

src/main/java/org/sasanlabs/internal/utility/GenericUtils.java
hks1: sure

src/main/java/org/sasanlabs/internal/utility/JSONSerializationUtils.java
hks1: sure

hks1: my bad, that's an oversight. I'll address it in next commit

preetkaran20:
@hks1 Thank you !!!

Issue #353, opened 2022-02-22
Currently docker images are not versioned which can be an issue going further


preetkaran20:
**Is your feature request related to a problem? Please describe.**
Docker images for VulnerableApp have just 2 versions latest and unreleased. It should be latest (whcih is unreleased) and specific tags for released versions.

For more information about the existing images:  https://hub.docker.com/r/sasanlabs/owasp-vulnerableapp

**How do we create images ?**
we have a Github action for creating and pushing docker images:
https://github.com/SasanLabs/VulnerableApp/blob/master/.github/workflows/docker.yml, we use a tool called jib for this

merry-degaga:
Hello, 

I would like to take on this issue. 

I was looking into GitHub Workflow, actions, jib, the current docker.yml, docker hub and had a few questions.

1. To make sure I understand the issue, the Docker 'unreleased' tag is on the most recently pushed image (that has not been released). 
2. Then, did you want the Docker 'latest' tag on the most recently pushed image (it can be either released or unreleased)? Or should it always be tagged to a released version?
3. I was also looking into gitversion to automate the semantic versioning. There is a GitHub action available for that or would you rather have developers create tags that the workflow can use. 
4. Currently the workflow is triggered when a push to main occurs. What do you think of having this triggered instead, when a pull request is approved? 

Thank you

preetkaran20:
@merry-degaga  sorry, I missed your message. Are you working on this task ?

preetkaran20:
> Hello,
> 
> I would like to take on this issue.
>
> I was looking into GitHub Workflow, actions, jib, the current docker.yml, docker hub and had a few questions.
> 
> 1. To make sure I understand the issue, the Docker 'unreleased' tag is on the most recently pushed image (that has not been released).

Yes, unreleased is the not yet release docker image.

> 2. Then, did you want the Docker 'latest' tag on the most recently pushed image (it can be either released or unreleased)? Or should it always be tagged to a released version?

I think latest should be released tag.

> 3. I was also looking into gitversion to automate the semantic versioning. There is a GitHub action available for that or would you rather have developers create tags that the workflow can use.

How would github action know, we want to release a newer version ?

> 4. Currently the workflow is triggered when a push to main occurs. What do you think of having this triggered instead, when a pull request is approved?

I think we should always triggered on merge on the main branch. For PR docker's, we are not getting any requirement till now.

> Thank you

merry-degaga:
[...]

PR #382, opened 2022-08-07
Feature/issue 353

merry-degaga:
Fixes: [353](https://github.com/SasanLabs/VulnerableApp/issues/353)

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -34,6 +34,7 @@ jobs:
         name: Publish release on GitHub
         uses: softprops/action-gh-release@v1
         with:
+          name: 'VulnerableApp-${{ steps.gitversion.outputs.semver }}'
           tag_name: '${{ steps.gitversion.outputs.semver }}'
       -
         name: Grant execute permission for gradlew
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -22,7 +22,7 @@ jobs:
           java-version: 1.8
           fetch-depth: 0
       -
-        name: Setup Gitveresion action
+        name: Setup GitVersion action
         uses: gittools/actions/gitversion/setup@v0.9.7
         with:
           versionSpec: '5.x'
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -20,9 +20,26 @@ jobs:
         uses: actions/checkout@v2
         with:
           java-version: 1.8
+          fetch-depth: 0
+      -
+        name: Setup Gitveresion action
+        uses: gittools/actions/gitversion/setup@v0.9.7
+        with:
+          versionSpec: '5.x'
+      -
+        name: Execute GitVersion action
+        id: gitversion
+        uses: gittools/actions/gitversion/execute@v0.9.7
+      -
+        name: Publish release on GitHub
+        uses: softprops/action-gh-release@v1
+        with:
+          tag_name: '${{ steps.gitversion.outputs.semver }}'
       -
         name: Grant execute permission for gradlew
         run: chmod +x gradlew
       -
         name: Build with Gradle and Push
-        run: ./gradlew jib
+        run: |
+          ./gradlew jib \
+          -Djib.to.tags='latest',$GITVERSION_SEMVER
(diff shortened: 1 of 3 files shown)


.github/workflows/docker.yml
```diff
@@ -20,9 +20,26 @@ jobs:
...
+          tag_name: '${{ steps.gitversion.outputs.semver }}'
       -
         name: Grant execute permission for gradlew
         run: chmod +x gradlew
       -
         name: Build with Gradle and Push
-        run: ./gradlew jib
+        run: |
+          ./gradlew jib \
+          -Djib.to.tags='latest',$GITVERSION_SEMVER
```
preetkaran20, line 45: what would be `GITVERSION_SEMVER`?

.github/workflows/docker.yml
```diff
@@ -20,9 +20,26 @@ jobs:
         uses: actions/checkout@v2
         with:
           java-version: 1.8
+          fetch-depth: 0
+      -
+        name: Setup GitVersion action
```
preetkaran20, line 25: can you please tell how will be mention Major and Minor? So we need to mention in PR name?

preetkaran20: i read your comment and it seems like we need to mention in commit message.

preetkaran20: can you please update the Readme with what we need to mention for major release and minor release? Also with example explain what is the impact of major release, minor release and Patch in the github tag and docker version ?

.github/workflows/docker.yml
merry-degaga: It is an output from gitversion action. When gitverson is executed it provides a number of outputs (see [here](https://github.com/GitTools/actions/blob/main/docs/examples/github/gitversion/execute/usage-examples.md) for full list) and semVer is one of them. It will contain the calculated version for a build (see [here](https://github.com/GitTools/actions/blob/main/docs/examples/github/gitversion/execute/usage-examples.md#example-6) for example)

.github/workflows/docker.yml
merry-degaga: ok will get this done

.github/workflows/docker.yml
preetkaran20: ok got it. can you please add few sample versions here which will be the output of GITVERSION_SEMVER?

merry-degaga: Sample outputs of GITVERSION_SEMVER on DockerHub :
image
image
image

preetkaran20: so we already have 10 tags ... will this auto increment start from those tags. i.e. we have 1.10.0 tag so it will start as 1.10.1 or 1.11.0 or 2.0.0 right ?

preetkaran20 (approved):
LGTM. Great work @merry-degaga .

thanks,
Karan

.github/workflows/docker.yml
preetkaran20: @merry-degaga I merged this PR and I see that it has done a github release for the project where as i was thinking that it will just update the github tag and we will do the release as per our will.
https://github.com/SasanLabs/VulnerableApp/releases


is there a way we can stop it to create a release? From release, I point it as a major change so we notify about it in multiple channels.

preetkaran20: I deleted the release for now.

preetkaran20:
@merry-degaga  please help with 
https://github.com/SasanLabs/VulnerableApp/pull/382#discussion_r945217490

.github/workflows/docker.yml
merry-degaga: Okay, I will take a look.

> @merry-degaga please help with [#382 (comment)](https://github.com/SasanLabs/VulnerableApp/pull/382#discussion_r945217490)

On OWASP-Vulnerableapp's DockerHub, I do see the updated version tag (1.11.0) and latest tags both attached to the most recently built image. Was there any other functionality you needed?

I will look into preventing auto-release creation.

merry-degaga: I was able to prevent auto-release creation.

PR #384, opened 2022-08-14
Feature/382 fix

merry-degaga:
Workflow does not auto-create a new release, it only creates a new tag.
Changed 'latest' to 'unreleased' in workflow for Docker tag for most recent build/deploy to better align with manual release creation on GitHub.

### DIFF
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -1,0 +1,48 @@
+# This workflow automates release creation and builds and deploys to DockerHub
+name: Automates release gets latest tag, creates release, builds and push to DockerHub with latest tag
+
+on:
+  workflow_dispatch:
+    inputs:
+      release_notes:
+        description: 'Optional notes for the release'
+        required: false
+        type: string
+
+jobs:
+  login:
+    runs-on: ubuntu-latest
+    steps:
+      -
+        name: Checkout new changes 
+        uses: actions/checkout@v2
+        with: 
+          java-version: 1.8
+          fetch-depth: 0
+      -
+        name: Get the latest tag in repository
+        id: TAG
+        run: |
+          echo "::set-output name=LATEST_VERSION::$(git tag | sort -Vr | head -n 1)"
+      -
+        name: Login to DockerHub
+        uses: docker/login-action@v1
+        with: 
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_PASSWORD }}
+      -
+        name: Publish release on GitHub
+        uses: softprops/action-gh-release@v1
+        with:
+          name: 'VulnerableApp-${{ steps.TAG.outputs.LATEST_VERSION }}'
+          tag_name: '${{ steps.TAG.outputs.LATEST_VERSION }}'
+          body: '${{ inputs.release_notes }}'
+      -
+        name: Grant execute permission for gradlew
+        run: chmod +x gradlew
+      -
+        name: Build and push image to Docker using JIB
+        run: |
+            ./gradlew jib \
+            -Djib.to.image=sasanlabs/owasp-vulnerableapp:latest \
+            -Djib.to.tags=${{ steps.TAG.outputs.LATEST_VERSION }}
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -1,5 +1,5 @@
 # This workflow automates release creation and builds and deploys to DockerHub
-name: Automates release gets latest tag, creates release, builds and push to DockerHub with latest tag
+name: Release project
 
 on:
   workflow_dispatch:
@@ -44,5 +44,4 @@ jobs:
         name: Build and push image to Docker using JIB
         run: |
             ./gradlew jib \
-            -Djib.to.image=sasanlabs/owasp-vulnerableapp:latest \
             -Djib.to.tags=${{ steps.TAG.outputs.LATEST_VERSION }}
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -1,5 +1,5 @@
 # This workflow will build a Java project with Gradle, create repository tag, and push a new docker image to docker hub
-name: Create Tag, Build and Push Unreleased Image to DockerHub
+name: Create Tag and Push Image to DockerHub
 
 on:
   push:
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -31,7 +31,7 @@ jobs:
         id: gitversion
         uses: gittools/actions/gitversion/execute@v0.9.7
       -
-        name: Create tag for repo
+        name: Create tag for GitHub
         uses: tvdias/github-tagger@v0.0.1
         with:
           repo-token: "${{ secrets.GITHUB_TOKEN }}"
@@ -43,4 +43,4 @@ jobs:
         name: Build with Gradle and Push
         run: |
           ./gradlew jib \
-          -Djib.to.tags='latest',$GITVERSION_SEMVER
+          -Djib.to.tags='unreleased',$GITVERSION_SEMVER
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -34,6 +34,7 @@ jobs:
         name: Publish release on GitHub
         uses: softprops/action-gh-release@v1
         with:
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -40,6 +40,33 @@ There are multiple ways in which you can contribute to the project:
 1. If you are a developer and trying to start on to the project, then the suggestion is to go through the list of [issues](https://github.com/SasanLabs/VulnerableApp/issues) which contains `good first issue` which can be a good starter.
 2. If you are a developer or a security professional looking to add new Vulnerability type then you can Generate the Sample Vulnerability by running `./gradlew GenerateSampleVulnerability`. It will generate the Sample Vulnerability template which has placeholders and comments. Modified files can be seen in the logs of the command or in the github history. You can navigate to those files, fill in the placeholders and then build the project to see the effect of the changes.
 3. In case you are looking to contribute to the project by publicising it or working on the growth of the project, please feel free to add your thoughts to discussions section or issues and we can discuss over them.
+### Semantic Versioning ###
```
preetkaran20, line 43: Please update https://github.com/SasanLabs/VulnerableApp/blob/master/docs/ReleaseVulnerableApp.md file. I have removed this from Readme as it is more of an admin's work.

.github/workflows/docker.yml
```diff
@@ -31,16 +31,16 @@ jobs:
         id: gitversion
         uses: gittools/actions/gitversion/execute@v0.9.7
       -
-        name: Publish release on GitHub
-        uses: softprops/action-gh-release@v1
+        name: Create tag for GitHub
```
preetkaran20, line 34: @merry-degaga my thought is that if we do a patch then we should only create tag and update the unreleased tag of docker image but if we do minor or major release then we should create a release, update the docker image tag of latest as well as unreleased.


example: say in one scenario: we want to do a small change to our project and that small change should only create a newer github tag, same tag for docker image and update the unreleased tag.

in another scenario, we want to do a minor or major change or release all the earlier patch changes so now we want a newer github tag, a newer github release, update the latest docker image tag as well as unreleased docker image tag.

is that possible? we are good even if we have another workflow that gets triggered for 2nd scenario.

merry-degaga: Okay I did find two ways of doing this, which would be most preferred? The first method will  automate the docker build/deploy for a manually published release. The second way will automate release creation and docker build/deploy but needs to be manually triggered. In either case, a new workflow yml file will be created.

Workflow 1
1. Auto triggers when a release is manually published 
2. Using github context, it will get the tag attached to the published release
3. This tag and the 'latest' tag attached to a built and deployed docker image

Workflow 2:
1. Manual trigger workflow using workflow_dispatch
2. Uses GitHub action https://github.com/softprops/action-gh-release to automate publish of release
3. This will automate the name creation and setting the tag_name for the release and publish the release
4. Docker image will be built and deployed with 'latest' tag and tag_name
5. Description for the release can be manually edited later

preetkaran20: I think workflow 2 should work for us.

merry-degaga:
Updated Workflows:

First workflow handles semantic versioning, creates a repository tag using the calculated version, building and deploying to DockerHub with the 'unreleased' tag and calculated version.

Second workflow automates the release, it is manually triggered with optional argument for the release notes, it gets the latest repository tag, publishes a release on GitHub, then uses the lates tag and 'latest' tag to build and push to DockerHub. 

Updated ReleaseVulnerableApp.md. 

Updated build.gradle since workflow Djib.to.tags are considered additional tags and 'latest' tag was still getting appended to the unreleased docker image. Resolved by placing Djib.to.image in workflows and setting unrleased and latest tag where appropriate.

If this looks good here, I can do a PR for the VulnerableApp-facade issue too.

Thanks

build.gradle
```diff
@@ -65,9 +65,6 @@ jib {
 	from {
     	  image = 'openjdk:8-jre-alpine'
   	}
-  	to {
```
preetkaran20, line 68: if we run `./gradlew jibDockerBuild` then it will fail if we remove this.

.github/workflows/create-release.yml
```diff
@@ -0,0 +1,48 @@
+# This workflow automates release creation and builds and deploys to DockerHub
+name: Automates release gets latest tag, creates release, builds and push to DockerHub with latest tag
```
preetkaran20, line 2: Name can be little simpler: `Release project` and the above comment is already explaining its working.

.github/workflows/docker.yml
```diff
@@ -1,5 +1,5 @@
-# This workflow will build a Java project with Gradle and push a new docker image to docker hub
-name: Build and Push Image to DockerHub
+# This workflow will build a Java project with Gradle, create repository tag, and push a new docker image to docker hub
+name: Create Tag, Build and Push Unreleased Image to DockerHub
```
preetkaran20, line 2: Name can be simpler as  `Create Tag and Push Image to DockerHub`

.github/workflows/docker.yml
```diff
@@ -31,16 +31,17 @@ jobs:
...
       -
         name: Grant execute permission for gradlew
         run: chmod +x gradlew
       -
         name: Build with Gradle and Push
         run: |
           ./gradlew jib \
-          -Djib.to.tags='latest',$GITVERSION_SEMVER
\ No newline at end of file
+          -Djib.to.image=sasanlabs/owasp-vulnerableapp:unreleased \
```
preetkaran20, line 46: can we just update the tag here and to.image it can get from build.gradle

.github/workflows/docker.yml
```diff
@@ -31,16 +31,17 @@ jobs:
...
+          -Djib.to.tags=$GITVERSION_SEMVER
```
preetkaran20, line 47: Here we can make the tag as:

```-Djib.to.tags=$GITVERSION_SEMVER,unreleased```

File: README.md [hunk 8]
```diff
@@ -40,6 +40,33 @@ There are multiple ways in which you can contribute to the project:
...
+Leveraging GitHub workflow and actions, semantic versioning is automated.
```
preetkaran20, line 44: we don't need this here as the same information is there in `docs/ReleaseVulnerableApp.md`

docs/ReleaseVulnerableApp.md
```diff
@@ -17,8 +19,18 @@ Examples of version change considering the current version is 1.10.0:
...
+   Versioning
+   Specifiction](https://semver.org/)
+
+The second workflow automates release creation. Using <code>workflow_dispatch</code>, this will get the latest 
+repository tag, create a GitHub release, build and deploy as latest to DockerHub. When initializing this workflow, 
+there is an option to provide release notes. 
+
+To initialize the workflow, from the repository:
+1. Select the 'Actions' tab.
+2. Under 'Workflows' select 'Automates release gets latest tag, creates release, builds and push to DockerHub with
```
preetkaran20, line 32: Need to update the name here after changing the name of workflow.

build.gradle
preetkaran20: Also add `to.tags` as `unreleased` here such that when we run  ``` ./gradlew jibDockerBuild ``` then it will build unreleased image in local which we will reference in docker-compose.yml

preetkaran20 (commented):
Overall LGTM. Just some minor comments.

.github/workflows/docker.yml
merry-degaga: It was like that before but Djib.to.tags [means additional tags](https://github.com/GoogleContainerTools/jib/tree/master/jib-maven-plugin#build-your-image).
When you build the image jib does not see these as to.tags so it will automatically append the 'latest' tag to the image then the additional tags. The avoid this behavior, either in the build.gradle or in Djib.to.image, use sasanlabs/owasp-vulnerableapp:unreleased. 

Found this issue with jib [here](https://github.com/GoogleContainerTools/jib/issues/2237). 

Using -Djib.to.tags=$GITVERSION_SEMVER,unreleased
would still append latest tag to the docker image.

build.gradle
merry-degaga: Please see comment below. Jib documentation states to.tags means additional tags. So if no initial tag is set in the image, then latest tag will automatically be added to the image. Additional tags do not prevent this behavior. 

I did find a workaround through, I can still have the image in the build.gradle file. But in the workflow still use to.image to overwrite the build.gradle image. Will commit that.

preetkaran20 (approved):
Amazing work @merry-degaga  LGTM. Thank you !!!!

Issue #445, opened 2023-09-15
Fix SSRFVulnerabilityTest to mock the network calls.


preetkaran20:
**Is your feature request related to a problem? Please describe.**
SSRF vulnerability test is doing the network calls while running tests and causing failures. Reported issue: https://github.com/SasanLabs/VulnerableApp/issues/443

**Describe the solution you'd like**
We would like to mock the network call such that these failures can be handled.

@bperry-mf @000panther

Vijay-948:
can you assign me?

tkomlodi:
@Vijay-948, I've already have a fix if you haven't gotten started yet. If you prefer, I can also just send you what I have. 

@preetkaran20, I should have probably signaled before getting started. In either case, I can contribute it or just throw it away.

Thanks

Vijay-948:
"No problem at all. I appreciate your willingness to contribute. Please go ahead.

preetkaran20:
Thanks a lot for the PR @tkomlodi. I have merged the changes.

tkomlodi:
@preetkaran20, thanks for accepting the fix! I've been using OWASP tools for many years so I'm happy I can give back in a very modest way. I'll see if I can pick up some more issues.
Thanks

preetkaran20:
@tkomlodi thanks and sure, please go ahead.

PR #447, opened 2023-10-27
Mocked network calls made in SSRFVulnerabilityTest.

tkomlodi:
Re issue #445.

The fix simply mocks the two network requests that happen from the SSRFVulnerabilityTest test. I tried to keep the changes to a minimum, especially to the SSRFVulnerability class. Here I simply moved the network call out to a new getResponseForURLConnection() call, mostly unchanged. This is the method that is mocked from the test.

I ran into some strange Mockito spy behavior which forced the creation of the interim ssrfSpy variable.

Thanks!


preetkaran20 (approved):
LGTM, thanks a lot for the PR.

PR #450, opened 2023-11-08
Tests for error based sql injection

13Anthony:
I added tests for ErrorBasedSQLInjectionVulnerability


src/main/java/org/sasanlabs/service/vulnerability/sampleVulnerability/SampleVulnerability.java
```diff
@@ -0,0 +1,108 @@
+package org.sasanlabs.service.vulnerability.sampleVulnerability;
```
preetkaran20, line 1: Please don't checkin sample vulnerability related files. Check in only the test file.

Issue #406, opened 2022-10-01
Add 2-3 secured implementations for Union SQL Injection


preetkaran20:
**Is your feature request related to a problem? Please describe.**
As main target for Vulnerable App is scanners so we need to find out false positives by Scanners such that they can improve on there scan rules. So in order to find out the false positives, we need to add more secured implementations.

**Describe the solution you'd like**
Add 2-3 more functions in https://github.com/SasanLabs/VulnerableApp/blob/master/src/main/java/org/sasanlabs/service/vulnerability/sqlInjection/UnionBasedSQLInjectionVulnerability.java which are secure.

karthik9393:
Can I work on this issue?

preetkaran20:
Hi @karthik9393 ,

Assigned it to you.

Thanks

preetkaran20:
Hi @karthik9393 

are you working on this issue?

PR #452, opened 2023-11-14
#406 secured implementations for Union SQL Injection

x7Git:
When using JPA with CarInformation-Entity id must always be parsed to an integer. This leads to an NumberFormatException  querying the entity-id field with an injection string. To ensure the implementations are safe the sql injection id should be replaced by an string field e.g. name.


preetkaran20 (approved):

preetkaran20:
@x7Git Thanks a lot for the PR. Looks good to me. Merging it now.

Issue #449, opened 2023-11-02
UnrestrictedFileUpload: uploaded file is not accessible when root directory "/" is not writable 


tkomlodi:
**Describe the bug**
When running the application using gradle on a linux system, which does not allow writing the root directory by the app, files will be uploaded into a temp folder which is not accessible subsequently.

**To Reproduce**
1. Start application using "./gradlew bootRun" on a linux system which does not allow creating new directories by the application in the root / folder. 
2. Go to the UnrestrictedFileUpload/LEVEL_1 vulnerability and upload a file.
3. Try to access the uploaded file using the suggested path: VulnerableApp/upload/<file_name>
4. This will generate a "not found" error.

**Expected behavior**
The above path should download the file.

**Additional context**
As far as I can tell, this issue is related to the fix for #256 and #255: 
https://github.com/SasanLabs/VulnerableApp/pull/256/commits/5f273b48ba1119a349c893bbf4550a383da51181 
The original issue was caused by a FileSystemNotFoundException exception that prevented the "root" property from initializing. In my environment, the current code successfully initializes and creates the "root" (static) path and folder, but fails creating the "contentDispositionRoot" folder. This triggers an exception which causes the "root" path being replaced with a tmp folder.
The file upload succeeds to the ("root") tmp folder, however, this location is not available for download. 

**Would you like to help fix this issue?**
I'm not completely clear how the current code is intended to behave when there is no filesystem (FileSystemNotFoundException), however, I have a tentative fix that works for my environment, and also when running the app as a jar file.

If the issue makes sense and is accepted, please assign it to me and I'll create a pull request with the fix and try to give more context at that point. The proposed fix only changes two lines.

preetkaran20:
Hi @tkomlodi , 

Thanks for detailed issue. Yes, please raise the fix with the PR. We can look into it. 

Thanks, 
Karan

preetkaran20:
Closing this issue as being addresses #454

PR #453, opened 2023-11-14
Fixed file upload directory creation when system root directory is not writable by application. #449

tkomlodi:
Updated code will set `root` and `contentDispositionRoot` to a temporary folder if they weren't created for any reason. If they were created successfully, they won't be changed.


preetkaran20 (approved):
LGTM, thanks a lot for this change.

Issue #351, opened 2022-02-20
Adding new level in Unrestricted File upload which doesn't have a check on size of file uploaded


preetkaran20:
**Is your feature request related to a problem? Please describe.**
Currently there is no level in Unrestricted File Upload which doesn't have a check on size of file uploaded.

**Describe the solution you'd like**
Adding a new level in [Unrestricted File Upload](https://github.com/SasanLabs/VulnerableApp/blob/master/src/main/java/org/sasanlabs/service/vulnerability/fileupload/UnrestrictedFileUpload.java#L162) with no size restriction such that it can cause DDOS in the vulnerable app. 
Springboot has a default limit so we need to disable that limit using the following link: https://stackoverflow.com/questions/40271484/spring-upload-file-size-limit.
If we disable using application.properties then it will be applicable to all the levels but we don't want to disable to all the levels.

Adding new level is very easy, you just copy a level from above and then add the details to it. For information on what does each annotation/property does, look at: https://github.com/SasanLabs/VulnerableApp/blob/master/src/main/resources/sampleVulnerability/sampleVulnerability/SampleVulnerability.java#L86 javadocs and also https://sasanlabs.github.io/VulnerableApp/DesignDocumentation.html#design.

preetkaran20:
if it is complex to implement in springboot then i would suggest to add this in PHP as it is quite simple. Have a look at: https://github.com/SasanLabs/VulnerableApp-php

ehizman:
I will like to work on this also

preetkaran20:
@ehizman please look at https://github.com/SasanLabs/VulnerableApp/issues/350 for more information.

ehizman:
Noted @preetkaran20

tkomlodi:
@preetkaran20, could you assign this to me? I think I have a relatively clean solution. 

In short, I'd overwrite the spring MultipartFilter bean to provide a custom MultipartResolver implementation for a specific controller path only. All other paths would keep using the default spring resolver.

Besides this, it would only need a new apache dependency to provide the basis for the custom implementation:
`implementation group: 'commons-fileupload', name: 'commons-fileupload', version: '1.5'`.

The bean definition would be something like this:

```
@Bean
@Order(0)
public MultipartFilter multipartFilter() {
    class MF extends MultipartFilter {
        @Override
        protected MultipartResolver lookupMultipartResolver(HttpServletRequest request) {
[...]

PR #454, opened 2023-11-17
New unrestricted file upload size vulnerability (#351)

tkomlodi:
As proposed earlier, I added a new level 10 endpoint to UnrestrictedFileUpload which accepts unlimited sized files. These files are stored in-memory on the heap to allow triggering an out-of-memory condition.

VulnerableAppConfiguration was updated to remove the default file size limit for the new endpoint only.

Two new maven dependencies were added:
1. commons-fileupload: facilitates the configuration change
2. spring-test: allows mocking multipart files in the new unit test

I didn't add the new labels to the non-English i18n files. Let me know if that is required.

The unit test does not test the actual enforcement (or lack of) the file size limit. That could only be done with an integration test which I did not want to introduce. I manually tested that the limit is removed from the new endpoint but it still applies to all the other ones.

This change contains the prior fix for #449 (https://github.com/SasanLabs/VulnerableApp/pull/453). This is the only way I can run the application in my environment. I hope that, if that is merged first, it won't cause any issues. However, I'm not 100% sure how github will react. 

Thanks!


src/main/java/org/sasanlabs/configuration/VulnerableAppConfiguration.java
```diff
@@ -123,4 +135,29 @@ public JdbcTemplate applicationJdbcTemplate(
...
+
+    /**
+     * Customized MultipartFilter bean disables default max upload size for multipart files and
+     * their overall requests, for select paths. See {@link
+     * UnrestrictedFileUpload#getVulnerablePayloadLevel10()} for usage.
+     */
+    @Bean
+    @Order(0)
+    public MultipartFilter multipartFilter() {
+        class CustomMF extends MultipartFilter {
```
preetkaran20, line 147: Please rename as per the working of the filter.

src/main/java/org/sasanlabs/configuration/VulnerableAppConfiguration.java
```diff
@@ -123,4 +135,29 @@ public JdbcTemplate applicationJdbcTemplate(
...
+            @Override
+            protected MultipartResolver lookupMultipartResolver(HttpServletRequest request) {
+                if (MAX_FILE_UPLOAD_SIZE_OVERRIDE_PATHS.contains(request.getServletPath())) {
```
preetkaran20, line 150: This way even simple url with CONTROLLER_PATH will match this filter. Does checking other way around be better?

build.gradle
```diff
@@ -129,6 +129,8 @@ dependencies {
     // https://mvnrepository.com/artifact/org.assertj/assertj-core
     testImplementation group: 'org.assertj', name: 'assertj-core', version: '3.17.2'
 
+    testImplementation group: 'org.springframework', name: 'spring-test'
```
preetkaran20, line 132: do we really need this? we are not using spring context in the code. I think this will make the project heavier

src/main/java/org/sasanlabs/service/vulnerability/fileupload/UnrestrictedFileUpload.java
```diff
@@ -92,10 +96,10 @@ public UnrestrictedFileUpload() throws IOException, URISyntaxException {
                     "If you are running vulnerableApp as a Jar then UnrestrictedFileUpload will not work. "
                             + "For more information: https://github.com/SasanLabs/VulnerableApp/issues/255",
                     e);
-            if (root != null) {
+            if (root == null || !root.toFile().exists()) {
```
preetkaran20, line 99: will we be able to retrieve the files from temp location after this change?

build.gradle
preetkaran20: For mocking multipart file, I think mockito should help by mocking the interface and it will dynamically create the implementation and that way you can return the values by using Mockito.when ... then construct.

src/main/java/org/sasanlabs/configuration/VulnerableAppConfiguration.java
tkomlodi: I'm not sure I understand this comment. The servlet path must exactly match
`"/" + UnrestrictedFileUpload.CONTROLLER_PATH + "/" + LevelConstants.LEVEL_10`
for condition to evaluate to true.

build.gradle
tkomlodi: Good point. I introduced spring-test when I was trying to make the integration test work. When I moved on from that, I didn't think of simply going back to mockito.

src/main/java/org/sasanlabs/service/vulnerability/fileupload/UnrestrictedFileUpload.java
tkomlodi: Good question. What I aimed for and tested is that my change does not impact the original behavior. E.g. I tested the app running as a jar, and also running with `./gradlew bootRun`. (I cannot use docker in my current setup.) In both cases the behavior stayed unchanged. However, I'm not 100% clear on how this is supposed to work. I don't see a way in the existing code to return files saved to a temp `root` location. When the `root` location stays a static location within the published web file structure, this gets served as all the other static files. However, when `root` gets modified to a tmp location, there is no way to load saved files from there. `PreflightController` does not have logic to serve files from a temp `root`.

I didn't enter a bug for this as I'm not 100% sure that I'm not misreading the code and this wasn't the focus of my work. But there might be an existing bug in there. My change should not impact this either way, the original logic should be unchanged.

src/main/resources/i18n/messages_en_US.properties
```diff
@@ -114,6 +114,7 @@ Important Links:<br/>\
 </ol>
 
 #### Attack Vector Description
+UNRESTRICTED_FILE_UPLOAD_UNCONTROLLED_RESOURCE_CONSUPTION=Maximum uploaded file size is not limited.
```
preetkaran20, line 117: please add this to other locale files as well.

src/main/java/org/sasanlabs/service/vulnerability/fileupload/UnrestrictedFileUpload.java
```diff
@@ -379,4 +383,28 @@ public ResponseEntity<GenericVulnerabilityResponseBean<String>> getVulnerablePay
...
+    @VulnerableAppRequestMapping(
+            value = LevelConstants.LEVEL_10,
+            htmlTemplate = "LEVEL_1/FileUpload",
+            requestMethod = RequestMethod.POST)
+    public ResponseEntity<GenericVulnerabilityResponseBean<String>> getVulnerablePayloadLevel10(
+            @RequestParam(REQUEST_PARAMETER) MultipartFile file) throws IOException {
+        // stores files in heap memory indefinitely to allow triggering OutOfMemoryError
+        heapMemoryFileStore.add(file.getBytes());
+        return new ResponseEntity<GenericVulnerabilityResponseBean<String>>(
+                new GenericVulnerabilityResponseBean<String>("File accepted.", true),
```
preetkaran20, line 403: we should return the path of file instead of returning `File Accepted` like we are doing this for other levels.

src/main/java/org/sasanlabs/service/vulnerability/fileupload/UnrestrictedFileUpload.java
```diff
@@ -379,4 +383,28 @@ public ResponseEntity<GenericVulnerabilityResponseBean<String>> getVulnerablePay
...
                 false);
     }
+
+    @AttackVector(
+            vulnerabilityExposed = {
+                VulnerabilityType.UNCONTROLLED_RESOURCE_CONSUPTION,
+                VulnerabilityType.DENIAL_OF_SERVICE
+            },
+            description = "UNRESTRICTED_FILE_UPLOAD_UNCONTROLLED_RESOURCE_CONSUPTION",
+            payload = "UNRESTRICTED_FILE_UPLOAD_PAYLOAD_LEVEL_10")
```
preetkaran20, line 393: can you please make this level as 9 and the current level 9 as level 10 so that we can give users a gradual experience with increasing complexity.

src/main/java/org/sasanlabs/service/vulnerability/fileupload/UnrestrictedFileUpload.java
```diff
@@ -41,11 +43,13 @@
  */
 @VulnerableAppRestController(
         descriptionLabel = "UNRESTRICTED_FILE_UPLOAD_VULNERABILITY",
-        value = "UnrestrictedFileUpload")
+        value = UnrestrictedFileUpload.CONTROLLER_PATH)
 public class UnrestrictedFileUpload {
     private Path root;
     private Path contentDispositionRoot;
+    private List<byte[]> heapMemoryFileStore = new ArrayList<>();
```
preetkaran20, line 50: Instead of this, shall we use the current filesystem like we are doing in all other levels?

src/main/java/org/sasanlabs/service/vulnerability/fileupload/UnrestrictedFileUpload.java
```diff
@@ -379,4 +383,28 @@ public ResponseEntity<GenericVulnerabilityResponseBean<String>> getVulnerablePay
...
+            description = "UNRESTRICTED_FILE_UPLOAD_UNCONTROLLED_RESOURCE_CONSUPTION",
+            payload = "UNRESTRICTED_FILE_UPLOAD_PAYLOAD_LEVEL_10")
+    @VulnerableAppRequestMapping(
+            value = LevelConstants.LEVEL_10,
+            htmlTemplate = "LEVEL_1/FileUpload",
+            requestMethod = RequestMethod.POST)
+    public ResponseEntity<GenericVulnerabilityResponseBean<String>> getVulnerablePayloadLevel10(
+            @RequestParam(REQUEST_PARAMETER) MultipartFile file) throws IOException {
+        // stores files in heap memory indefinitely to allow triggering OutOfMemoryError
+        heapMemoryFileStore.add(file.getBytes());
```
preetkaran20, line 401: I would suggest using 
```
genericFileUploadUtility(
                root,
                RANDOM.nextInt() + "_" + file.getOriginalFilename(),
                validator,
                file,
                true,
                false);
```

src/main/java/org/sasanlabs/service/vulnerability/fileupload/UnrestrictedFileUpload.java
```diff
@@ -379,4 +383,28 @@ public ResponseEntity<GenericVulnerabilityResponseBean<String>> getVulnerablePay
...
+        return new ResponseEntity<GenericVulnerabilityResponseBean<String>>(
+                new GenericVulnerabilityResponseBean<String>("File accepted.", true),
+                HttpStatus.OK);
+    }
+
+    List<byte[]> getStoredFiles() {
```
preetkaran20, line 407: This is not needed.

src/main/java/org/sasanlabs/service/vulnerability/fileupload/UnrestrictedFileUpload.java
preetkaran20: please check the other levels for mechanism of storing files

src/main/java/org/sasanlabs/configuration/VulnerableAppConfiguration.java
tkomlodi: MAX_FILE_UPLOAD_SIZE_OVERRIDE_PATHS is a list of string values. I set it up this  way so later we can add to it if necessary. The check here is `contains()`, as it checks if there is a string in the list, which is equal to the path. In this context, this is really an "equals" check, with any of the values of the list. Which is, for now, just one possible value.
I can change this to be just a single value, but I believe, the behavior is already what you are suggesting.

src/main/resources/i18n/messages_en_US.properties
tkomlodi: I can add placeholders, but I don't speak the other languages, so can't add actual translations. I could do google translate, if that is acceptable.

src/main/java/org/sasanlabs/configuration/VulnerableAppConfiguration.java
preetkaran20: aauch ... My mistake.

src/main/resources/i18n/messages_en_US.properties
preetkaran20: add just in message.properties

preetkaran20 (approved):
Thanks a lot for the PR. merging it now.

preetkaran20:
Thanks @tkomlodi for the PR and presenting Idea for fixing this issue.

Issue #398, opened 2022-10-01
Add unit tests for JWT Vulnerability


preetkaran20:
**Is your feature request related to a problem? Please describe.**
We are looking at increasing the Test coverage for the project for its stability so this task is to add junit tests for https://github.com/SasanLabs/VulnerableApp/tree/master/src/main/java/org/sasanlabs/service/vulnerability/jwt

preetkaran20:
Parent issue: #212

ehizman:
Working on this Karan.
Yes, I am back!

preetkaran20:
@kjosh has merged the Unit tests for JWT. Hence closing the ticket.

PR #467, opened 2023-12-19
Add JWT unit tests

kjosh:
I wrote some unit tests for the JWT vulnerabilities (#398), testing the controller functions and some validator exploits to the best of my ability.

I refactored some relatively minor things to make the controller easier to test: 

* created a method to create a token with a JWK header in the existing `IJWTTokenGenerator`, so I can also use it in the test without too much duplication
* removed the static initialization logic in `JWTAlgorithmKMS`, manage it's singleton lifecycle via Spring instead, mainly so I can inject it into the controller and spy on it in the test
* replace the `String` type of the RequestEntity parameters with `Void`, this should be the correct type for GET-Requests and allows creation of test objects

### DIFF
--- a/src/main/java/org/sasanlabs/service/vulnerability/jwt/keys/JWTAlgorithmKMS.java
+++ b/src/main/java/org/sasanlabs/service/vulnerability/jwt/keys/JWTAlgorithmKMS.java
@@ -19,22 +19,20 @@ import java.util.Set;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.sasanlabs.internal.utility.JSONSerializationUtils;
+import org.springframework.stereotype.Component;
 
 /**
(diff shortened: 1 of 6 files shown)


src/main/java/org/sasanlabs/service/vulnerability/jwt/keys/JWTAlgorithmKMS.java
```diff
@@ -19,22 +19,20 @@
...
  *
  * @author KSASAN preetkaran20@gmail.com
  */
+@Component
 public class JWTAlgorithmKMS {
 
-    private static final Object MUTEX = new Object();
+    private Set<SymmetricAlgorithmKey> symmetricAlgorithmKeySet;
 
-    private static volatile boolean initDone = false;
```
preetkaran20, line 33: Thank you for making this class unit testable. !!!

preetkaran20 (approved):
Thanks a lot for the PR. PR looks good to me. merging it.

PR #471, opened 2024-04-30
Update PathTraversalVulnerability.java

tidaaartorhem:
Key Changes Explained:

- Enhanced security checks: Added checks to prevent Path Traversal attacks by verifying that the file name does not contain ".." and is within the list of allowed file names.

- Improved error handling: Changed the HTTP response codes to more accurately reflect the nature of the error (e.g., returning 404 Not Found for missing files and 403 Forbidden for unauthorized access attempts).

- Condition validation: Immediately returns a 400 Bad Request if the precondition is not met, which helps in quickly identifying issues with request parameters.

- Error Logging: Now logs different types of errors distinctly for better diagnostics.

### DIFF
--- a/src/main/java/org/sasanlabs/service/vulnerability/pathTraversal/PathTraversalVulnerability.java
+++ b/src/main/java/org/sasanlabs/service/vulnerability/pathTraversal/PathTraversalVulnerability.java
@@ -41,31 +41,35 @@ public class PathTraversalVulnerability {
             LogManager.getLogger(PathTraversalVulnerability.class);
 
     private static final String URL_PARAM_KEY = "fileName";
-
     private ResponseEntity<GenericVulnerabilityResponseBean<String>> readFile(
             Supplier<Boolean> condition, String fileName) {
-        if (condition.get()) {
-            InputStream infoFileStream =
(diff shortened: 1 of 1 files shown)


src/main/java/org/sasanlabs/service/vulnerability/pathTraversal/PathTraversalVulnerability.java
```diff
@@ -41,31 +41,35 @@ public class PathTraversalVulnerability {
...
-                            HttpStatus.OK);
-                } catch (IOException e) {
-                    LOGGER.error("Following error occurred: ", e);
-                }
+        if (!condition.get()) {
+            return ResponseEntity.badRequest().body(new GenericVulnerabilityResponseBean<>("Invalid request condition", false));
+        }
+
+        // Preventing Path Traversal and ensuring only allowed filenames are processed.
+        if (fileName.contains("..") || !ALLOWED_FILE_NAMES.contains(fileName)) {
```
preetkaran20, line 51: Why this change? we wanted to introduce vulnerability in the application and this will disallow ".." in file name.

Issue #486, opened 2025-09-29
geting error for version missing on image:owasp-vulnerableapp-facade


ehasan101:
compose up error showing while run
but when  add version it works` Line17: image: sasanlabs/owasp-vulnerableapp-facade:1.2.1`

PR #489, opened 2025-12-25
fix: remove deprecated Docker compose version and pin facade image to 1.2.1

Aryan-Pillai7:
Fixes warning + issue #486 in docker-compose.yml:
1. Removes obsolete [version: 3.9] field (Docker Compose has deprecated version in V2 so no need) 
2. Pinned VulnerableApp-facade image to specific version instead of floating ":latest" tag - issue fix of #486

## Manual test
- "docker-compose up" no longer shows "version is obsolete" warning
- Application runs successfully at localhost
- Both `:latest` and `:1.2.1` tags exist on Docker Hub (`docker pull` confirmed)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,4 +1,3 @@
-version: "3.9"
 services:
     VulnerableApp-base:
       image: sasanlabs/owasp-vulnerableapp:unreleased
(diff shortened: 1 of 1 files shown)


preetkaran20 (approved):
LGTM

Issue #399, opened 2022-10-01
Add unit tests for FileUpload Vulnerability


preetkaran20:
**Is your feature request related to a problem? Please describe.**
We are looking at increasing the Test coverage for the project for its stability so this task is to add junit tests for https://github.com/SasanLabs/VulnerableApp/tree/master/src/main/java/org/sasanlabs/service/vulnerability/fileupload

preetkaran20:
Parent issue: #212

shlokam:
@preetkaran20 Can you assign this issue to me?

shlokam:
@preetkaran20 For the purpose of testing, I need to update a thing. For that should I add a new dependency or can I add some code to the existing source code?

preetkaran20:
@shlokam as you wish, please raise the PR and we will discuss on that. However, which library you are looking to add and what does it helps in?

PR #493, opened 2026-01-27
Increased Junit coverage  for 'PreflightController' & 'UnrestrictedFileUpload' class 

demoralizerr:
Issue - https://github.com/SasanLabs/VulnerableApp/issues/399: Increased Junit coverage  for 'PreflightController' & 'UnrestrictedFileUpload' class


src/main/java/org/sasanlabs/service/vulnerability/fileupload/PreflightController.java
```diff
@@ -34,14 +36,16 @@ public PreflightController(UnrestrictedFileUpload unrestrictedFileUpload) {
...
-        HttpHeaders httpHeaders = new HttpHeaders();
-        httpHeaders.add(HttpHeaders.CONTENT_DISPOSITION, "attachment");
-        return new ResponseEntity<byte[]>(
-                IOUtils.toByteArray(inputStream), httpHeaders, HttpStatus.OK);
+
+        // Resolve path using Path API
+        Path filePath = unrestrictedFileUpload.getContentDispositionRoot().resolve(fileName);
+
+        // Try-with-resources ensures the stream closes automatically
+        try (InputStream inputStream = new FileInputStream(filePath.toFile())) {
```
demoralizerr, line 44: Added try-with-resources to guarantee closing of InputStream  resource

src/main/java/org/sasanlabs/service/vulnerability/sampleVulnerability/SampleVulnerability.java
```diff
@@ -0,0 +1,109 @@
+package org.sasanlabs.service.vulnerability.sampleVulnerability;
```
preetkaran20: please remove this file. This should not be there

build.gradle
```diff
@@ -136,6 +136,9 @@ dependencies {
     // https://mvnrepository.com/artifact/org.assertj/assertj-core
     testImplementation group: 'org.assertj', name: 'assertj-core', version: '3.17.2'
 
+    // https://mvnrepository.com/artifact/org.springframework.boot/spring-boot-starter-test
+    testImplementation 'org.springframework.boot:spring-boot-starter-test'
```
preetkaran20, line 140: do we need this?

demoralizerr: yes we need this for **mockmvc** classes ,  which is used in junits for PreflightController

src/main/java/org/sasanlabs/service/vulnerability/sampleVulnerability/SampleVulnerability.java
demoralizerr: removed from PR 
Since this file should never be committed or pushed, should we add it to .gitignore?

PR #499, opened 2026-02-16
Fix: Refactor JWT Level 13 to use cookie-based flow and vulnerable JWK verification

subhamkumarr:
### Description: 
This PR fixes the broken implementation of JWT Vulnerability Level 13 (Header Injection).

### Issues Addressed
- The previous implementation used unused UI parameters ("Header Name", "Header Value").
- The frontend had a hardcoded validated JWT, making the level unsolvable/unrealistic.
- The backend relied on the Authorization header instead of the standard cookie-based flow used in other levels.

### Backend changes (JWTVulnerability.java):
- Refactored getHeaderInjectionVulnerability to use RequestEntity and queryParams.
- Implemented a ?fetch=true flow to generate a valid token and set it as a JWT cookie.
- Implemented a verification flow that blindly trusts the jwk parameter in the JWT header for signature verification.

### Frontend changes (HeaderInjection_Level13.html/js):
- Replaced the input fields with standard "Fetch Token" and "Verify" buttons.
- Removed the hardcoded JWT.
- Updated the UI to display the token and verification results.

### How to Verify
1. Go to JWT Vulnerability -> Level 13.
2. Click Fetch Token. A valid JWT should be displayed and set as a cookie.
3. Click Verify. It should say "JWT is valid".
4. **Exploit:** Intercept the request, create a new JWT with your own private key, embed the corresponding public key in the jwk header, and replace the cookie. The backend verification should accept it.

### DIFF
--- a/src/main/resources/static/templates/JWTVulnerability/LEVEL_13/HeaderInjection_Level13.css
+++ b/src/main/resources/static/templates/JWTVulnerability/LEVEL_13/HeaderInjection_Level13.css
@@ -3,27 +3,43 @@
     text-align: justify;
 }
 
-#enterHeader {
+#description {
     font-size: 15px;
-    display: flex;
+    visibility: hidden;
     margin: 10px;
-    flex-direction: column;
 }
 
-#headerName, #headerValue {
-    flex: 1;
+#Note {
+    font-size: 15px;
+    margin: 10px;
+}
+
+#verificationResponse {
+    font-size: 15px;
+    margin: 10px;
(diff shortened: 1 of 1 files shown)


subhamkumarr:
done

preetkaran20:
@subhamkumarr UI still looks the same to me. Can you please fix it and validate it? Also please post the screenshot. I can see JWT is in bold characters as well as verify jwt response is in bold characters. Similarly button is not using same CSS as other levels.

Issue #500, opened 2026-02-16
Create llm based agentic workflow based vulnerable app


preetkaran20:
As agentic workflows/LLMs are becoming so popular, there comes security tools like shannon that uses LLM to find vulnerabilities. Similarly we can add some vulnerabilities of llm as well to the app.

PR #503, opened 2026-02-19
Feat: Implement LLM Injection Vulnerability (Issue #500)

subhamkumarr:
### Description
This PR addresses Issue #500 by introducing a new LLM/AI Injection vulnerability type. This allows users to practice Prompt Injection attacks in a simulated environment without requiring an actual LLM backend.

### Changes Included
- New Vulnerability Type: Added LLM_INJECTION (CWE-1059) to VulnerabilityType.java.
- Simulated AI Backend: Implemented LLMVulnerability controller with a "hidden system prompt" that can be tricked via keyword injection.
- Modern Chat UI: Added a responsive chat interface (LEVEL_1) for interacting with the bot.
- Internationalization: Added descriptive labels and attack vectors to messages.properties.
- Build Config: Updated build.gradle to exclude new JS files from Spotless checks to resolve CI environment issues.

### Verification
- Unit Tests: Added LLMVulnerabilityTest.java verifying normal chat flow and successful jailbreak scenarios.
- Manual Testing: Verified locally via Docker.

### DIFF
--- a/src/main/java/org/sasanlabs/service/vulnerability/llm/LLMVulnerability.java
+++ b/src/main/java/org/sasanlabs/service/vulnerability/llm/LLMVulnerability.java
@@ -1,0 +1,56 @@
+package org.sasanlabs.service.vulnerability.llm;
+
+import java.util.Map;
+import org.sasanlabs.internal.utility.LevelConstants;
+import org.sasanlabs.internal.utility.annotations.AttackVector;
+import org.sasanlabs.internal.utility.annotations.VulnerableAppRequestMapping;
+import org.sasanlabs.internal.utility.annotations.VulnerableAppRestController;
+import org.sasanlabs.service.vulnerability.bean.GenericVulnerabilityResponseBean;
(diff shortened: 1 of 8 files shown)


src/main/java/org/sasanlabs/service/vulnerability/llm/LLMVulnerability.java
```diff
@@ -0,0 +1,56 @@
...
+                    new GenericVulnerabilityResponseBean<>(
+                            "Hello! How can I help you today?", true),
+                    HttpStatus.OK);
+        }
+
+        String response;
+        String lowerMessage = userMessage.toLowerCase();
+
+        // Simulated LLM Logic
+        if (lowerMessage.contains("password") && !lowerMessage.contains("ignore")) {
```
preetkaran20, line 42: I think we can inject the LLM model here. you can look at VulnerableApp-facade project to see how we can add a newer docker container and then add the docker container which will run the LLM Model and we can call that model from here.

Issue #497, opened 2026-02-15
Add NOSQL Injection vulnerability.


preetkaran20:
Applications using NoSQL databases often construct queries dynamically using user input.
If input validation and sanitization are not implemented correctly, attackers may manipulate the query structure and bypass authentication or retrieve sensitive data.

This issue proposes adding a NoSQL Injection vulnerability module to VulnerableApp that demonstrates common exploitation techniques in modern applications using document-based databases such as MongoDB.

Following are some **suggested levels** but upto the developer to choose to implement as they wish.

**Vulnerability Levels**
**Level 1 — Basic Authentication Bypass**

A login API accepts username and password as JSON input and directly inserts them into a Mongo-style query.
Example vulnerable query:
db.users.find({ username: input.username, password: input.password })
An attacker can inject operators like $ne to bypass authentication.
Example payload:
```
{
  "username": { "$ne": null },
  "password": { "$ne": null }
}
```
**Level 2 — Query Manipulation via Input Fields**

User search functionality constructs a query dynamically based on user input.
Example vulnerable pattern:
db.products.find({ name: userInput })
If the application accepts JSON input directly, attackers can inject operators like $regex or $gt.
Example payload:
```{ "$regex": ".*" }```

**Level 3 — Blind NoSQL Injection**

The application returns only true/false responses instead of full query results.
Attackers must infer information using conditional operators.
Example vulnerable query:
db.users.find({ username: input })
Attackers can inject logic operators:
```{ "$regex": "^a" }```

**Level 4 — Data Extraction via Logical Operators**

The application attempts to sanitize input but allows nested query objects.
Example query:
db.accounts.find({ username: input })
Attackers inject logical operators:
```
{
 "$or": [
   { "role": "admin" },
   {}
 ]
}
```
**Level 5 — Advanced NoSQL Injection with Aggregation Pipeline**

The application uses an aggregation pipeline for analytics queries.
Example vulnerable code:
db.orders.aggregate([
  { "$match": { "user": input } }
])

Attackers inject aggregation stages.

Example payload:
```
{
 "$ne": null
}
```
Or manipulate pipeline logic to expose sensitive data.

Level 6 --- Secure implementation.

MohammedGhallab:
> @MohammedGhallab feel free to pick this vulnerability as @subhamkumarr  is no longer working on it. You can built on top of his PR as well.

[...]

PR #534, opened 2026-03-16
Add Cache Poisoning Vulnerability

### LINKED ISSUE #516: Implement Cache Poisoning Vulnerability
Add a module to demonstrate Cache Poisoning, where an attacker can inject malicious or incorrect responses into the cache. Users visiting the same page receive the poisoned content. This vulnerability teaches how caching mechanisms can be exploited if inputs aren’t properly validated.

Add 4-5 levels of this with each level secures a bit nad eventually a real secure level

luks-santos:
## Summary

Closes #516.  
                                                                                                                                                                                                                                                                                                                                                                                                                                                   This PR delivers the full level progression for the Cache Poisoning vulnerability, from the basic route-only cache key issue to the secure implementation, and adds the supporting frontend, hints, i18n, and automated tests     
  needed for the lab experience.                                                                                                                                                                                                    
                                                                                                                                                                                                                                    
  ## Implemented levels
  - **LEVEL_1**: reflected `banner` query parameter is rendered unescaped (also exploitable as stored XSS via the shared cache) while the cache key uses only the route.
  - **LEVEL_2**: naive filtering of the `banner` parameter does not fix the underlying issue because the cache key is still route-only — alternative unkeyed inputs keep the poisoning viable.                                      
  - **LEVEL_3**: the query parameter is now part of the cache key, but the page remains vulnerable through an unkeyed `X-Forwarded-Host` header used to build absolute asset URLs.                                                  
  - **LEVEL_4**: forwarded host headers are ignored, but personalized cookie-driven content (including derived email and tracked IP) is still cached publicly, demonstrating cross-user PII leakage from a shared cache.            
  - **LEVEL_5**: secure variant — ignores untrusted forwarding headers, escapes reflected content, uses a trusted asset host, and returns personalized responses with `Cache-Control: private, no-store`.


src/main/java/org/sasanlabs/service/vulnerability/cachePoisoning/CachePoisoningVulnerability.java
```diff
@@ -0,0 +1,301 @@
...
+    static final String TRUSTED_ASSET_HOST = "static.vulnerableapp.local";
+    static final String DEFAULT_DEMO_USER = "guest";
+    static final Duration DEFAULT_TTL = Duration.ofSeconds(60);
+    private static final Pattern SCRIPT_BLOCK_PATTERN =
+            Pattern.compile("(?is)<\\s*script\\b[^>]*>.*?<\\s*/\\s*script\\s*>");
+    private static final Pattern SCRIPT_TAG_PATTERN =
+            Pattern.compile("(?is)<\\s*/?\\s*script\\b[^>]*>");
+    private static final Pattern JAVASCRIPT_SCHEME_PATTERN = Pattern.compile("(?i)javascript\\s*:");
+
+    private final ConcurrentHashMap<String, CachedResponse> cache = new ConcurrentHashMap<>();
```
preetkaran20: Can you please also add a way to clear up the cache so that users can retry attacker payloads multiple times?

src/main/resources/i18n/messages.properties
```diff
@@ -368,4 +368,18 @@ LDAP_LEVEL_2_OR_BYPASS=Information Disclosure and Username Enumeration via LDAP
...
 LDAP_LEVEL_5_BLIND_INJECTION=Blind LDAP Injection leading to Authentication Bypass. The application does not display user data but reveals success or failure responses. Attackers can manipulate LDAP filters to bypass authentication or infer valid credentials.
-LDAP_LEVEL_6_SECURE=This level demonstrates a fully secure LDAP authentication implementation. \ User input is properly encoded, password verification is done using salted hashing, \ and no sensitive infformation is exposed in responses. LDAP Injection is not possible.
\ No newline at end of file
+LDAP_LEVEL_6_SECURE=This level demonstrates a fully secure LDAP authentication implementation. \ User input is properly encoded, password verification is done using salted hashing, \ and no sensitive infformation is exposed in responses. LDAP Injection is not possible.
+
+# Cache Poisoning Vulnerability
+CACHE_POISONING_VULNERABILITY=Web Cache Poisoning happens when an attacker makes a cache store an incorrect response and that response is then reused for other users. \
```
preetkaran20: Can you please add more details about real world scenario of this like Reverse Proxy, CDN they cache larger responses to deliver the results quickly and attackers can use this to poison cache?

src/main/resources/i18n/messages.properties
```diff
@@ -368,4 +368,18 @@ LDAP_LEVEL_2_OR_BYPASS=Information Disclosure and Username Enumeration via LDAP
...
+This level uses a shared in-memory cache to show the core issue: the response changes with user input, but the cache key only uses the route. \
```
preetkaran20: I would suggest that the information about vulnerability should have more details about vulnerabilities not about levels. The levels can have information like this as part of hints i.e. attack vector annotation

src/main/resources/i18n/messages.properties
```diff
@@ -368,4 +368,18 @@ LDAP_LEVEL_2_OR_BYPASS=Information Disclosure and Username Enumeration via LDAP
...
+Important Links:<br/>\
+<ol> <li> <a href="https://owasp.org/www-community/attacks/Cache_Poisoning" target="_blank">OWASP Cache Poisoning</a> \
```
preetkaran20, line 407: if possible add some youtube links as well that could help users learn more.

src/main/resources/static/templates/CachePoisoning/LEVEL_1/CachePoisoning.js
```diff
@@ -0,0 +1,89 @@
...
+  document.getElementById("cacheKey").textContent =
+    xmlHttpRequest.getResponseHeader("X-Cache-Key") || "-";
+}
+
+function updateResponseArea(content) {
+  document.getElementById("cachePoisoningResponse").innerHTML = content;
+}
+
+function sendCachePoisoningRequest(method, url) {
+  let xmlHttpRequest = new XMLHttpRequest();
```
preetkaran20: framework exposes the way to do get request. please use that method. doGetAjaxCall is the method name from vulnerableApp.js.

src/main/java/org/sasanlabs/service/vulnerability/cachePoisoning/CachePoisoningVulnerability.java
```diff
@@ -0,0 +1,301 @@
...
+
+    @AttackVector(
```
preetkaran20, line 65: we can have multiple attackvector annotations and each will become hint in the UI. Here it is also vulnerable to XSS, etc and similarly many other levels too are vulnerable to multiple vulnerabilities. i would be great of we add more attack vectors. Like

src/main/java/org/sasanlabs/service/vulnerability/cachePoisoning/CachePoisoningVulnerability.java
```diff
@@ -0,0 +1,301 @@
...
+            description = "CACHE_POISONING_LEVEL_1_ROUTE_ONLY_CACHE_KEY",
+            payload = "CACHE_POISONING_PAYLOAD_LEVEL_1")
+    @VulnerableAppRequestMapping(
+            value = LevelConstants.LEVEL_1,
+            htmlTemplate = "LEVEL_1/CachePoisoning")
+    public ResponseEntity<GenericVulnerabilityResponseBean<String>> getVulnerablePayloadLevel1(
+            @RequestParam(value = BANNER_QUERY_PARAMETER, required = false) String banner,
+            HttpServletRequest request) {
+        String responseContent = buildLevel1Response(banner);
+        return buildCachedResponse(
```
preetkaran20, line 82: lets have first level also vulnerable to XSS like level 2 is vulnerable. i.e. no escapeHtml4 method.

src/main/resources/static/templates/CachePoisoning/LEVEL_4/CachePoisoning.js
```diff
@@ -0,0 +1,88 @@
...
+    xmlHttpRequest.getResponseHeader("Cache-Control") || "-";
+  document.getElementById("varyHeader").textContent =
+    xmlHttpRequest.getResponseHeader("Vary") || "-";
+}
+
+function updateResponseArea(content) {
+  document.getElementById("cachePoisoningResponse").innerHTML = content;
+}
+
+function sendCachePoisoningRequest(method, url) {
```
preetkaran20: I feel there is a lot of things which are common in levels. Like Level 1 and Level 2 can have same JS, HTML and CSS templates. You can add a simple config for text changes in these levels and do the change via JS file. Similarly for other levels too.

Please only do this, if you think it is possible.

src/main/java/org/sasanlabs/service/vulnerability/cachePoisoning/CachePoisoningVulnerability.java
```diff
@@ -0,0 +1,301 @@
...
+        String safeBanner = StringEscapeUtils.escapeHtml4(normalizeBanner(banner));
+        String assetUrl = buildAssetUrl(resolveUntrustedForwardedHost(request));
+        return "<section class=\"cache-poisoning-response\">"
+                + "<h3>Forwarded host poisoning</h3>"
+                + "<p><strong>Banner key:</strong> "
+                + safeBanner
+                + "</p>"
+                + "<p><strong>Boot asset URL:</strong> <a href=\""
+                + assetUrl
+                + "\" target=\"_blank\" rel=\"noreferrer\">"
```
preetkaran20, line 209: I think it gives more indications to user about vulnerability if we make it load the js/html file into the context. What do you think?

src/main/java/org/sasanlabs/service/vulnerability/cachePoisoning/CachePoisoningVulnerability.java
```diff
@@ -0,0 +1,301 @@
...
+    private String buildLevel4Response(HttpServletRequest request) {
+        String safeUser =
+                StringEscapeUtils.escapeHtml4(
+                        StringUtils.defaultIfBlank(resolveDemoUser(request), DEFAULT_DEMO_USER));
+        return "<section class=\"cache-poisoning-response\">"
+                + "<h3>Personalized response in a shared cache</h3>"
+                + "<p><strong>Viewer:</strong> "
+                + safeUser
+                + "</p>"
+                + "<p><strong>Dashboard message:</strong> Welcome back, "
```
preetkaran20: I think level 4 can show more impact. please think of adding impacts and even add those things in payloads.

src/main/resources/attackvectors/CachePoisoningPayload.properties
```diff
@@ -0,0 +1,25 @@
+CACHE_POISONING_PAYLOAD_LEVEL_1=This level simulates a shared cache that stores the response only by route.</br>\
+<b>How to exploit this level?</b><br/>\
+1. Call <code>/VulnerableApp/CachePoisoning/LEVEL_1?banner=FAKE%20URGENT%20UPDATE</code> to poison the cache.<br/>\
+2. Call <code>/VulnerableApp/CachePoisoning/LEVEL_1</code> without the query parameter.<br/>\
+3. The second response is served from cache and still contains the poisoned banner.
+CACHE_POISONING_PAYLOAD_LEVEL_2=This level adds a naive filter that removes only obvious script payloads, but the cache key is still based on the route alone.</br>\
+<b>How to exploit this level?</b><br/>\
+1. Call <code>/VulnerableApp/CachePoisoning/LEVEL_2?banner=%3Cb%3EURGENT%20STATUS%20UPDATE%3C/b%3E</code> to poison the cache with harmless-looking HTML.<br/>\
```
preetkaran20, line 13: to exploit it we can say use `` and this will cause XSS. You can add more such payloads for each level.

preetkaran20 (changes requested):
@luks-santos I must say this is amazing code change. Very thorough PR. I really liked the PR completeness. I have added small comments, can you please address?

src/main/java/org/sasanlabs/service/vulnerability/cachePoisoning/CachePoisoningVulnerability.java
luks-santos: Great idea. I'll implement an iframe to load the asset directly, providing a clear visual of the cache poisoning exploit to make the vulnerability easier to understand

luks-santos:
Hi @preetkaran20 
                                                                                                                                                                                                                                    
  I've pushed the changes addressing the feedback. Summary of what was implemented:                                                                                                                                                 
                                                                                                                                                                                                                                    
  **Frontend / UX**                                                                                                                                                                                                                 
  - Moved level-specific templates into a common directory to reduce duplication (`230afec`).
  - Refactored levels 1–5 to reuse `doGetAjaxCall` and pass `xhr` to callbacks (`6049137`).                                                                                                                                         
  - Added a "reset cache" button on every level so the lab is reproducible without restarting (`3bae76d`).                                                                                                                          
                                                                                                                                                                                                                                    
  **Vulnerability depth**                                                                                                                                                                                                           
  - Made Level 1 exploitable as XSS through the poisoned cached response (`179777a`).                                                                                                                                               
  - Exposed multiple attack vectors per level with matching hints/payloads (`5afcb79`, `f0f30a9`).                                                                                                                                  
  - Enriched Level 4 PII leak (derived email + tracked IP) and tightened the overall lab (`447b47c`).                                                                                                                               
  - Improved Level 3 visualization and fixed the simulation (`ba62756`, `2673d64`).                                                                                                                                                 
                                                                                                                                                                                                                                                                                                                                                                                                                             
   The PR description has been updated to reflect the final scope. Let me know if there's anything else you'd like me to adjust!

src/main/resources/static/templates/CachePoisoning/Common/CachePoisoning.js
```diff
@@ -0,0 +1,310 @@
...
+  btn.disabled = !hasCache;
+}
+
+function fetchDataCallback(data, request) {
+  updateDiagnostics(request);
+  document.getElementById("cachePoisoningResponse").innerHTML = data.content;
+}
+
+function initLevel() {
+  const level = vulnerabilityLevelSelected || "LEVEL_1";
```
preetkaran20, line 208: This is breaking in Modern UI. Instead of using this, use below function to get the selectedLevel:
getCurrentVulnerabilityLevel(), this is present in both legacy and modern ui.

In order to test with modern ui use: https://github.com/SasanLabs/VulnerableApp#testing-with-modern-ui

src/main/resources/static/templates/CachePoisoning/Common/CachePoisoning.js
```diff
@@ -0,0 +1,310 @@
...
+      doGetAjaxCall(
+        fetchDataCallback,
+        getRequestUrl(false),
+        true,
+        getNoBrowserCacheHeaders()
+      );
+    });
+
+  // Configure Signals
+  if (config.signals.includes("cacheControl"))
```
preetkaran20, line 304: I feel like code became very complex after merging templates. It becomes very heavy on JS side. I think we should have different templates or group of templates for each level. Sorry, I felt easier initially but now it feels little difficult.

src/main/resources/static/vulnerableApp.js
```diff
@@ -225,9 +225,9 @@ function genericResponseHandler(xmlHttpRequest, callBack, isJson) {
     // XMLHttpRequest.DONE == 4
     if (xmlHttpRequest.status == 200 || xmlHttpRequest.status == 401) {
       if (isJson) {
-        callBack(JSON.parse(xmlHttpRequest.responseText));
+        callBack(JSON.parse(xmlHttpRequest.responseText), xmlHttpRequest);
```
preetkaran20, line 228: Thank you for this change. Really appreciate this.

src/main/resources/i18n/messages_en_US.properties
```diff
@@ -288,4 +288,20 @@ SSRF_VULNERABILITY_URL_IF_NOT_FILE_PROTOCOL_AND_INTERNAL_METADATA_URL=file:// pr
...
+CACHE_POISONING_LEVEL_1_XSS=Cross-Site Scripting (XSS) via Cache Poisoning. Since the banner parameter is reflected without sanitization and the cache only keys on the route, an attacker can store a malicious script in the cache that will execute for all subsequent users.
+CACHE_POISONING_LEVEL_2_NAIVE_FILTER=This level attempts to fix the issue by stripping script tags from the input, but the cache key still only uses the route. Attackers can still poison the cache with misleading content or other HTML payloads that bypass the naive filter.
+CACHE_POISONING_LEVEL_2_XSS=XSS via Cache Poisoning with Filter Bypass. The application strips obvious <script> tags, but does not sanitize other HTML attributes. Attackers can use event handlers like 'onerror' in tags like <img> to execute scripts and poison the cache.
+CACHE_POISONING_LEVEL_3_JS_INJECTION=JavaScript Injection via Cache Poisoning. Since the application trusts the X-Forwarded-Host header to build asset URLs, an attacker can poison the cache to point to a malicious domain, causing all users to load and execute external scripts.
```
preetkaran20, line 306: this seems to be not in sync with messages.properties. can you please fix it?

src/main/java/org/sasanlabs/service/vulnerability/cachePoisoning/CachePoisoningVulnerability.java
```diff
@@ -0,0 +1,402 @@
...
+            }
+        }
+        return null;
+    }
+
+    private String buildAssetUrl(String host) {
+        return "https://" + StringEscapeUtils.escapeHtml4(host) + "/assets/cache-poisoning-demo.js";
+    }
+
+    void clearCache() {
```
preetkaran20, line 399: I think this should just clear the cache of that level, not other levels, but for now it is ok as this PR is having a lot of changes. we can take it as a newer issue.

src/main/java/org/sasanlabs/service/vulnerability/cachePoisoning/CachePoisoningVulnerability.java
```diff
@@ -0,0 +1,402 @@
...
+    public ResponseEntity<GenericVulnerabilityResponseBean<String>> getVulnerablePayloadLevel1(
+            @RequestParam(value = BANNER_QUERY_PARAMETER, required = false) String banner,
+            @RequestParam(
+                            value = RESET_CACHE_QUERY_PARAMETER,
+                            required = false,
+                            defaultValue = "false")
+                    boolean resetCache,
+            HttpServletRequest request) {
+        if (resetCache) {
+            clearCache();
```
preetkaran20, line 82: resetCache flag is present in all the levels. I think we can have a newer function with spring annotations and expose just clearcache as an api call example:

 @PostMapping("/clearCache")
    public String clearCache() {
        cache.clear();
        return "Cache cleared";
    }

then the URL will become something like /vulnerableapp/cachepoisioning/clearCache. you can create a newer function in vulnerableApp.js to return base url for the vulnerability and add clearCache to it.

In case you please to add a function to VulnerableApp.js,  please add it to https://github.com/SasanLabs/VulnerableApp-facade/blob/main/facade-app/public/GlobalUtility.js as well

src/main/java/org/sasanlabs/service/vulnerability/cachePoisoning/CachePoisoningVulnerability.java
```diff
@@ -0,0 +1,402 @@
...
+    @AttackVector(
+            vulnerabilityExposed = VulnerabilityType.REFLECTED_XSS,
+            description = "CACHE_POISONING_LEVEL_1_XSS",
+            payload = "CACHE_POISONING_PAYLOAD_LEVEL_1_XSS")
+    @VulnerableAppRequestMapping(
+            value = LevelConstants.LEVEL_1,
+            htmlTemplate = "Common/CachePoisoning")
+    public ResponseEntity<GenericVulnerabilityResponseBean<String>> getVulnerablePayloadLevel1(
+            @RequestParam(value = BANNER_QUERY_PARAMETER, required = false) String banner,
+            @RequestParam(
```
preetkaran20, line 75: I tried testing clearCache but it is not working. Seems like some bug. On clicking clear cache, if i again make a victim request, it is still returning me cached values.

src/main/java/org/sasanlabs/service/vulnerability/cachePoisoning/CachePoisoningVulnerability.java
```diff
@@ -0,0 +1,402 @@
...
+    static final String DEMO_USER_DOMAIN = "vulnerableapp.local";
+
+    static final Duration DEFAULT_TTL = Duration.ofSeconds(60);
+    private static final Pattern SCRIPT_BLOCK_PATTERN =
+            Pattern.compile("(?is)<\\s*script\\b[^>]*>.*?<\\s*/\\s*script\\s*>");
+    private static final Pattern SCRIPT_TAG_PATTERN =
+            Pattern.compile("(?is)<\\s*/?\\s*script\\b[^>]*>");
+    private static final Pattern JAVASCRIPT_SCHEME_PATTERN = Pattern.compile("(?i)javascript\\s*:");
+
+    private final ConcurrentHashMap<String, CachedResponse> cache = new ConcurrentHashMap<>();
```
preetkaran20, line 60: I think we need to make it static final, else it is not working as per expected.

src/main/java/org/sasanlabs/service/vulnerability/cachePoisoning/CachePoisoningVulnerability.java
```diff
@@ -0,0 +1,402 @@
...
+    static final String RESET_CACHE_QUERY_PARAMETER = "resetCache";
+
+    static final String FORWARDED_HOST_HEADER = "X-Forwarded-Host";
+    static final String CACHE_STATUS_HEADER = "X-Cache-Status";
+    static final String CACHE_KEY_HEADER = "X-Cache-Key";
+
+    static final String EMPTY_CACHE_KEY = "-";
+    static final String CACHE_STATUS_HIT = "HIT";
+    static final String CACHE_STATUS_MISS = "MISS";
+    static final String CACHE_CONTROL_PUBLIC = "public, max-age=60";
```
preetkaran20, line 41: I think it is causing a bit of difficulty to understand. this is a double edged sword. 

I am thinking of adding extra levels where it is always fetches from the server i.e. no caching on browser layer for students and levels where browser side cache is there for scanners to find out that poisioning can be a candidate.

Any thoughts? shall we add multiple levels or add a flag within the level where users can choose browser level cache along with server level cache or just server cache?

preetkaran20 (commented):
Overall, @luks-santos this is really a great PR. I have added a few comments, I think once they are resolved, i think we can merge it quickly. Just a few thoughts on browser side caching issues which we need to work on.

src/main/resources/static/templates/CachePoisoning/Common/CachePoisoning.js
luks-santos: Done! Replaced with `getCurrentVulnerabilityLevel()` and tested with Modern UI — working fine. Thanks!

src/main/resources/static/vulnerableApp.js
luks-santos: Thank you!

src/main/resources/static/templates/CachePoisoning/Common/CachePoisoning.js
luks-santos: I agree. I think we should keep common JS/CSS shared across levels, but have a separate template structure for each level. This would reduce complexity and make the code easier to maintain.

src/main/resources/static/templates/CachePoisoning/Common/CachePoisoning.js
preetkaran20: Sorry, I can still see older code here

src/main/resources/static/templates/CachePoisoning/Common/CachePoisoning.js
preetkaran20: Would it be possible to incorporate it?

src/main/java/org/sasanlabs/service/vulnerability/cachePoisoning/CachePoisoningVulnerability.java
luks-santos: With the implementation of cacheClear per route using @PostMapping("/clearCache"), it will be possible to clear the cache by level. This is because the level will need to be passed in order to build the appropriate response for each level, so the route will receive the level as a parameter.

src/main/resources/static/templates/CachePoisoning/Common/CachePoisoning.js
luks-santos: I’m addressing the comments as I go.
I forgot to push the changes 😅
I've pushed them now, including the template refactor.
This specific change was replaced as part of the template refactor

src/main/resources/static/templates/CachePoisoning/Common/CachePoisoning.js
luks-santos: Yes, it’s already been incorporated as part of the template refactor

src/main/java/org/sasanlabs/service/vulnerability/cachePoisoning/CachePoisoningVulnerability.java
luks-santos: I'll add a helper function to centralize the base URL for the vulnerability:

```javascript
function getUrlForVulnerability() {
  return "/VulnerableApp/" + vulnerabilitySelected;
}
```

With that, I can construct the `clearCache` endpoint dynamically, including the level:

```javascript
const clearCacheUrl =
  getUrlForVulnerability() +
  "/clearCache" +
  "?level=" +
  encodeURIComponent(getCurrentVulnerabilityLevel());
```

This will result in something like:

```
/VulnerableApp/CachePoisoning/clearCache?level=LEVEL_1
```
I'll also add this helper function to `GlobalUtility.js`, as mentioned, to keep things consistent across the project.

src/main/java/org/sasanlabs/service/vulnerability/cachePoisoning/CachePoisoningVulnerability.java
luks-santos: Ok, makes sense. This ensures the cache is global to the application. Even though Spring normally manages controllers as singletons by default, I think keeping it `static final` better represents a shared cache

src/main/java/org/sasanlabs/service/vulnerability/cachePoisoning/CachePoisoningVulnerability.java
luks-santos: I made some changes to the cache behavior and tested the `clearCache` flow again. It seems to be working now, and I could not reproduce the issue anymore. If it still happens in a specific scenario, I can investigate that case  in more detail

src/main/java/org/sasanlabs/service/vulnerability/cachePoisoning/CachePoisoningVulnerability.java
luks-santos: I agree it's a real trade-off, I also ran into this same confusion while iterating on the levels.

With `Cache-Control: public, max-age=60`, the browser would also store the poisoned response, and clearing the simulated cache through the new `clear-cache ` endpoint wouldn't evict the browser copy, so would still see stale poisoned content and assume the server cache hadn't actually been cleared.

Rather than splitting each level into "with browser cache" / "without browser cache" variants, which would dilute the per-level concept (each level today teaches a distinct shape: unkeyed route, naive filter, unkeyed `X-Forwarded-Host`, public-cached personalization), I believe a cleaner option is a single toggle inside each level that switches between two cache-header modes:

- Browser + shared cache (default): `Cache-Control: public, max-age=60` 
- Shared cache only: Cache-Control: `public, s-maxage=60, max-age=0`

The key piece is `s-maxage`. It instructs shared caches (CDNs, reverse proxies) to retain the response for 60 seconds, while telling the browser that the response is already stale (max-age=0) and must be revalidated on every use. The simulated cache and the X-Cache-Status: HIT/MISS headers stay identical across both modes.

I think this way scanners would still be able to detect this vulnerability. What do you think? Can I move forward with an implementation along these lines, or do you have another suggestion in mind?

preetkaran20: Yes makes sense @luks-santos .Please go ahead.

luks-santos:
Hi @preetkaran20,

I've pushed the changes addressing the feedback. Here's a principal summary of what was implemented:

## Cache reset reliability

* Moved cache clearing into a single `POST /clearCache?level=<level>` endpoint. This ensures that resetting one level no longer flushes the entire shared cache, each reset now evicts only entries for the targeted level (5d02696).

## Browser cache toggle (Levels 1–4)

* Added a **Browser cache** checkbox to the Controls block on each vulnerable level (4178173).
* When **OFF**, the server emits:

  ```
  Cache-Control: public, s-maxage=60, max-age=0
  ```
  instead of:

  ```
  Cache-Control: public, max-age=60
  ```

This ensures the browser always revalidates while the shared cache layer (the one being poisoned) keeps its 60-second TTL.

* I think `s-maxage` directive preserves scanner detection.

* Level 5 (secure variant) remains unchanged.

---
Let me know if there's anything else you'd like me to adjust!

I've opened the PR at https://github.com/SasanLabs/VulnerableApp-facade/pull/112

preetkaran20 (approved):
Amazing job @luks-santos One of the best change.

preetkaran20:
Merged the PR @luks-santos Amazing work!!!!!

preetkaran20:
🎉 Thanks for contributing @luks-santos!
   https://docs.google.com/forms/d/e/1FAIpQLSfwWVdnULUhtfruA-DN328NwKnBGebaWg9U5y0xivLLxxoMog/viewform?usp=pp_url&entry.1414771947=luks-santos
   
👉 [#497 - Add NOSQL Injection vulnerability.](https://github.com/SasanLabs/VulnerableApp/issues/497)

preetkaran20:
@all-contributors please add @luks-santos for code

PR #593, opened 2026-04-22
Create SECURITY.md for security policy

bogdansharp:
Added a security policy document outlining supported versions and vulnerability reporting.


preetkaran20 (changes requested):
We are already a vulnerableapp so vulnerability reporting doesn't apply to us.

Issue #524, opened 2026-03-13
In the deprecated UI, there is a small flicker which is impacting the UI while navigating from one lebvel to another


preetkaran20:
When navigating between pages in the deprecated UI, parts of the layout briefly repaint, causing a visible flicker. This happens because dynamic content is injected or updated via JavaScript after the initial paint. Rendering should be adjusted so page content is prepared before first paint, ensuring users only see the final styled state and eliminating the flicker.

nirupam-06:
Hi! I'd like to work on this issue. Could you please assign it to me? Thanks!

nirupam-06:
"Opened a fix here: https://github.com/SasanLabs/VulnerableApp/pull/707

PR #596, opened 2026-04-24
Enhance Cryptographic failures module

aashpis:
Reordered and added new levels, now 1-11, to the Cryptographic failures module according to issue 523. 

Additional methods have been added to `PasswordHashingUtils` for new levels

`EncryptionUtils` was created as part of the `internal.utility` package and contains methods related to encryption.

Descriptions for new levels were added to `message.properties` 

**New Level Order:**

1. plaintext storage
2. Base64 encoding is not encryption
3. Caesar Cipher
4. Security By Obscurity/Custom logic
5. MD4 Hash
6. MD5 Hash
7. SHA-1 Hash
8. LM Hash
9. Unsalted SHA-256 
10. AES with weak key
11. Secure example: Bcrypt

## Release Notes

* **New Features**
  * Added Cryptographic Failures vulnerability module with 11 progressive challenge levels covering plaintext storage, weak encryption methods, and obsolete hashing algorithms.
  * Implemented automatic database seeding on application startup.

* **Bug Fixes**
  * Corrected label typo in error messages.

[Review Change Stack](https://app.coderabbit.ai/change-stack/SasanLabs/VulnerableApp/pull/596?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


src/main/java/org/sasanlabs/service/vulnerability/cryptographicFailures/CryptographicFailuresVulnerability.java
```diff
@@ -36,41 +38,266 @@ public class CryptographicFailuresVulnerability {
...
-
-    static {
-        try {
-            MessageDigest digest = MessageDigest.getInstance("SHA-256");
-            digest.update(LEVEL5_SALT.getBytes());
-            byte[] hash = digest.digest(LEVEL5_SECRET.getBytes());
-            LEVEL5_HASH = bytesToHex(hash);
-        } catch (NoSuchAlgorithmException e) {
-            throw new RuntimeException("SHA-256 not available", e);
+    private static final String LEVEL1_SECRET = "letmein";
```
preetkaran20: I think we should create a table and insert secrets in the table and let user find it out by looking at the table. The challenge should not say what is the secret like in our case level 1 is showing the password. Look at Authentication Vulnerability, we just want to tell user to find out the secret by looking at database. Example:

Image

src/main/java/org/sasanlabs/service/vulnerability/cryptographicFailures/CryptographicFailuresVulnerability.java
```diff
@@ -36,41 +38,266 @@ public class CryptographicFailuresVulnerability {
...
+            htmlTemplate = "LEVEL_1/CryptographicFailures")
+    public ResponseEntity<GenericVulnerabilityResponseBean<String>> getVulnerablePayloadLevel1(
+            @RequestParam Map<String, String> queryParams) {
+        String password = queryParams.get(PASSWORD_PARAM);
+
+        if (password == null || password.isEmpty()) {
+            // Vulnerable: password is exposed in plaintext in the API response
+            return new ResponseEntity<>(
+                    new GenericVulnerabilityResponseBean<>(
+                            "CHALLENGE: The system stores passwords in plaintext."
```
preetkaran20, line 62: This challenge should be that password is stored as plain text, look at database to findout what is the password and crack it.

src/main/java/org/sasanlabs/service/vulnerability/cryptographicFailures/CryptographicFailuresVulnerability.java
```diff
@@ -36,41 +38,266 @@ public class CryptographicFailuresVulnerability {
...
+            value = LevelConstants.LEVEL_2,
+            htmlTemplate = "LEVEL_1/CryptographicFailures")
+    public ResponseEntity<GenericVulnerabilityResponseBean<String>> getVulnerablePayloadLevel2(
+            @RequestParam Map<String, String> queryParams) {
+        String password = queryParams.get(PASSWORD_PARAM);
+
+        if (password == null || password.isEmpty()) {
+            return new ResponseEntity<>(
+                    new GenericVulnerabilityResponseBean<>(
+                            "CHALLENGE: The system 'encrypts' passwords using Base64 encoding."
```
preetkaran20: dont tell it is base64 encoded, let user find it out. The hint should tell it is base64 encoded. i.e.

Image

src/main/java/org/sasanlabs/service/vulnerability/cryptographicFailures/CryptographicFailuresVulnerability.java
```diff
@@ -36,41 +38,266 @@ public class CryptographicFailuresVulnerability {
...
+            value = LevelConstants.LEVEL_3,
+            htmlTemplate = "LEVEL_1/CryptographicFailures")
+    public ResponseEntity<GenericVulnerabilityResponseBean<String>> getVulnerablePayloadLevel3(
+            @RequestParam Map<String, String> queryParams) {
+        String password = queryParams.get(PASSWORD_PARAM);
+
+        if (password == null || password.isEmpty()) {
+            return new ResponseEntity<>(
+                    new GenericVulnerabilityResponseBean<>(
+                            "CHALLENGE: A user's password is encrypted with a Caesar cipher:"
```
preetkaran20: Dont tell it is caesar cipher in challenge. Challenge is finding out password. Hint should tell it is Caesar cipher

src/main/java/org/sasanlabs/service/vulnerability/cryptographicFailures/CryptographicFailuresVulnerability.java
```diff
@@ -36,41 +38,266 @@ public class CryptographicFailuresVulnerability {
...
-            LEVEL5_HASH = bytesToHex(hash);
-        } catch (NoSuchAlgorithmException e) {
-            throw new RuntimeException("SHA-256 not available", e);
+    private static final String LEVEL1_SECRET = "letmein";
+
+    private static final String LEVEL2_SECRET = "s3cretKey";
+    private static final String LEVEL2_ENCODED =
+            Base64.getEncoder().encodeToString(LEVEL2_SECRET.getBytes());
+
+    private static final String LEVEL3_SECRET = "helloworld";
```
preetkaran20: Is it possible to make the secrets little more complex ?

src/main/java/org/sasanlabs/service/vulnerability/cryptographicFailures/CryptographicFailuresVulnerability.java
```diff
@@ -36,41 +38,266 @@ public class CryptographicFailuresVulnerability {
...
+    private static final int LEVEL3_CHAR_SHIFT = 3;
+    private static final String LEVEL3_ENCRYPTED =
+            EncryptionUtils.caesarCipher(LEVEL3_SECRET, LEVEL3_CHAR_SHIFT);
+
+    private static final String LEVEL4_SECRET = "obscuredpassword";
```
preetkaran20: make it a bit complex

src/main/java/org/sasanlabs/service/vulnerability/cryptographicFailures/CryptographicFailuresVulnerability.java
```diff
@@ -36,41 +38,266 @@ public class CryptographicFailuresVulnerability {
...
+            htmlTemplate = "LEVEL_1/CryptographicFailures")
+    public ResponseEntity<GenericVulnerabilityResponseBean<String>> getVulnerablePayloadLevel4(
+            @RequestParam Map<String, String> queryParams) {
+        String password = queryParams.get(PASSWORD_PARAM);
+
+        // No password param: return the challenge hash
+        if (password == null || password.isEmpty()) {
+            return new ResponseEntity<>(
+                    new GenericVulnerabilityResponseBean<>(
+                            "CHALLENGE: A user's password is stored using custom logic: "
```
preetkaran20, line 197: This is great !!! we need challenge for user.

src/main/java/org/sasanlabs/service/vulnerability/cryptographicFailures/CryptographicFailuresVulnerability.java
```diff
@@ -36,41 +38,266 @@ public class CryptographicFailuresVulnerability {
...
+            htmlTemplate = "LEVEL_1/CryptographicFailures")
+    public ResponseEntity<GenericVulnerabilityResponseBean<String>> getVulnerablePayloadLevel5(
+            @RequestParam Map<String, String> queryParams) {
+        String password = queryParams.get(PASSWORD_PARAM);
+
+        // No password param: return the challenge hash
+        if (password == null || password.isEmpty()) {
+            return new ResponseEntity<>(
+                    new GenericVulnerabilityResponseBean<>(
+                            "CHALLENGE: A user's password is stored as MD4 hash: "
```
preetkaran20, line 242: great !!!

src/main/java/org/sasanlabs/service/vulnerability/cryptographicFailures/CryptographicFailuresVulnerability.java
```diff
@@ -36,41 +38,266 @@ public class CryptographicFailuresVulnerability {
...
+
+    private static final String LEVEL3_SECRET = "helloworld";
+    private static final int LEVEL3_CHAR_SHIFT = 3;
+    private static final String LEVEL3_ENCRYPTED =
+            EncryptionUtils.caesarCipher(LEVEL3_SECRET, LEVEL3_CHAR_SHIFT);
+
+    private static final String LEVEL4_SECRET = "obscuredpassword";
+    private static final String LEVEL4_ENCRYPTED = EncryptionUtils.customCipher(LEVEL4_SECRET);
+
+    private static final String LEVEL5_SECRET = "password123";
```
preetkaran20: make secret really strong in this case.

src/main/java/org/sasanlabs/service/vulnerability/cryptographicFailures/CryptographicFailuresVulnerability.java
```diff
@@ -36,41 +38,266 @@ public class CryptographicFailuresVulnerability {
...
+    private static final String LEVEL5_HASH = PasswordHashingUtils.md4Hash(LEVEL5_SECRET);
+
+    private static final String LEVEL6_SECRET = "password123";
```
preetkaran20: Each level password should be different from other levels.

src/main/java/org/sasanlabs/service/vulnerability/cryptographicFailures/CryptographicFailuresVulnerability.java
```diff
@@ -149,182 +377,199 @@ public ResponseEntity<GenericVulnerabilityResponseBean<String>> getVulnerablePay
...
         if (password == null || password.isEmpty()) {
-            // Vulnerable: password is exposed in plaintext in the API response
             return new ResponseEntity<>(
                     new GenericVulnerabilityResponseBean<>(
-                            "CHALLENGE: The system stores passwords in plaintext. "
-                                    + "The stored password for user 'admin' is: "
-                                    + LEVEL3_SECRET
-                                    + " — Enter it below to authenticate!",
-                            false),
+                            "SECURE: This password is hashed with LM. Hash: "
```
preetkaran20: Should be CHALLENGE not SECURE

src/main/java/org/sasanlabs/service/vulnerability/cryptographicFailures/CryptographicFailuresVulnerability.java
```diff
@@ -36,41 +38,266 @@ public class CryptographicFailuresVulnerability {
...
+    private static final String LEVEL5_SECRET = "password123";
+    private static final String LEVEL5_HASH = PasswordHashingUtils.md4Hash(LEVEL5_SECRET);
+
+    private static final String LEVEL6_SECRET = "password123";
+    private static final String LEVEL6_HASH = DigestUtils.md5Hex(LEVEL6_SECRET);
+
+    private static final String LEVEL7_SECRET = "admin";
+    private static final String LEVEL7_HASH = DigestUtils.sha1Hex(LEVEL7_SECRET);
+
+    private static final String LEVEL8_SECRET = "admin123";
```
preetkaran20: make it little complex secret

src/main/java/org/sasanlabs/service/vulnerability/cryptographicFailures/CryptographicFailuresVulnerability.java
```diff
@@ -36,41 +38,266 @@ public class CryptographicFailuresVulnerability {
...
+    private static final String LEVEL8_HASH = PasswordHashingUtils.lmHash(LEVEL8_SECRET);
+
+    private static final String LEVEL9_SECRET = "qwertyuiop";
```
preetkaran20: add few numbers too as that adds complexity.

src/main/java/org/sasanlabs/service/vulnerability/cryptographicFailures/CryptographicFailuresVulnerability.java
```diff
@@ -149,182 +377,199 @@ public ResponseEntity<GenericVulnerabilityResponseBean<String>> getVulnerablePay
...
+                                    + LEVEL_10_SECRET
+                                    + "'. Even though AES-128 is a secure encryption method, using a key that is easily guessable weakens the encryption's strength."
+                                    + " It provides zero security — anyone can decode it instantly.",
+                            true),
+                    HttpStatus.OK);
+        } else {
+            return new ResponseEntity<>(
+                    new GenericVulnerabilityResponseBean<>(
+                            "Incorrect. Hint: Think of common passwords '"
+                                    + LEVEL_10_SECRET
```
preetkaran20: we should not return secret.

src/main/java/org/sasanlabs/service/vulnerability/cryptographicFailures/CryptographicFailuresVulnerability.java
```diff
@@ -36,41 +38,266 @@ public class CryptographicFailuresVulnerability {
...
+
+    private static final String LEVEL9_SECRET = "qwertyuiop";
+    private static final String LEVEL9_HASH = DigestUtils.sha256Hex(LEVEL9_SECRET);
+
+    public static final String LEVEL_10_SECRET = "user";
+    public static final String LEVEL_10_ENCRYPTED =
+            EncryptionUtils.encrypt(
+                    LEVEL_10_SECRET, EncryptionUtils.getKeyFromPassword(LEVEL_10_SECRET));
+
+    private static final String LEVEL11_SECRET = "uncrackablePassword";
```
preetkaran20: make this passwrod random generated on each run as this is very secure and we want to tell that no one can guess this.

src/main/java/org/sasanlabs/service/vulnerability/cryptographicFailures/CryptographicFailuresVulnerability.java
preetkaran20: Image

Secret should not be told to user on failure.

src/main/java/org/sasanlabs/service/vulnerability/cryptographicFailures/CryptographicFailuresVulnerability.java
```diff
@@ -36,41 +38,266 @@ public class CryptographicFailuresVulnerability {
...
+                            "Correct! The password was '"
+                                    + LEVEL2_SECRET
+                                    + "'.\n Base64 is an encoding, NOT encryption."
+                                    + " It provides zero security — anyone can decode it instantly.",
+                            true),
+                    HttpStatus.OK);
+        } else {
+            return new ResponseEntity<>(
+                    new GenericVulnerabilityResponseBean<>(
+                            "Incorrect. Hint: Base64 is not encryption — try decoding '"
```
preetkaran20: Just tell it is incorrect. let user user hint section to find out that base64 is used.

src/main/java/org/sasanlabs/service/vulnerability/cryptographicFailures/CryptographicFailuresVulnerability.java
```diff
@@ -36,41 +38,266 @@ public class CryptographicFailuresVulnerability {
...
+                                    + LEVEL3_SECRET
+                                    + "'. Caesar Cipher is an insecure cipher and is trivial to crack."
+                                    + " There is a limited number of shifts and the same input will always have the same output",
+                            true),
+                    HttpStatus.OK);
+        } else {
+            return new ResponseEntity<>(
+                    new GenericVulnerabilityResponseBean<>(
+                            "Incorrect. The password has been shifted "
+                                    + LEVEL3_CHAR_SHIFT
```
preetkaran20: dont tell the shift size. let user find it out using hint. just tell that the password is incorrect.

src/main/java/org/sasanlabs/service/vulnerability/cryptographicFailures/CryptographicFailuresVulnerability.java
```diff
@@ -36,41 +38,266 @@ public class CryptographicFailuresVulnerability {
...
+                            "Correct! The password was '"
+                                    + LEVEL4_SECRET
+                                    + "'. Security through obscurity or custom logic is not secure."
+                                    + " Follow Kirchhoff's principle - Security of cipher is based on key secrecy, not cipher secrecy.",
+                            true),
+                    HttpStatus.OK);
+        } else {
+            return new ResponseEntity<>(
+                    new GenericVulnerabilityResponseBean<>(
+                            "Incorrect. "
```
preetkaran20, line 217: This is good hiding. Api should return just this.

src/main/java/org/sasanlabs/service/vulnerability/cryptographicFailures/CryptographicFailuresVulnerability.java
```diff
@@ -36,41 +38,266 @@ public class CryptographicFailuresVulnerability {
...
+                            "Correct! The password was '"
+                                    + LEVEL5_SECRET
+                                    + "'. MD4 is an insecure and hashes can be reversed using rainbow tables and online databases.",
+                            true),
+                    HttpStatus.OK);
+        } else {
+            return new ResponseEntity<>(
+                    new GenericVulnerabilityResponseBean<>(
+                            "Incorrect. Your input hashed to: "
+                                    + guessHash
```
preetkaran20, line 263: great !!!! telling guessHash is good as user gets some information using this.

preetkaran20 (commented):
Overall PR looks great. Just some small comments.

src/main/java/org/sasanlabs/service/vulnerability/cryptographicFailures/CryptographicFailuresVulnerability.java
aashpis: Agree on making a table for the secrets. I think having two tables, "vault" for the hashed/encrypted/encoded secrets for the challenge and then "secrets" for all plaintext secrets which could be could be used as answer key when the user can't figure out a password. 

Rather than manually producing hashes and ciphertext, I think we can use CommandLineRunner to seed the database using PasswordHashingUtils and EncryptionUtils. It can also be used to create the hashed passwords for the AuthenticationVulnerability module as the same hash algos are being used. Then you can update the raw password without having to manually produce the processed one and update the SQL.

Here is my plan for the vault schema:

1. id
2. password (the raw password after hashed/encrypted/encoded)
3. salt
4. algorithm
5. level

and for the secret schema

1. id 
2. rawPassword

preetkaran20: @aashpis great points. Agreed on the utility approach and the overall direction.

I would just suggest keeping the table names vulnerability-specific, for example:

* `CryptographicFailures_Vault`
* `CryptographicFailures_Secret`

The reason is isolation between vulnerabilities. One vulnerability should only expose data related to that vulnerability and should not accidentally expose data belonging to another module/challenge.

---

Also, great point regarding the `Secret` table. My concern is that we probably should not expose this table at all. One of the goals of the secure levels is to demonstrate that the implementation is actually secure. If the plaintext secret table itself becomes accessible, then even the secure levels are effectively compromised, which weakens the educational value of those levels.

For now, I think we should avoid having a `Secret` table altogether and instead let users discover the secret through the intended challenge/hack flow itself.

We should still build the utility layer though, so that we can easily generate newer secrets, hashes, encrypted values, salts, etc. whenever we want without manually updating SQL or challenge data.

src/main/java/org/sasanlabs/service/vulnerability/cryptographicFailures/repo/CryptographicFailuresSeeder.java
```diff
@@ -0,0 +1,66 @@
...
+
+    public CryptographicFailuresSeeder(CryptographicFailuresVaultRepository repository) {
+        this.repository = repository;
+    }
+
+
+    @Override
+    public void seed() {
+        // Level 1: Cleartext (Broken Cryptography)
+        repository.save(new VaultEntity(1, "letmein", null, "CLEARTEXT"));
```
preetkaran20: make it secure random instead.

src/main/java/org/sasanlabs/service/vulnerability/cryptographicFailures/repo/CryptographicFailuresSeeder.java
```diff
@@ -0,0 +1,66 @@
...
+
+        // Level 2: Base64 Encoding (Not Encryption)
+        repository.save(new VaultEntity(2, Base64.getEncoder().encodeToString("s3cretKey".getBytes()), null, "BASE64"));
```
preetkaran20: This one can be secure random too.

src/main/java/org/sasanlabs/service/vulnerability/cryptographicFailures/repo/CryptographicFailuresSeeder.java
```diff
@@ -0,0 +1,66 @@
...
+
+        // Level 3: Caesar Cipher (Weak Symmetric)
+        repository.save(new VaultEntity(3, EncryptionUtils.caesarCipher("weattackatdawn"), null, "CAESAR"));
```
preetkaran20: at runtime generate the string as securerandom, then we never need to make it hidden anywhere even if it is there in code as plain text.

src/main/java/org/sasanlabs/service/vulnerability/cryptographicFailures/repo/CryptographicFailuresSeeder.java
```diff
@@ -0,0 +1,66 @@
...
+        repository.save(new VaultEntity(7, DigestUtils.sha1Hex("1q2w3e4r"), null, "SHA-1"));
+
+        // Level 8: LM Hash (Legacy/Weak Windows Hash)
+        repository.save(new VaultEntity(8, PasswordHashingUtils.lmHash("admin123"), null, "LM"));
+
+        // Level 9: Unsalted SHA-256 (Fast Hash/Vulnerable to Rainbow Tables)
+        repository.save(new VaultEntity(9, DigestUtils.sha256Hex("qwertyuiop"), null, "SHA-256"));
+
+        // Level 10: AES-128 (Weak Key/Password is Key)
+        String level10Secret = "aa123456";
```
preetkaran20: Level10Secret can be as a plain text but everything else can be randomly generated at runtime

src/main/java/org/sasanlabs/internal/utility/EncryptionUtils.java
```diff
@@ -0,0 +1,82 @@
...
+        try {
+            SecretKeyFactory factory = SecretKeyFactory.getInstance("PBKDF2WithHmacSHA256");
+            byte[] emptySalt = new byte[16];
+            KeySpec spec = new PBEKeySpec(password.toCharArray(), emptySalt, 1, 128);
+
+            return new SecretKeySpec(factory.generateSecret(spec).getEncoded(), "AES");
+        } catch (NoSuchAlgorithmException | InvalidKeySpecException e) {
+            throw new RuntimeException("Error generating AES key from password: " + e);
+        }
+    }
```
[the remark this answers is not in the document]
preetkaran20: can you please fix this

src/main/java/org/sasanlabs/configuration/DatabaseSeeder.java
```diff
@@ -0,0 +1,35 @@
...
+import org.springframework.context.event.EventListener;
+import org.springframework.stereotype.Component;
+
+import java.util.List;
+/**
+ * Runs all {@link ModuleSeeder} implementations once the application is ready,
+ * ensuring the database is populated before the app accepts requests.
+ */
+@Component
+public class DatabaseSeeder {
```
preetkaran20, line 15: move this to src/main/java/org/sasanlabs/configuration location isntead

src/main/java/org/sasanlabs/service/vulnerability/cryptographicFailures/repo/CryptographicFailuresSeeder.java
```diff
@@ -0,0 +1,66 @@
...
+
+    private final CryptographicFailuresVaultRepository repository;
+
+    public CryptographicFailuresSeeder(CryptographicFailuresVaultRepository repository) {
+        this.repository = repository;
+    }
+
+
+    @Override
+    public void seed() {
```
preetkaran20: ensure making it transactional.

build.gradle
```diff
@@ -155,8 +155,13 @@ dependencies {
 
     implementation 'com.unboundid:unboundid-ldapsdk:6.0.5'
     implementation 'org.springframework.security:spring-security-crypto'
+
+    // https://mvnrepository.com/artifact/org.bouncycastle/bcprov-jdk18on
+    implementation group: 'org.bouncycastle', name: 'bcprov-jdk18on', version: '1.84'
```
preetkaran20, line 163: does bouncy castle supports LMHash and other algorithms we have in PasswordHashingUtils? If so, possible to use it for them?

aashpis: It doesn't have LMHash, but I believe has support for the rest of the algorithms. I think it would be better to rely on BC as spring security password encoders can include salt with the hashes which would prevent users from attempting a rainbow table attack or using hash generators or crackers. 

It has been difficult to find to any java implementation of LM Hash. I have only found JCIFS, which appears to be a legacy library no longer in use. 

The best I found is this method, but can't confirm if it's the correct algorithm. 
https://javadoc.io/doc/org.codelibs/jcifs/latest/jcifs/org/codelibs/jcifs/smb/impl/NtlmUtil.html#getNTHash(java.lang.String)

Here is the maven page for it, but the homepage is no longer in use
https://mvnrepository.com/artifact/jcifs/jcifs

This seems to be the current repo: 
https://github.com/ioplex/jcifs

Without having a reliable library to use, one suggestions is to remove the LM hash level. It's been deprecated since Windows Vista and would be a niche issue when working with legacy window server system. 

I think the MD4 and MD5 levels would demonstrate enough about insecure hashes.

src/main/java/org/sasanlabs/service/vulnerability/cryptographicFailures/repo/CryptographicFailuresSeeder.java
aashpis: Would this include the other hashed password levels (md4, md5, etc)? I think it would be good to have at least one level that allows for a simple rainbow table style attack or generating hashes based on common passwords lists?

build.gradle
preetkaran20: I think it is ok if we don't find LMHash algorithm in BC. Howevern I feel It is good to have LMHash as it gives another flavour.

src/main/java/org/sasanlabs/service/vulnerability/cryptographicFailures/repo/CryptographicFailuresSeeder.java
preetkaran20: I think no harm in having all others as random as users could find the real value by using rainbow tables.

preetkaran20: major reason is https://github.com/SasanLabs/VulnerableApp/pull/596#discussion_r3228650699 and also that will help us make it like a CTF challenge too if we need in future.

src/main/resources/i18n/messages.properties
```diff
@@ -309,13 +309,19 @@ Important Links:<br/>\
...
+CRYPTOGRAPHIC_FAILURES_MD5_HASHING=Password is hashed using MD5 algorithm which is cryptographically broken. The user must crack the MD5 hash using rainbow tables or online tools to find the original password.
+CRYPTOGRAPHIC_FAILURES_SHA1_HASHING=Password is hashed using SHA1 algorithm which is deprecated. The user must crack the SHA1 hash using rainbow tables or online tools to find the original password.
+CRYPTOGRAPHIC_FAILURES_LM_HASHING=Password is hashed using LM algorithm which hasv been compromised. The user is show how the LM hash is vulnerable and easy to crack.
+CRYPTOGRAPHIC_FAILURES_SHA256_HASHING=Password is hashed using SHA-256 without a salt, making it vulnerable to rainbow table and precomputed hash attacks. The user must crack the SHA-256 hash using online tools to find the original password.
 CRYPTOGRAPHIC_FAILURES_SECURE_SHA256=Password is hashed using SHA-256 with salt, making rainbow table attacks ineffective. In production, use bcrypt, Argon2, or PBKDF2 for password hashing.
+CRYPTOGRAPHIC_FAILURES_INSECURE_AES-128=Password is encrypted with AES-128 but implemented with a weak key making it secure. The user must crack the password using
```
preetkaran20: line is not complete

src/main/resources/application.properties
```diff
@@ -20,7 +20,7 @@ spring.h2.console.settings.web-allow-others=true
 spring.h2.console.path=/h2
 
 # Pre-Populating data and schema is handled by DataSourceInitializer bean in VulnerableAppConfiguration.java
-spring.jpa.hibernate.ddl-auto=none
+spring.jpa.hibernate.ddl-auto=update
```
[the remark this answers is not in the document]
preetkaran20: we should have it as none.

src/main/java/org/sasanlabs/service/vulnerability/cryptographicFailures/repo/VaultEntity.java
```diff
@@ -0,0 +1,68 @@
+package org.sasanlabs.service.vulnerability.cryptographicFailures.repo;
+
+import org.springframework.lang.Nullable;
+
+import javax.persistence.*;
+
+@Entity
+@Table(name="CryptographicFailures_Vault")
```
preetkaran20: please create this table schema via db scripts. We have DataSourceInitializer bean in VulnerableAppConfiguration.java for more information

preetkaran20: Also create newer username which only has access to this table.  example: 
https://github.com/SasanLabs/VulnerableApp/blob/master/src/main/resources/scripts/Authentication/db/schema.sql
or
https://github.com/SasanLabs/VulnerableApp/blob/master/src/main/resources/scripts/IDOR/db/schema.sql

src/main/java/org/sasanlabs/configuration/DatabaseSeeder.java
```diff
@@ -0,0 +1,35 @@
...
+    //  Finds every @Component that implements ModuleSeeder
+    private final List<ModuleSeeder> seeders;
+
+    public DatabaseSeeder(List<ModuleSeeder> seeders) {
+        this.seeders = seeders;
+    }
+
+    @EventListener(ApplicationReadyEvent.class) // Runs when  application is ready
+    public void seedAllModules() {
+        System.out.println("Starting Global Database Seeding...");
```
preetkaran20: please use logger instead of println. 

**Example**:  private static final transient Logger LOGGER = LogManager.getLogger(PathTraversalVulnerability.class);

src/main/java/org/sasanlabs/configuration/DatabaseSeeder.java
```diff
@@ -0,0 +1,35 @@
...
+
+        for (ModuleSeeder seeder : seeders) {
+            if (!seeder.isSeeded()) {
+                seeder.seed();
+            }
+        }
+
+        System.out.println("Seeding complete. Processed " + seeders.size() + " modules.");
```
preetkaran20: please fix this with logger and use example:

logger.info("Seeding complete. Processed {} modules", seeders.size());

src/main/java/org/sasanlabs/internal/utility/EncodingUtils.java
```diff
@@ -0,0 +1,17 @@
...
+    public static String bytesToHex(byte[] data) {
+        StringBuilder builder = new StringBuilder(data.length * 2);
+        for (byte value : data) {
+            builder.append(String.format("%02x", value));
+        }
+        return builder.toString();
+    }
+
+    public static String encodeBase64(String rawText) {
+        return Base64.getEncoder().encodeToString(rawText.getBytes());
```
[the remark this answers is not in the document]
preetkaran20: @aashpis please fix this as suggested.

src/main/java/org/sasanlabs/internal/utility/EncryptionUtils.java
```diff
@@ -0,0 +1,87 @@
...
+    private EncryptionUtils() {}
+
+    /**
+     * INSECURE: Caesar Cipher shifts alphabetic characters positions to the right overflowing to
+     * the beginning of the alphabet. 'z' will shift to 'a' and so on.
+     *
+     * @param rawPassword plaintext password to encrypt
+     */
+    public static String caesarCipher(String rawPassword) {
+        int shift = 3; // number of positions to shift each alphabetic character to the right
```
preetkaran20: please take shift as input to this function as then we can reuse this utility at other places as well.

src/main/java/org/sasanlabs/internal/utility/EncryptionUtils.java
```diff
@@ -0,0 +1,87 @@
...
+
+    public static SecretKey getKeyFromPassword(String password) {
+        try {
+            SecretKeyFactory factory = SecretKeyFactory.getInstance("PBKDF2WithHmacSHA256");
+            KeySpec spec = new PBEKeySpec(password.toCharArray(), salt, 1, 128);
+
+            return new SecretKeySpec(factory.generateSecret(spec).getEncoded(), "AES");
+        } catch (NoSuchAlgorithmException | InvalidKeySpecException e) {
+            throw new RuntimeException("Error generating AES key from password: " + e);
+        }
```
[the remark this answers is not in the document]
preetkaran20: @aashpis please follow this suggestion. this is good suggestion.

src/main/java/org/sasanlabs/internal/utility/EncryptionUtils.java
```diff
@@ -0,0 +1,87 @@
...
+    public static String encrypt(String plaintext, SecretKey key) {
+        try {
+            // VULNERABILITY NOTE: ECB mode does not use an IV and reveals patterns (CWE-327)
+            Cipher cipher = Cipher.getInstance("AES/ECB/PKCS5Padding");
+            cipher.init(Cipher.ENCRYPT_MODE, key);
+
+            byte[] encrypted = cipher.doFinal(plaintext.getBytes(StandardCharsets.UTF_8));
+            return java.util.Base64.getEncoder().encodeToString(encrypted);
+
+        } catch (NoSuchPaddingException | NoSuchAlgorithmException e) {
```
preetkaran20, line 94: I think better to create a newer exception class which is for all the encryption related failures so we can ignore RuntimeException and also checked exception is better so that caller is aware of it and can handle it.

src/main/java/org/sasanlabs/internal/utility/ModuleSeeder.java
```diff
@@ -0,0 +1,16 @@
+package org.sasanlabs.internal.utility;
+
+import org.sasanlabs.configuration.DatabaseSeeder;
+
+/**
+ * Implement to seed the database for a module. ModuleSeeders are auto-discovered and called by
+ * {@link DatabaseSeeder} on application startup.
+ */
+public interface ModuleSeeder {
```
preetkaran20: I think we should have it as part of configurations folder only. can you please move it to org/sasanlabs/configuration

src/main/java/org/sasanlabs/service/vulnerability/cryptographicFailures/repo/CryptographicFailuresSeeder.java
```diff
@@ -0,0 +1,94 @@
...
+            new RandomStringGenerator.Builder()
+                    .usingRandom(secureRandom::nextInt)
+                    .withinRange(33, 126)
+                    .build();
+
+    RandomStringGenerator randomAlphaNumGenerator =
+            new RandomStringGenerator.Builder()
+                    .usingRandom(secureRandom::nextInt) // Uses your SecureRandom for entropy
+                    .selectFrom(
+                            "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789"
```
preetkaran20: can you please create a string constant for this.

src/main/java/org/sasanlabs/service/vulnerability/cryptographicFailures/CryptographicFailuresVulnerability.java
preetkaran20: can you please change this string to something like "CHALLENGE: The system stores passwords with a weak encoding schema, Check the database for the password and crack the challenge"

src/main/java/org/sasanlabs/service/vulnerability/cryptographicFailures/CryptographicFailuresVulnerability.java
```diff
@@ -107,224 +313,253 @@ public ResponseEntity<GenericVulnerabilityResponseBean<String>> getVulnerablePay
...
             return new ResponseEntity<>(
                     new GenericVulnerabilityResponseBean<>(
-                            "SECURE: This password is hashed with SHA-256 + salt. Hash: "
-                                    + LEVEL5_HASH
-                                    + " | Salt: "
-                                    + LEVEL5_SALT
-                                    + " — Try to crack it (you likely can't with rainbow tables"
-                                    + " because the salt defeats pre-computed lookups).",
-                            true),
+                            "CHALLENGE: The system encrypts passwords using AES-128 encryption. AES-128 is normally a secure encryption however "
```
preetkaran20: We should somehow depicts this to the user.
The system encrypts passwords using AES-128 encryption. This has two critical flaws: first, the password is used as its own encryption key, making it trivially crackable. But more fundamentally — passwords should never be encrypted at all. Encryption is a two-way function, meaning anyone with the key can recover the original password. Passwords should always be stored using a one-way hashing function (e.g. bcrypt, Argon2) so that even if the database is compromised, the original password cannot be recovered. The stored ciphertext is: {LEVEL_10_CIPHERTEXT} — crack it and enter the original password!

src/main/resources/i18n/messages.properties
```diff
@@ -309,12 +309,17 @@ Important Links:<br/>\
...
+CRYPTOGRAPHIC_FAILURES_MD5_HASHING=Password is hashed using MD5 algorithm which is cryptographically broken. The user must crack the MD5 hash using rainbow tables or online tools to find the original password.
+CRYPTOGRAPHIC_FAILURES_SHA1_HASHING=Password is hashed using SHA1 algorithm which is deprecated. The user must crack the SHA1 hash using rainbow tables or online tools to find the original password.
+CRYPTOGRAPHIC_FAILURES_LM_HASHING=Password is hashed using LM algorithm which has been compromised. The user is shown how the LM hash is vulnerable and easy to crack.
+CRYPTOGRAPHIC_FAILURES_SHA256_HASHING=Password is hashed using SHA-256 without a salt, making it vulnerable to rainbow table and precomputed hash attacks. The user must crack the SHA-256 hash using online tools to find the original password.
+CRYPTOGRAPHIC_FAILURES_INSECURE_AES128=Password is encrypted with AES-128 but implemented with a weak key making it insecure to brute-force attacks. The user must crack the password using common password lists or key-recovery techniques.
```
preetkaran20: can you please add detail that  Encryption is a two-way function, meaning anyone with the key can recover the original password. Passwords should always be stored using a one-way hashing function (e.g. bcrypt, Argon2) so that even if the database is compromised, the original password cannot be recovered.

src/main/resources/i18n/messages.properties
preetkaran20, line 302: can you please add database access details here using the newer user we created for cryptographic module?

src/main/resources/scripts/CryptographicFailures/db/schema.sql
```diff
@@ -0,0 +1,14 @@
...
+    level INT PRIMARY KEY ,
+    password VARCHAR(500),
+    algorithm VARCHAR(50)
+);
+
+-- Application user has full access (for functional purposes)
+GRANT ALL ON cryptographic_failures_vault TO application;
+
+-- A read-only user for exploration by the attacker/user
+CREATE USER IF NOT EXISTS readonly_user PASSWORD 'readonly_password';
```
preetkaran20: create specific user for cryptographic failures.

src/main/resources/i18n/messages.properties
preetkaran20: Example:

<b>Note:</b> For most levels, you can explore the database contents directly at the H2 Console:<br/>
- Legacy UI: <a href=\"http://localhost:9090/VulnerableApp/h2/\" target=\"_blank\">http://localhost:9090/VulnerableApp/h2/</a><br/>
- Docker UI: <a href=\"http://localhost/VulnerableApp/h2/login.jsp\" target=\"_blank\">http://localhost/VulnerableApp/h2/login.jsp</a><br/>
(URL: <code>jdbc:h2:mem:testdb</code>, Username: <code>cryptographic_failures_user</code>, Password: <code>cryptographic_failures_password</code>)<br/><br/>

preetkaran20 (approved):
@aashpis can you please do some small fixes. From my side, this entire change looks good to me overall. we will merge this very soon.

aashpis:
Thank you for all your reviews, I will keep working at it.

aashpis:
I had some issues with my commits while trying to implement a hash exception class, I ended up undoing it do some cascading issues that caused with other modules outside cryptographic failures. I believe I undid any changes I did, but wanted to note in case there's anything I might have overlooked. 

Previously mentioned changes have been implemented otherwise.

preetkaran20:
🎉 Thanks for contributing @aashpis!
   https://docs.google.com/forms/d/e/1FAIpQLSfwWVdnULUhtfruA-DN328NwKnBGebaWg9U5y0xivLLxxoMog/viewform?usp=pp_url&entry.1414771947=aashpis
   
👉 [#524 - In the deprecated UI, there is a small flicker which is impacting the UI while navigating from one lebvel to another](https://github.com/SasanLabs/VulnerableApp/issues/524)

preetkaran20:
@all-contributors please add @aashpis for code

preetkaran20:
@aashpis Wow !!! Great work on this PR. Thanks a lot for all the changes.

PR #598, opened 2026-04-24
Add GraphQL Authorization & Access Control vulnerability category (#530)

### LINKED ISSUE #530: GraphQL Authorization & Access Control Vulnerabilities
This issue introduces multiple vulnerable GraphQL endpoints that demonstrate common authorization and access control weaknesses. GraphQL APIs often rely on resolver-level authorization logic, and misconfigurations can easily expose sensitive data or allow privilege escalation.

### Sample Vulnerability Scenarios
1. **Insecure Direct Object Reference (IDOR)**  
   Users can query resources belonging to other users by modifying object identifiers.

   Example:

   ```graphql
   {
     user(id: 5) {
       email
       role
     }
   }

2. **Broken Object Level Authorization**
GraphQL resolvers return objects without verifying ownership or permissions.

3. **Privilege Escalation via Mutation**
Attackers can modify their role using mutations.

mutation {
  updateUserRole(id: 2, role: "ADMIN")
}

4. **Admin Mutation Exposure**
Mutations intended for administrators are accessible to regular users.

5. Secure implementation

zhouchristopher:
Closes #530.

Adds a new top-level vulnerability category, `GraphQL Authorization & Access Control`, with the four unsafe variants + one secure reference that matches the category-layout convention used by the rest of the app.

## Levels

| Level | Surface | Bug / defense |
|---|---|---|
| LEVEL_1 | `user(id:)` query | **IDOR** — no authentication at all |
| LEVEL_2 | `order(id:)` query | **BOLA** — token required but ownership is not checked |
| LEVEL_3 | `updateUserRole` mutation | **Privilege escalation** — any authenticated user can change any role |
| LEVEL_4 | `adminReport` query + `deleteUser` mutation | **Admin mutation exposure** — no role check |
| LEVEL_5 | same schema as LEVEL_4 | **Secure** — owner check on per-record reads; admin ops re-read role from DB on every request so a tampered token cannot grant admin |

Full per-level schemas, exploit `curl` commands, and a reproduction walk-through are in [`docs/GraphQLVulnerability.md`](https://github.com/zhouchristopher/VulnerableApp/blob/feature/graphql-authz-access-control-530/docs/GraphQLVulnerability.md).

## Implementation notes

- Uses `graphql-java:21.5` directly rather than `spring-boot-starter-graphql`. The starter auto-configures a single `/graphql` endpoint, which would conflict with VulnerableApp's per-level `/<Category>/LEVEL_n` pattern.
- Each level builds its own `GraphQL` instance from its own SDL file so resolver fixes do not leak across levels.
- Authentication reuses `IDORLoginService` (Base64-encoded JSON tokens); the decoded `User` is attached to `GraphQLContext`. No new password-handling code.
- Two new H2 tables (`graphql_users`, `graphql_orders`) seeded via `DataSourceInitializer`. Kept separate from `idor_users` so mutations in this category don't corrupt IDOR state.
- New enum constant `VulnerabilityType.GRAPHQL_BROKEN_ACCESS_CONTROL` (CWE-285, WASC-2).
- Registration with `/scanner` and the facade UI is automatic via the existing `@VulnerableAppRestController` / `@AttackVector` annotations.
- One shared UI template under `static/templates/GraphQLVulnerability/LEVEL_1/` drives all five levels; the JS pre-fills a level-appropriate default query.

## Tests

`src/test/java/org/sasanlabs/service/vulnerability/graphql/GraphQLVulnerabilityTest.java` — 12 JUnit 5 + Mockito cases covering:

- **Levels 1–4**: exploit succeeds as documented
- **Level 5**: blocks `updateUserRole` and `deleteUser` for non-admins; blocks non-owner `user(id:)`; allows owner self-query; allows genuine admin; **blocks a tampered token claiming `role=ADMIN` when the DB says the caller is `USER`**

## Verification

- `./gradlew clean build` — green locally (Temurin JDK 17)
- `./gradlew spotlessCheck` — green (no style changes needed)
- `./gradlew bootRun` + `curl` smoke of every level — all exploits behave as documented; all Level 5 defenses reject the same payloads
- `curl /VulnerableApp/scanner` returns five entries with `vulnerabilityTypes: ["GRAPHQL_BROKEN_ACCESS_CONTROL"]` (UNSECURE × 4, SECURE × 1) — the new module auto-registers without manual catalog edits

## Files

- Java (controller + login + POJOs): `src/main/java/org/sasanlabs/service/vulnerability/graphql/`
- SDL: `src/main/resources/scripts/GraphQL/schema/level{1..4}.graphqls`
- SQL: `src/main/resources/scripts/GraphQL/db/{schema,data}.sql`
- UI: `src/main/resources/static/templates/GraphQLVulnerability/LEVEL_1/GraphQL.{html,css,js}`
- i18n: `src/main/resources/i18n/messages{,_en_US}.properties`
- Payloads: `src/main/resources/attackvectors/GraphQLVulnerabilityPayload.properties`
- Tests: `src/test/java/org/sasanlabs/service/vulnerability/graphql/GraphQLVulnerabilityTest.java`
- Docs: `docs/GraphQLVulnerability.md`, plus entry #14 in `README.md`
- Config/wiring: `build.gradle` (graphql-java dep), `VulnerableAppConfiguration` (DB script registration), `VulnerabilityType` (enum)

## Summary by CodeRabbit

* **New Features**
  * Added GraphQL Authorization & Access Control vulnerability module with five progressive levels demonstrating IDOR, BOLA, privilege escalation, and unauthorized admin operation exposure.
  * Interactive GraphQL query interface for testing authorization controls at each level.

* **Documentation**
  * Added comprehensive documentation for the GraphQL Authorization & Access Control module, including schema details, exploit examples, and QA guidance.


preetkaran20:
@coderabbitai review

preetkaran20:
Sorry for the delay. I will be reviewing this PR in next couple of days.

thanks,
Karan

src/main/java/org/sasanlabs/service/vulnerability/graphql/GraphQLLoginController.java
```diff
@@ -0,0 +1,42 @@
...
+    private final IDORLoginService idorLoginService;
+
+    public GraphQLLoginController(IDORLoginService idorLoginService) {
+        this.idorLoginService = idorLoginService;
+    }
+
+    @PostMapping("login")
+    public ResponseEntity<GenericVulnerabilityResponseBean<Object>> login(
+            @RequestParam String username, @RequestParam String password) {
+        IDORLoginDetails loginDetails = idorLoginService.loginWithToken(username, password);
```
preetkaran20, line 33: Thank you for using IDOR Login Service. I found a flaw in that service. I will try to fix it with JWT based setup where HS256 key is generated on server startup and will be used to sign the JWT which will be used to validate it in the later end.

preetkaran20: We did the change. please checkout.

src/main/resources/scripts/GraphQL/db/data.sql
```diff
@@ -0,0 +1,8 @@
+INSERT INTO graphql_users VALUES (1, 'Alice', 'alice@example.com', 50000, 'USER');
```
preetkaran20, line 1: I think we are mutating the data so we should have per level data i.e. introduce level column into the table os one level doesn't update other level's data.

src/main/resources/static/templates/GraphQLVulnerability/LEVEL_1/GraphQL.html
```diff
@@ -0,0 +1,51 @@
...
+            <div id="loginPill" class="login-pill hidden">
+                Logged in as <strong id="loginUserName"></strong>
+                (id=<span id="loginUserId"></span>, role=<span id="loginUserRole"></span>)
+                <div class="token-row">token: <code id="loginToken"></code></div>
+            </div>
+        </div>
+    </div>
+
+    <div class="gql-card">
+        <h5>Query</h5>
```
preetkaran20, line 34: I would suggest to hide this from user and let use find out that a graphql query is fired in the backgound and manipulate the query.

Create a simple button called say fetch profile or fetch order. In the hints section tell the user to fetch the profiule or other users order.

src/main/resources/static/templates/GraphQLVulnerability/LEVEL_1/GraphQL.js
```diff
@@ -0,0 +1,137 @@
+const GQL_DEFAULT_QUERIES = {
+  LEVEL_1: `# IDOR - no authentication required.\n# Change the id argument to fetch any user's data.\n{\n  user(id: 3) {\n    id\n    username\n    email\n    salary\n    role\n  }\n}`,
+  LEVEL_2: `# BOLA - login first, then fetch an order owned by another user.\n# Order 103 belongs to Bob. Login as Alice, then run this query.\n{\n  order(id: 103) {\n    id\n    userId\n    item\n    amount\n  }\n}`,
+  LEVEL_3: `# Privilege escalation - login as Alice (USER), then change her own role.\nmutation {\n  updateUserRole(id: 1, role: "ADMIN") {\n    id\n    username\n    role\n  }\n}`,
+  LEVEL_4: `# Admin mutation exposure - login as any regular user, then call an admin-only op.\n{\n  adminReport {\n    id\n    username\n    email\n    role\n  }\n}`,
```
preetkaran20, line 5: We should make the progression consistent across levels so the learning flow is clearer.

Level 1: vulnerable profile/order access where any user can fetch or update another user’s profile or order by changing the ID, and can also fetch their own order history.
Level 2: login is required, but a user can still fetch another user’s order and manipulate it.
Level 3: vulnerable to privilege escalation, where a regular user can modify their own role.
Level 4: vulnerable to function-level authorization issues, where a regular user can call an admin-only action. I am not sure how we can have consistent UX for this.
Level 5: the same actions should now be properly blocked by authorization checks.

This way each level follows the same UI mental model, and the security issue becomes easier to understand.

src/main/java/org/sasanlabs/service/vulnerability/graphql/GraphQLVulnerability.java
```diff
@@ -0,0 +1,474 @@
...
+                                                .dataFetcher("order", orderFetcher)
+                                                .dataFetcher("myOrders", myOrdersFetcher))
+                        .build();
+        return build("scripts/GraphQL/schema/level2.graphqls", wiring);
+    }
+
+    private GraphQL buildLevel3() throws IOException {
+        // LEVEL 3 - Privilege Escalation: any authenticated user can change any user's role.
+        DataFetcher<GraphQLUser> meFetcher = env -> fetchUser(requireUser(env).getUserId());
+        DataFetcher<GraphQLUser> userFetcher =
```
preetkaran20, line 242: I think we can improve the design by extracting a core GraphQL service and placing a lightweight vulnerability/level layer on top of it.
This would keep schema + resolvers consistent across levels, and allow each level to only vary in authorization strategy (rather than duplicating GraphQL instances).
It will also make it easier to add future GraphQL attack scenarios like introspection abuse or query complexity attacks.

preetkaran20 (changes requested):
great PR @zhouchristopher and this seems like an amazing introduction to the project. Can you please fix some of the comments.

Also, backend is not returning token when we login, can you start setting cookie for each level like level1_token or level2_token and then that cookie will be attached with the calls to backend from UI.

preetkaran20:
@zhouchristopher please let me know if you need any help with this PR.

Issue #554, opened 2026-04-10
Enhance authentication related module


preetkaran20:
We have built an amazing authentication module recently with 9 levels but authentication is really an amazing topic and can be enhanced to more levels. In this task we wanted to enhance it more.

Goal is to create a newer vulnerability may be calling it authentication 2 or extension to auth and build following levels:  

1. Low iteration password hashing - this can be added to existing authentication module
2. Session fixation
3. Predictable session IDs
4. Missing logout invalidation
5. Password reset token predictable
6. Password reset token reuse
7. Password reset without identity verification
8. and more can be added. 

These can be added as newer module as this is vast. 
- OAuth misconfiguration
- Missing MFA support
- JWT signed with weak secret
- JWT algorithm confusion
- and many more as per implementation thoughts

preetkaran20:
Previous issue for authentication module first iteration https://github.com/SasanLabs/VulnerableApp/issues/513

It has a lot of details.

aryankshl:
Hi @preetkaran20 , thank you for merging my previous PR! I would love to take on this challenge. Since this covers a wide range of vulnerabilities, I suggest starting by creating the Authentication 2.0 module skeleton and implementing the first few levels (like Low iteration hashing and Session fixation) to establish the structure.

Please assign this to me, and I will share a brief implementation plan for the first set of levels shortly.

preetkaran20:
@aryankshl sure, please go ahead.

preetkaran20:
Hi everyone,

Low iteration was fixed but we have multiple other things which can be added. we welcome contributions to this issue.

PR #620, opened 2026-05-18
feat(auth): implement Level 10 - Low Iteration Password Hashing

aryankshl:
## Summary
This PR implements the first part of the Authentication enhancements by adding **Level 10: Low Iteration Password Hashing** to the existing authentication module. This level demonstrates the risk of utilizing modern adaptive hashing algorithms (like BCrypt) with a dangerously low work factor.

## Changes
- **Controller Layer**: Added the Level 10 endpoint (`level10LowIterationHashing`) inside `AuthenticationVulnerability.java` mapping to `LevelConstants.LEVEL_10` and utilizing a secure `POST` method.
- **Service Layer**: Injected custom verification logic inside `AuthLoginService.java` to handle the new algorithm verification.
- **Data/Enum Layer**: Extended the `AuthUserAlgorithm` enum to support the `BCRYPT_LOW_ITERATION` constant.

---

## Technical Context: Why 4 Iteration Rounds?

In modern authentication systems, algorithms like BCrypt are designed to be "computationally expensive" to prevent high-speed brute-force attacks. The work factor parameter operates exponentially as $2^{\text{cost}}$.

- **Secure Standard (Cost 10-12)**: A cost factor of 10 requires $2^{10} = 1,024$ key derivation iterations, taking roughly 100ms per verification.
- **Vulnerable Standard (Cost 4)**: A cost factor of 4 is the **absolute theoretical and structural minimum** allowed by Java's `BCryptPasswordEncoder` implementation. It requires only $2^4 = 16$ iterations.

### Why this is a vulnerability:
By reducing the iterations from a secure baseline (e.g., 10) to 4, we decrease the computational complexity by a factor of 64 ($1024 / 16$). Because the hash computes near-instantaneously, an attacker who compromises the database can execute dictionary or brute-force attacks 64 times faster than on a properly hardened system, entirely defeating the purpose of an adaptive hashing algorithm.

---

## How to Test
1. Compile the changes using `./gradlew compileJava`.
2. (Note: Once the DB seeding structure is finalized in the next step, this endpoint will process live mock users containing the `BCRYPT_LOW_ITERATION` flag).

Part of #554

## Summary by CodeRabbit

* **New Features**
  * Added a Level 10 "Low Iteration Hashing" authentication demo endpoint.
  * Added support for a low-iteration BCrypt variant for password verification and display.
  * Added a new Level 8 authentication UI (HTML/CSS/JS) for interactive testing.
  * Updated client-side login behavior to consistently send credentials and display responses.

* **Documentation**
  * Added localized messages describing the low-iteration hashing demo and payload.

[Review Change Stack](https://app.coderabbit.ai/change-stack/SasanLabs/VulnerableApp/pull/620?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


src/main/java/org/sasanlabs/service/vulnerability/authentication/AuthenticationVulnerability.java
```diff
@@ -228,6 +228,32 @@ public ResponseEntity<GenericVulnerabilityResponseBean<Object>> level9Secure(
         return response(buildProfile(result.getUser()), true);
     }
 
+    // ------------------------------------------------------------------ Level 10 — Low Iteration
+    // Hashing
+    @AttackVector(
+            vulnerabilityExposed = VulnerabilityType.WEAK_PASSWORD_HASHING,
+            description = "AUTH_LEVEL_10_LOW_ITERATION_HASHING",
```
preetkaran20, line 235: please add this description label to messages.properties.

src/main/java/org/sasanlabs/service/vulnerability/authentication/AuthenticationVulnerability.java
```diff
@@ -228,6 +228,32 @@ public ResponseEntity<GenericVulnerabilityResponseBean<Object>> level9Secure(
...
+            payload = "AUTH_PAYLOAD_LEVEL_10")
```
preetkaran20, line 236: Also add this payload too

src/main/java/org/sasanlabs/service/vulnerability/authentication/AuthenticationVulnerability.java
```diff
@@ -228,6 +228,32 @@ public ResponseEntity<GenericVulnerabilityResponseBean<Object>> level9Secure(
...
+    @VulnerableAppRequestMapping(
```
preetkaran20, line 237: The code changes are not working and are throwing 405 response code. can you please test the changes before making the code ready for review?

aryankshl: Hi @preetkaran20,

I tracked down the root cause of the `405` Method Not Allowed error. The backend endpoint is configured to handle explicit `POST` requests (matching the conventions of Level 8 and 9). However, because Level 10 reuses the shared `LEVEL_1/Auth` frontend UI template layout, the frontend asset `(/src/main/resources/static/templates/AuthenticationVulnerability/LEVEL_1/Auth.js)` defaults to dispatching form inputs via `GET` requests due to a hardcoded conditional evaluation block:
```
if (level === "LEVEL_8" || level === "LEVEL_9") { ... doPostAjaxCall ... } 
else { ... doGetAjaxCall ... }
```
Since Level 10 falls straight into the fallback block, the browser triggers a `GET` request, causing the backend validation filter to reject it with a `405`.

To move forward sustainably, how would you prefer this to be resolved? I can implement either option:

**Backend Alignment (GET)**: Remove the strict `POST` method requirement from the backend controller for Level 10. This aligns it immediately with the existing frontend behavior and mirrors the weak hashing setups used in Levels 4, 5, and 6.

**Frontend Adjustment (POST)**: Update the JavaScript routing block. We can append `"LEVEL_10"` to the existing conditional statement, or refactor it more scalably to read the request method dynamically from the vulnerability scanner's metadata payload to avoid hardcoding future levels.

Let me know your thoughts on the preferred direction and I will update the code right away!

preetkaran20: Oh got it. I think current pattern is wrong. Can you please create a newer template for level 8, level 9 and level 10 and start using that template for these levels and remove if check as that is not scalable?

aryankshl: Thank you for the feedback! I completely agree that refactoring this into a clean, dedicated template is the right long-term approach for the project's scalability.
I will create the new reusable frontend template, migrate Levels 8, 9, and 10 to utilize it, and clean up the hardcoded conditional checks from the legacy script. I'll ping you here as soon as the full-stack changes are ready for review.

preetkaran20 (approved):
Thanks  a lot for the PR @aryankshl . great work !!!

aryankshl:
Thanks for merging! I'll be picking up the next tracking vulnerability from the roadmap very soon.

PR #625, opened 2026-05-20
Added @Profile("public"/"unsafe") to all vulnerability controllers

prajp98:
## Summary by CodeRabbit

* **Chores**
  * Demo controllers and endpoints are now environment-gated and only load for the relevant profile ("public" or "unsafe").
  * Default active profiles now include both "public" and "unsafe", enabling both sets of demo features by default.
  * Benchmarking and unsafe-demo endpoints remain restricted to the "unsafe" profile for selective activation.

[Review Change Stack](https://app.coderabbit.ai/change-stack/SasanLabs/VulnerableApp/pull/625?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/main/java/org/sasanlabs/benchmark/controller/BenchmarkController.java
+++ b/src/main/java/org/sasanlabs/benchmark/controller/BenchmarkController.java
@@ -9,6 +9,7 @@ import org.sasanlabs.benchmark.model.BenchmarkResult;
 import org.sasanlabs.benchmark.model.ScannerFindings;
 import org.sasanlabs.benchmark.service.BenchmarkResultWriter;
 import org.sasanlabs.benchmark.service.BenchmarkService;
+import org.springframework.context.annotation.Profile;
 import org.springframework.http.HttpStatus;
 import org.springframework.http.ResponseEntity;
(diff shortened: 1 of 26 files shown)


preetkaran20 (commented):
@prajp98 there is 3 more files 
1. in resources folder src\main\resources\sampleVulnerability\sampleVulnerability\SampleVulnerability.java
2. these 2 files as well image

can you please update that too and add profile as public.

prajp98:
Spring doesn't load java files in resources folder. So annotation is not needed for SampleVulnerability.
Updated PR

preetkaran20:
> Spring doesn't load java files in resources folder. So annotation is not needed for SampleVulnerability. Updated PR

Correct but we have a gradle task that copies sample vulnerability from resources into src/main/java. So the use case of sample vulnerability is if someone want to implement newer vulnerability, they can just run the task which copies it to src/main/java and then they rename things and it becomes newer vulnerability

prajp98:
@preetkaran20 got it, updated the PR

preetkaran20 (approved):

Issue #512, opened 2026-03-08
Enhance IDOR Module with Multiple Real-World Scenarios


preetkaran20:
1. IDOR Using UUID Identifiers
Modern applications often use UUIDs instead of sequential IDs.
Example:
GET /profile/4f7e0c1a-889e-4e2a-9a4a-22a34e
Although UUIDs are harder to guess, they may still be exposed through:
API responses
logs
browser history
shared links
This scenario helps demonstrate that UUIDs alone do not prevent IDOR vulnerabilities.

2. IDOR in REST API Updates
Another common case is when APIs allow modification of resources without verifying ownership.
Example:
PUT /api/users/{id}
{
  "email": "attacker@email.com"
}
If authorization is missing, users could update other users' information.
This scenario demonstrates both:
IDOR
Horizontal privilege escalation

if there are other things, please add that to IDOR as it is an important vulnerability

anatolii-kandiuk:
Hello! I'm looking to make my first contribution to this project. I would like to work on this issue. Could you please assign it to me? Thank you!

preetkaran20:
@anatolii-kandiuk sure please go ahead.

anatolii-kandiuk:
I am currently working on implementing the Level 6 (UUID) scenario for this issue and would like to align on the architectural approach before submitting a PR.
To implement UUID-based IDOR, we need to map UUIDs to the existing users. However, the current idor_users table schema does not have a UUID column.

Solutions:
1. Database Schema Migration:
Add a UUID column to the database and generate UUIDs for existing test users.
Risk: This introduces a high risk of regression and might break backward compatibility for Levels 1-5, which strictly rely on the existing SQL queries and DB structure.

2. Static In-Memory Mapping:
Create a static in-memory map (Map<String, Integer>) directly inside the IDORVulnerability class to bind 2-3 predefined UUIDs to the existing database user IDs.

Please let me know if you approve this approach or if you'd prefer a full DB schema update.

preetkaran20:
@anatolii-kandiuk adding a newer column to existing database is better. Our database is ephemeral i.e it gets created on each new run of the server so it doesn’t cause any regression.

preetkaran20:
Great thought btw for thinking everything in depth🚀

preetkaran20:
@anatolii-kandiuk if you are working on this, please let us know.

PR #628, opened 2026-05-21
Implement Level 6 (UUID) and Level 7 (REST Update) for IDOR #512

anatolii-kandiuk:
- Level 6 (UUID Based IDOR): 
Added uuid column (VARCHAR(36)) to idor_users table in the database schema.
Created endpoint /LEVEL_6 that blindly trusts the provided UUID parameter, exposing the user profile without authorization checks.

- Level 7 (REST API Updates / Privilege Escalation):
Granted UPDATE privileges to the DB user for the idor_users table.
Created PUT /LEVEL_7 endpoint that accepts a JSON Payload (@RequestBody).
The server processes the update (e.g., changing salary and role) based on the userId provided in the JSON body, leading to Mass Assignment / Privilege Escalation.

## Summary by CodeRabbit

* **New Features**
  * Added three IDOR demo levels: Level 6 — retrieve profiles by optional UUID; Level 7 — update a user’s salary/role via PUT by userId; Level 8 — update a user’s salary/role via PUT by UUID (requires UUID in request).

* **Chores**
  * Added UUID to user model and seed data (expanded user set); DB schema updated with UUID column and granted UPDATE rights; added related payloads and i18n messages.

[Review Change Stack](https://app.coderabbit.ai/change-stack/SasanLabs/VulnerableApp/pull/628?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


src/main/java/org/sasanlabs/service/vulnerability/idor/IDORVulnerability.java
```diff
@@ -234,6 +241,45 @@ public ResponseEntity<GenericVulnerabilityResponseBean<Object>> level5(
         }
     }
 
+    @AttackVector(
+            vulnerabilityExposed = VulnerabilityType.INSECURE_DIRECT_OBJECT_REFERENCE,
+            description = "IDOR_LEVEL_6_UUID",
+            payload = "IDOR_PAYLOAD_LEVEL_6")
```
preetkaran20, line 252: please define the payload and description messages keys in messages.properties file. You can look at other levels for more details.

src/main/java/org/sasanlabs/service/vulnerability/idor/IDORVulnerability.java
```diff
@@ -234,6 +241,45 @@ public ResponseEntity<GenericVulnerabilityResponseBean<Object>> level5(
...
+    @VulnerableAppRequestMapping(value = LevelConstants.LEVEL_6, htmlTemplate = "LEVEL_4/IDOR")
+    public ResponseEntity<GenericVulnerabilityResponseBean<Object>> level6(
+            @RequestParam(required = false) String uuid) {
+
+        if (uuid != null) {
```
preetkaran20: IDOR vulnerability links with User Login so add login setup as well and after login, user can guess and fetch other users.

src/main/java/org/sasanlabs/service/vulnerability/idor/IDORVulnerability.java
```diff
@@ -234,6 +241,45 @@ public ResponseEntity<GenericVulnerabilityResponseBean<Object>> level5(
...
+    @VulnerableAppRequestMapping(
+            value = LevelConstants.LEVEL_7,
+            htmlTemplate = "LEVEL_7/IDOR",
+            requestMethod = RequestMethod.PUT)
+    public ResponseEntity<GenericVulnerabilityResponseBean<Object>> level7(
+            @RequestBody User userToUpdate) {
+
+        int rowsAffected =
+                updateUser(
+                        userToUpdate.getUserId(), userToUpdate.getSalary(), userToUpdate.getRole());
```
preetkaran20: Updating existing data is tricky as it will impact existing levels as well. best to not update existing users which are used in other levels. Can you please create 3 new users and update them from this level. Also if possible showcase update for both UUID and ID levels.

preetkaran20 (changes requested):

src/main/java/org/sasanlabs/service/vulnerability/idor/IDORVulnerability.java
anatolii-kandiuk: Done

src/main/java/org/sasanlabs/service/vulnerability/idor/IDORVulnerability.java
anatolii-kandiuk: Done

src/main/java/org/sasanlabs/service/vulnerability/idor/IDORVulnerability.java
anatolii-kandiuk: Done

src/main/java/org/sasanlabs/service/vulnerability/idor/IDORVulnerability.java
```diff
@@ -252,12 +365,48 @@ private User fetchUserById(int id) {
...
+    }
+
+    private int updateUserById(int id, Integer salary, String role) {
+        if (id <= 3) {
+            return 0;
+        }
+        return jdbcTemplate.update(SQL_UPDATE_PROFILE_BY_ID, new Object[] {salary, role, id});
+    }
+
+    private int updateUserByUuid(String uuid, Integer salary, String role) {
```
preetkaran20, line 391: I think better way is introduce a field in database called Level and update based levels will have there own users which they can manipulate.

src/main/java/org/sasanlabs/service/vulnerability/idor/IDORVulnerability.java
```diff
@@ -252,12 +365,48 @@ private User fetchUserById(int id) {
...
+        if ("747e108e-10b1-463e-9328-c453070a1975".equals(uuid)
```
preetkaran20: If checks in code are not good.

src/main/java/org/sasanlabs/service/vulnerability/idor/IDORVulnerability.java
```diff
@@ -234,6 +246,106 @@ public ResponseEntity<GenericVulnerabilityResponseBean<Object>> level5(
...
+        }
+    }
+
+    @AttackVector(
+            vulnerabilityExposed = VulnerabilityType.INSECURE_DIRECT_OBJECT_REFERENCE,
+            description = "IDOR_LEVEL_7_UPDATE",
+            payload = "IDOR_PAYLOAD_LEVEL_7")
+    @VulnerableAppRequestMapping(
+            value = LevelConstants.LEVEL_7,
+            htmlTemplate = "LEVEL_7/IDOR",
```
preetkaran20, line 282: I dont see these templates in the code. please add the html templates like it is done for other levels. 

Image

preetkaran20 (changes requested):
I would suggest testing these code changes as they are not working.

src/main/resources/scripts/IDOR/db/data.sql
```diff
@@ -1,3 +1,8 @@
...
-INSERT INTO idor_users VALUES (2, 'Bob', '6F1A9D3C7B2E4A10:8944bc8b1b5e87e1d2046a27c8e524c10cf6f6169c2a66cfb19b74513b679307', 60000, 'USER');
-INSERT INTO idor_users VALUES (3, 'Charlie', 'C4E8B1D9A2F63750:ee5282141fdffb8adeb83550ac95436cdf2e5bee3be1353fcf2b284a03eba13e', 70000, 'ADMIN');
+-- Base users (LVL 1-6)
+INSERT INTO idor_users VALUES (1, '324bdf11-ef1a-48f0-b342-436779c24456', 'Alice', 'A1B2C3D4E5F60708:141efd4aa55f37300ad1c06d63854eca83a8d41953c5667b9d24eaff4a3edd1d', 50000, 'USER', 1);
+INSERT INTO idor_users VALUES (2, '747e108e-10b1-463e-9328-c453070a1975', 'Bob', '6F1A9D3C7B2E4A10:8944bc8b1b5e87e1d2046a27c8e524c10cf6f6169c2a66cfb19b74513b679307', 60000, 'USER', 1);
+INSERT INTO idor_users VALUES (3, '7b6b8212-66ff-4193-8915-e35399a37945', 'Charlie', 'C4E8B1D9A2F63750:ee5282141fdffb8adeb83550ac95436cdf2e5bee3be1353fcf2b284a03eba13e', 70000, 'ADMIN', 1);
+-- Users (LVL 7-8)
+INSERT INTO idor_users VALUES (4, '44444444-4444-4444-4444-444444444444', 'David', 'A1B2C3D4E5F60708:141efd4aa55f37300ad1c06d63854eca83a8d41953c5667b9d24eaff4a3edd1d', 10000, 'USER', 7);
```
preetkaran20, line 6: can you please use different passwords across levels?

preetkaran20 (changes requested):
reviewing

src/main/resources/i18n/messages.properties
```diff
@@ -344,6 +344,9 @@ IDOR_LEVEL_2_COOKIE_TAMPERING=Secure login, but the object ID is controlled by a
 IDOR_LEVEL_3_JWT_TAMPERING=Secure login via token, but authorization is based on a client-side role_level3 cookie. Changing this cookie to ADMIN bypasses access controls.
 IDOR_LEVEL_4_BROKEN_RBAC=Secure login with Base64-encoded role cookie, but cookie values can be tampered with client-side. Modify the role_level4 cookie to 'ADMIN' to bypass authorization.
 IDOR_LEVEL_5_SECURE_RBAC=Secure login with proper server-side authorization. The server validates that the user owns the requested resource or has admin privileges before returning data. Cross-user access is blocked.
+IDOR_LEVEL_6_UUID=Secure login, but objects are referenced by UUID instead of sequential IDs. Enumeration is prevented, but authorization is still missing.
```
preetkaran20, line 347: we should not show the UUID in the UI. I am just thinking to show authorisation check is missing, we can have one user with uuid as say 0000000000000 or something like you have set 4444444444 but only one user not all users.

src/main/resources/static/templates/IDORVulnerability/LEVEL_6/IDOR.html
```diff
@@ -0,0 +1,60 @@
...
+        </div>
+    </div>
+
+    <div class="idor-card">
+        <h5>Fetch Profile</h5>
+        <div id="profileSection" class="idor-row">
+            <div class="idor-field" style="flex: 2;">
+                <label for="profileUuid">Profile UUID</label>
+                <select id="profileUuid">
+                    <option value="324bdf11-ef1a-48f0-b342-436779c24456">Alice (324bdf11-ef1a-48f0-b342-436779c24456)</option>
```
preetkaran20, line 38: i think we should not show UUID

src/main/resources/static/templates/IDORVulnerability/LEVEL_6/IDOR.html
```diff
@@ -0,0 +1,60 @@
...
+    <h3 class="idor-title">IDOR Vulnerability (UUID Based)</h3>
+    <p class="idor-subtitle">Login and fetch profile data. This level uses non-sequential UUIDs instead of classic numeric IDs to make direct references harder to guess.</p>
+
+    <div class="idor-card">
+        <h5>Login</h5>
+        <div id="loginSection" class="idor-row">
+            <div class="idor-field">
+                <label for="username">User</label>
+                <select id="username">
+                    <option value="Alice">Alice</option>
```
preetkaran20, line 12: for level 6,7,8 please use users of those levels not other levels.

src/main/resources/static/templates/IDORVulnerability/LEVEL_6/IDOR.html
```diff
@@ -0,0 +1,60 @@
...
+                    <option value="Bob">Bob</option>
+                    <option value="Charlie">Charlie</option>
+                    <option value="David">David</option>
+                    <option value="Eve">Eve</option>
+                    <option value="Frank">Frank</option>
+                </select>
+            </div>
+            <div class="idor-field">
+                <label for="password">Password</label>
+                <input type="password" id="password" value="P@ssw0rd!2026" autocomplete="current-password" />
```
preetkaran20, line 22: have a new password for David and make David selected for these levels.

src/main/resources/static/templates/IDORVulnerability/LEVEL_7/IDOR.html
```diff
@@ -0,0 +1,83 @@
...
+                <label for="lookupMethod">Method</label>
+                <select id="lookupMethod">
+                    <option value="ID" selected>ID</option>
+                    <option value="UUID">UUID</option>
+                </select>
+            </div>
+            <div class="idor-field">
+                <label for="targetUser">Target User</label>
+                <select id="targetUser">
+                    <option value="1" data-uuid="324bdf11-ef1a-48f0-b342-436779c24456">Alice (base)</option>
```
preetkaran20, line 46: same as above. no need to show the UUID

src/main/resources/static/templates/IDORVulnerability/LEVEL_7/IDOR.js
```diff
@@ -0,0 +1,181 @@
...
+  // Client-side protection constraint
+  if (lookupMethod === "ID" && userId <= 3) {
+    showUpdateResponse(
+      "Action denied: Cannot modify protected base users.",
+      false
+    );
+    return;
+  }
+  if (
+    lookupMethod === "UUID" &&
```
preetkaran20, line 140: I would suggest not showing those users at all. Have a check in backend instead of client side, client side checks can be bypassed. Also check should be on type of user.

src/main/resources/static/templates/IDORVulnerability/LEVEL_7/IDOR.js
```diff
@@ -0,0 +1,181 @@
...
+    (uuid === "324bdf11-ef1a-48f0-b342-436779c24456" ||
```
preetkaran20, line 141: please remove UUID based checks. this is very difficult to maintain.

src/main/resources/scripts/IDOR/db/schema.sql
```diff
@@ -2,10 +2,12 @@ DROP TABLE IF EXISTS idor_users;
 
 CREATE TABLE idor_users (
     id INT PRIMARY KEY,
+    uuid VARCHAR(36),
     username VARCHAR(50),
     password VARCHAR(500),
     salary INT, 
-    role VARCHAR(20)
+    role VARCHAR(20),
+    level INT
```
preetkaran20, line 10: have a field called Type. 

2 types are there ID based and UUID based users.

preetkaran20 (changes requested):
**Schema / data layer**
The level column on idor_users is heading in the right direction but the intent isn't clear enough. Replace it with a type column — values ID or UUID. Levels 1–5 query only type='ID' users, levels 6+ query only type='UUID' users. This makes the separation explicit in the data model and removes the fragile level-number comparisons scattered through the SQL (WHERE level=7 etc.).

**No cross-type access — enforce it in the backend**
The backend must filter by type on every query. Level 6's fetchUserByUuid should append AND type='UUID' to its SQL, and levels 1–5's fetchUserById should append AND type='ID'. That way it's physically impossible for a UUID-level request to touch an ID-level user and vice versa, regardless of what the client sends.
**Remove all client-side protection checks**
The JS in IDOR.js has guards like "Cannot modify protected base users" with hardcoded UUID comparisons. These need to go entirely. Client-side checks teach the wrong lesson in a security training app — the whole point is that they're bypassable. If the backend enforces type separation, you don't need any of this.

**Don't show UUIDs in the UI for level 6**
The dropdown in LEVEL_6/IDOR.html currently lists Alice's full UUID next to her name. The UUID should not appear in the UI at all before the user fetches the profile. The hint for how to obtain it should come from the attack vector description — e.g. "you can find other users' UUIDs by inspecting API responses or the database." That's the realistic attack scenario and it's a better learning moment than handing it to the student in a dropdown.


**Level 6/7/8 should only show UUID-type users**
Give the memorable UUID (44444444-4444-4444-4444-444444444444) to a user who is not in the login dropdown — so the student can never obtain it legitimately from their own session. Eve or Frank would work. David logs in, sees his own real UUID in the response, but Eve (or Frank) has the guessable one sitting there waiting to be found. The student has to notice the pattern and try it manually — that's the "aha" moment the level is trying to create. The remaining user (the third one) gets a realistic random UUID so there are only two targets: one guessable, one not.

preetkaran20:
@anatolii-kandiuk let me know if you need any help

anatolii-kandiuk:
Hi! I've updated the backend validation as requested (removed client-side checks and added type validation).
For the new users (David, Eve, Frank), I need to set up their passwords in the idor_users SQL script. What is our standard/preferred way to generate the salted hashes for these initial DB inserts? Should I just run PasswordHashingUtils.sha256Hex() locally and hardcode the results in the SQL file, or do we have a specific script/tool in the project for generating test credentials?

preetkaran20:
> Hi! I've updated the backend validation as requested (removed client-side checks and added type validation). For the new users (David, Eve, Frank), I need to set up their passwords in the idor_users SQL script. What is our standard/preferred way to generate the salted hashes for these initial DB inserts? Should I just run PasswordHashingUtils.sha256Hex() locally and hardcode the results in the SQL file, or do we have a specific script/tool in the project for generating test credentials?

Great question. I think when the first 5 levels were implemented it was manually generated but we have a code which can be used to generate the passwords under database seeders class (https://github.com/aashpis/VulnerableApp/blob/master/src/main/java/org/sasanlabs/configuration/DatabaseSeeder.java). I am not sure if that is perfect for this case too but you can try to use it or enhance it for this use case. 

@antriksh-9 how did we generated other level passwords? Please correct me if I am wrong.

antriksh-9:
> > Hi! I've updated the backend validation as requested (removed client-side checks and added type validation). For the new users (David, Eve, Frank), I need to set up their passwords in the idor_users SQL script. What is our standard/preferred way to generate the salted hashes for these initial DB inserts? Should I just run PasswordHashingUtils.sha256Hex() locally and hardcode the results in the SQL file, or do we have a specific script/tool in the project for generating test credentials?
> 
> Great question. I think when the first 5 levels were implemented it was manually generated but we have a code which can be used to generate the passwords under database seeders class (https://github.com/aashpis/VulnerableApp/blob/master/src/main/java/org/sasanlabs/configuration/DatabaseSeeder.java). I am not sure if that is perfect for this case too but you can try to use it or enhance it for this use case.
> 
> @antriksh-9 how did we generated other level passwords? Please correct me if I am wrong.

@preetkaran
The password hashes used by the previous IDOR levels were pre-generated and stored in data.sql. During runtime, the application only verifies them using `PasswordHashingUtils.isValidSaltedSha256()`.

preetkaran20:
@anatolii-kandiuk you can manually generate it for one time or use the tool to generate it for all the levels if easy. Upto you.

anatolii-kandiuk:
@preetkaran20 Hi! Unfortunately, due to certain circumstances, I am currently unable to work on this task. I apologize for any inconvenience this may cause. Please feel free to reassign this to someone else so as not to block the progress.

As soon as I have some free time, I would love to return and continue contributing to the project. Thank you for understanding!

PR #653, opened 2026-06-09
Added Session Management Vulnerability

### LINKED ISSUE #647: Adding Session Authentication related vulnerabilities
VulnerableApp currently has Authentication related vulnerabilites but it never had session related vulnerabilities.

The goal of this vulnerabilities is to introduce sessions to VulnerableApp.

**Session management**
1. Session fixation (CWE-384) -- Session ID not regenerated after login. Attacker pre-sets the session cookie.
2. Predictable session IDs (CWE-330) -- Session tokens generated from a sequential counter instead of a CSPRNG.
3. Missing logout invalidation (CWE-613) -- Logout clears the cookie but does not destroy the server-side session.
4. No login rate limiting (CWE-307) -- Unlimited login attempts with no delay, lockout, or CAPTCHA.

**Password reset**
5. Predictable reset token -- Token generated from timestamp or sequential ID.
6. Reset token reuse -- Token not invalidated after first use.
7. Reset without identity verification -- No email/security question check before allowing password change.

We are planning to add local SMTP server as well, if this issue is picked after that, please try adding scenarios of password reset with emails.

CHI7440:
**Resolves #647**

This introduces a new **Session Management Vulnerability** module to VulnerableApp. The goal is to add session-related security scenarios that were previously missing, while keeping the implementation aligned with the existing VulnerableApp pattern of vulnerable earlier levels and a secure later level.

The new module is exposed as:

```text
/VulnerableApp/SessionManagementVulnerability
```

It uses annotation-driven discovery through `@VulnerableAppRestController`, `@VulnerableAppRequestMapping`, and `@AttackVector`, so the new levels appear in the legacy UI, scanner metadata, sitemap, and facade definitions.

**Implemented Levels**

`LEVEL_1 - Session Fixation`

Demonstrates CWE-384. The application accepts a client-supplied `VAPP_SESSION_ID` before login and keeps the same session token after authentication. The response shows the pre-login token, post-login token, and whether session fixation was confirmed.

`LEVEL_2 - Predictable Session IDs`

Demonstrates CWE-330. The application generates session IDs using a predictable counter such as `SESSION-1001`, `SESSION-1002`, etc. Users can log in multiple times and check profiles by manually supplied session tokens to demonstrate guessable session IDs.

`LEVEL_3 - Missing Logout Invalidation`

Demonstrates CWE-613. Logout clears the client-side cookie but intentionally does not destroy the server-side session. Reusing the old session token after logout still allows profile access.

`LEVEL_4 - No Login Rate Limiting`

Demonstrates CWE-307. The application allows repeated failed login attempts without throttling, lockout, delay, or CAPTCHA. Responses explicitly show that attempts remain allowed and rate limiting is not applied.

`LEVEL_5 - Secure Session Management`

Adds a secure comparison level marked with `Variant.SECURE`. This level regenerates the session token after login, uses random UUID session IDs, invalidates server-side sessions on logout, and applies a simple username-based failed-login limit.

**Database Support**

Added a new `session_users` table for this module instead of reusing `auth_users`, keeping session-management scenarios independent from authentication vulnerabilities.

Seeded users:

```text
session_admin / password123 / ADMIN
session_user / password123 / USER
```

**Frontend Templates**

Added separate legacy UI templates for each level under:

```text
src/main/resources/static/templates/SessionManagementVulnerability/
```

Each level has focused HTML/CSS/JS to keep the UI simple and avoid level-specific conditionals.

**Scanner Metadata**

Added new vulnerability types:

```text
SESSION_FIXATION
PREDICTABLE_SESSION_ID
MISSING_LOGOUT_INVALIDATION
NO_LOGIN_RATE_LIMITING
```

Added corresponding attack-vector payloads and i18n descriptions.

**Tests**

Added focused service tests covering all implemented levels:

```text
SessionManagementServiceTest
```

The tests verify vulnerable and secure behavior for session fixation, predictable IDs, logout invalidation, missing rate limiting, secure token regeneration, secure logout invalidation, and failed-login blocking.

**Password Reset And SMTP Scope**

This PR focuses only on session-management vulnerabilities. Password reset vulnerabilities and SMTP-based reset flows are intentionally left out of scope because they represent a separate vulnerability area and require different user flows, backend logic, and possibly SMTP infrastructure.

Those scenarios should be handled in a follow-up issue/PR.

image

## Summary by CodeRabbit

* **New Features**
  * Added a new Session Management vulnerability module with level-based login, profile, and logout flows (including a secure Level 6 flow).
  * Introduced session user storage plus automatic admin database initialization (schema and seed data).
  * Added dedicated UI templates and styling for the session management challenges.

* **Bug Fixes**
  * Implemented behaviors such as session invalidation and login failure rate limiting in the relevant flows.

* **Documentation**
  * Added in-app messaging and challenge prompts/hints for session management levels.

* **Tests**
  * Added comprehensive service-level tests covering Levels 1–6.


src/main/resources/static/templates/SessionManagementVulnerability/LEVEL_5/SecureSessionManagement.js
```diff
@@ -0,0 +1,60 @@
...
+      encodeURIComponent(username) +
+      "&password=" +
+      encodeURIComponent(password);
+
+    doGetAjaxCall(showSessionResponse, url, true);
+  });
+
+document.getElementById("profileBtn").addEventListener("click", function () {
+  var token = document.getElementById("sessionToken").value;
+  doGetAjaxCall(
```
preetkaran20: please use doPost instead of doGet as doGet is not secure because urls can be logged in access logs, server history.

src/main/resources/static/templates/SessionManagementVulnerability/LEVEL_5/SecureSessionManagement.js
```diff
@@ -0,0 +1,60 @@
...
+    2
+  );
+
+  if (data.isValid && data.content && data.content.sessionId) {
+    document.getElementById("sessionToken").value = data.content.sessionId;
+  }
+}
+
+function setSessionCookie(token) {
+  document.cookie =
```
preetkaran20: Javascript should not manipulate cookies as that is a vulnerability. lets send it from backend. If you want, you can create a unsecure level were cookie is not having flag called HTTP_ONLY and that will be a security vulnerability and secure level has that flag set.

src/main/java/org/sasanlabs/service/vulnerability/sessionManagement/SessionManagementService.java
```diff
@@ -0,0 +1,237 @@
...
+import org.sasanlabs.service.vulnerability.bean.GenericVulnerabilityResponseBean;
+import org.springframework.http.HttpHeaders;
+import org.springframework.http.ResponseCookie;
+import org.springframework.http.ResponseEntity;
+import org.springframework.stereotype.Service;
+
+@Service
+public class SessionManagementService {
+
+    public static final String COOKIE_NAME = "VAPP_SESSION_ID";
```
preetkaran20: I think it is better to have different cookie for each level as that will give good segregation and will help against interference of levels with cookies.

src/main/java/org/sasanlabs/service/vulnerability/sessionManagement/SessionManagementVulnerability.java
```diff
@@ -0,0 +1,149 @@
...
+            @RequestParam(required = false) String action,
+            @RequestParam(required = false) String username,
+            @RequestParam(required = false) String password,
+            @RequestParam(required = false) String token,
+            @CookieValue(value = SessionManagementService.COOKIE_NAME, required = false)
+                    String sessionId) {
+        if (LOGIN_ACTION.equals(action)) {
+            return sessionManagementService.level5Login(username, password, sessionId);
+        }
+        if (PROFILE_ACTION.equals(action)) {
```
preetkaran20: I think we don't need profile action. on successful login, we should send profile information to the user.

src/main/java/org/sasanlabs/service/vulnerability/sessionManagement/SessionManagementVulnerability.java
```diff
@@ -0,0 +1,149 @@
...
+            description = "NO_LOGIN_RATE_LIMITING_DESCRIPTION",
+            payload = "NO_LOGIN_RATE_LIMITING_PAYLOAD_LEVEL_4")
+    @VulnerableAppRequestMapping(
+            value = LevelConstants.LEVEL_4,
+            htmlTemplate = "LEVEL_4/NoLoginRateLimiting")
+    public ResponseEntity<GenericVulnerabilityResponseBean<Object>> level4NoLoginRateLimiting(
+            @RequestParam(required = false) String action,
+            @RequestParam(required = false) String username,
+            @RequestParam(required = false) String password) {
+        if (LOGIN_ACTION.equals(action)) {
```
preetkaran20: we dont need action for this level as we have only one action.

src/main/java/org/sasanlabs/service/vulnerability/sessionManagement/SessionManagementVulnerability.java
```diff
@@ -0,0 +1,149 @@
...
+            return sessionManagementService.level4Login(username, password);
+        }
+        return ResponseEntity.ok(
```
preetkaran20: If level 3 has logout and login, shall we have that functionality at least in this too?

preetkaran20: Levels are adding layers of security, with each next level security is added.

src/main/java/org/sasanlabs/service/vulnerability/sessionManagement/SessionManagementVulnerability.java
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    @AttackVector(
+            vulnerabilityExposed = VulnerabilityType.PREDICTABLE_SESSION_ID,
+            description = "PREDICTABLE_SESSION_ID_DESCRIPTION",
+            payload = "PREDICTABLE_SESSION_ID_PAYLOAD_LEVEL_2")
+    @VulnerableAppRequestMapping(
+            value = LevelConstants.LEVEL_2,
+            htmlTemplate = "LEVEL_2/PredictableSessionId")
+    public ResponseEntity<GenericVulnerabilityResponseBean<Object>> level2PredictableSessionId(
```
preetkaran20: any reason for not having logout ?

src/main/java/org/sasanlabs/service/vulnerability/sessionManagement/SessionManagementService.java
```diff
@@ -0,0 +1,237 @@
...
+    private static final String COOKIE_PATH = "/VulnerableApp";
+
+    private final Map<String, SessionUser> sessions = new ConcurrentHashMap<>();
+    private final Map<String, Integer> level5FailedAttempts = new ConcurrentHashMap<>();
```
preetkaran20: what is the frequency of cleaning this level5FailedAttempts? Find out some simple way to build a ttl based cache or a libvrary that supports it.

src/main/java/org/sasanlabs/service/vulnerability/sessionManagement/SessionManagementService.java
```diff
@@ -0,0 +1,237 @@
...
+            content.put("sessionFixationConfirmed", incomingSessionId.equals(sessionId));
+            content.put("vulnerability", "Client-supplied session ID was kept after login");
+        }
+
+        return ResponseEntity.ok()
+                .header(HttpHeaders.SET_COOKIE, buildSessionCookie(sessionId).toString())
+                .body(new GenericVulnerabilityResponseBean<>(content, true));
+    }
+
+    public ResponseEntity<GenericVulnerabilityResponseBean<Object>> level2Login(
```
preetkaran20: we can have 2 levels with this, one where we use simple incrementing session id and one where we little obscure session id which is incrementing by a factor of say 1000 or something else but may be base64 encoded or something like that.

src/main/java/org/sasanlabs/service/vulnerability/sessionManagement/SessionManagementService.java
```diff
@@ -0,0 +1,237 @@
...
+import org.springframework.http.ResponseEntity;
+import org.springframework.stereotype.Service;
+
+@Service
+public class SessionManagementService {
+
+    public static final String COOKIE_NAME = "VAPP_SESSION_ID";
+    private static final String COOKIE_PATH = "/VulnerableApp";
+
+    private final Map<String, SessionUser> sessions = new ConcurrentHashMap<>();
```
preetkaran20: I think we should ensure one level's login/session doesn't impact other levels login so sessions map key should have level as prefix or suffix.

src/main/java/org/sasanlabs/service/vulnerability/sessionManagement/SessionManagementService.java
```diff
@@ -0,0 +1,237 @@
...
+                sessionUserRepository.findByUsernameAndPassword(username, password);
+        if (user.isEmpty()) {
+            return response("Invalid credentials", false);
+        }
+
+        String sessionId = "SESSION-" + predictableSessionCounter.incrementAndGet();
+        sessions.put(sessionId, user.get());
+
+        Map<String, Object> content = new LinkedHashMap<>();
+        content.put("message", "Logged in successfully");
```
preetkaran20: please create string constants for these strings as they are common across levels

preetkaran20 (changes requested):

src/main/resources/static/templates/SessionManagementVulnerability/LEVEL_1/SessionManagement.html
```diff
@@ -0,0 +1,43 @@
+<div id="session_management_ui" class="session-shell">
+    <h3 class="session-title">Session Management Vulnerability</h3>
+    <p class="session-subtitle">Level 1 keeps a pre-set session token after login.</p>
+
+    <div class="session-card">
+        <h5>Prepare Session</h5>
+        <div class="session-row">
+            <div class="session-field">
+                <label for="sessionToken">Session token</label>
```
preetkaran20: no need to give a text box for this. let user of the application do it from the browser console.

src/main/resources/scripts/SessionManagement/db/data.sql
```diff
@@ -0,0 +1,2 @@
+INSERT INTO session_users VALUES (1, 'session_admin', 'password123', 'ADMIN');
```
preetkaran20: I need to think on the best way to handle login and credentials. I will think more and let you know.

src/main/resources/i18n/messages.properties
```diff
@@ -416,6 +416,18 @@ AUTH_PAYLOAD_LEVEL_9=This is the secure implementation. Try to enumerate usernam
 AUTH_LEVEL_10_LOW_ITERATION_HASHING=The application utilizes a modern adaptive hashing algorithm (BCrypt) but configures it with a dangerously low work factor/cost parameter (Cost: 4). This significantly reduces the computational cycles required to generate a hash, making the password database highly vulnerable to rapid offline brute-force and dictionary attacks if compromised.
 AUTH_PAYLOAD_LEVEL_10=Submit standard credentials (e.g., admin/admin) to witness the computational speed and exposure of the weak BCrypt hash returned in the metadata payload.
 
+# Session Management Vulnerability
+SESSION_MANAGEMENT_VULNERABILITY=<b>Session Management Vulnerabilities</b> occur when an application handles session tokens insecurely. Weaknesses such as keeping the same token after login, predictable session IDs, missing logout invalidation, or missing login throttling can allow attackers to hijack accounts or abuse authentication flows.<br/><br/>\
```
preetkaran20: This is good. can you please add payloads and hints as well for exploiting these levels ?

preetkaran20 (changes requested):
Thank you for the PR @CHI7440. Please let me know if you have any questions.

src/main/resources/static/templates/SessionManagementVulnerability/LEVEL_5/SecureSessionManagement.js
CHI7440: `doPost `instead of `doGet`
I agree and have already aligned the implementation accordingly. The login/logout actions now use POST where appropriate so credentials and sensitive session data are not exposed in query strings.

src/main/resources/static/templates/SessionManagementVulnerability/LEVEL_5/SecureSessionManagement.js
CHI7440: JavaScript should not manipulate cookies
Agreed. We removed client-side cookie manipulation and shifted cookie handling to the backend. Also, the `HttpOnly` flag is used for the secure levels, while the intentionally kept no `HttpOnly` in Level 1 leading to the insecure behavior for demonstration.

src/main/java/org/sasanlabs/service/vulnerability/sessionManagement/SessionManagementService.java
CHI7440: Agreed. We use separate cookie names per level so the flows stay isolated and one level does not interfere with another.

src/main/java/org/sasanlabs/service/vulnerability/sessionManagement/SessionManagementVulnerability.java
CHI7440: Agreed for the secure flow where applicable. For level 5 where the goal is to show the session behavior directly after login, the response now includes the relevant profile/session information instead of requiring a separate profile action.

src/main/java/org/sasanlabs/service/vulnerability/sessionManagement/SessionManagementVulnerability.java
CHI7440: We've added profile as well as logout action here to progressively show the security in levels

src/main/java/org/sasanlabs/service/vulnerability/sessionManagement/SessionManagementVulnerability.java
CHI7440: Agreed. The flow is being structured so that each next level improves on the previous one and includes the earlier protections where appropriate.

src/main/java/org/sasanlabs/service/vulnerability/sessionManagement/SessionManagementVulnerability.java
CHI7440: There is no functional security reason to exclude logout at Level 2. Considering Level 2 as part of the progression, we have added logout there as well so the level set remains consistent.

src/main/java/org/sasanlabs/service/vulnerability/sessionManagement/SessionManagementService.java
CHI7440: Good point. We have considered TTL-based cleanup for failed-attempt tracking, we can also use small cache library such as Caffeine would be a reasonable fit if we want to make that explicit.

src/main/java/org/sasanlabs/service/vulnerability/sessionManagement/SessionManagementService.java
CHI7440: Agreed. That is a sensible split: level 2 now demonstrates a straightforward counter-based token, and level 3 shows a weakly obscured but still predictable token generation pattern.

src/main/java/org/sasanlabs/service/vulnerability/sessionManagement/SessionManagementService.java
CHI7440: Agreed. We have ensured session storage is level-scoped with the help of `sessionKey` function to create level specific key so one level’s login/session does not affect another level’s state.

src/main/java/org/sasanlabs/service/vulnerability/sessionManagement/SessionManagementService.java
CHI7440: Added 'SUCCESSFUL_LOGIN_MESSAGE' constant variable

src/main/resources/static/templates/SessionManagementVulnerability/LEVEL_1/SessionManagement.html
CHI7440: Agreed. For the vulnerable Level 1 flow, the token will be set through the browser console/devtools rather than adding a dedicated input field in the UI.

src/main/resources/scripts/SessionManagement/db/data.sql
CHI7440: Understood. We will revisit the credential setup separately and align it with the cleanest approach for the login flow.

src/main/resources/i18n/messages.properties
CHI7440: Agreed. We have added the base descriptions, and we can also add payload guidance and exploitation hints for the purpose of challenge

src/main/resources/i18n/messages.properties
```diff
@@ -472,6 +472,51 @@ AUTHENTICATION_VULNERABILITY_LEVEL_10_HINT_3=Because the cost is only 4, a commo
...
+SESSION_MANAGEMENT_LEVEL5_PAYLOAD_VALUE=username=admin&password=wrongpassword (repeat multiple times)
+
 # Cache Poisoning Vulnerability
 CACHE_POISONING_VULNERABILITY=Web Cache Poisoning occurs when an attacker manipulates a shared web cache\u2014such as a <b>CDN</b>, <b>Reverse Proxy</b>, or <b>Load Balancer</b>\u2014to store a malicious or unintended response. These systems are designed to optimize performance by serving cached versions of frequently requested resources to reduce latency and backend load. The attack succeeds when there is a discrepancy between the <b>Cache Key</b> (the set of request attributes the cache uses to identify a resource) and the application's processing logic. By manipulating <b>unkeyed inputs</b> (headers, cookies, or parameters that are reflected in the response but not included in the cache key), an attacker can "poison" the cache for a specific resource. Once stored, this poisoned response is served to all subsequent users requesting that resource until the cache entry expires or is cleared.<br/><br/>\
 Important Links:<br/>\
```
preetkaran20, line 477: @CHI7440 can you please add more details to description example:

src/main/resources/i18n/messages.properties
```diff
@@ -472,6 +472,51 @@ AUTHENTICATION_VULNERABILITY_LEVEL_10_HINT_3=Because the cost is only 4, a commo
 AUTHENTICATION_VULNERABILITY_LEVEL_10_PAYLOAD_DESCRIPTION=Read the low-cost BCrypt hash from the database, crack it offline with a wordlist, and log in.
 AUTHENTICATION_VULNERABILITY_LEVEL_10_PAYLOAD_VALUE=SELECT username,password FROM auth_users WHERE username='admin_lowcost' (run in the /h2 console) - note the $2a$04$ cost factor, then hashcat -m 3200.
 
+# Session Management Vulnerability
+SESSION_MANAGEMENT_VULNERABILITY=<b>Session Management Vulnerabilities</b> occur when an application handles session tokens insecurely. Weaknesses such as keeping the same token after login, predictable session IDs, missing logout invalidation, or missing login throttling can allow attackers to hijack accounts or abuse authentication flows.<br/><br/>\
+Important Links:<br/>\
+<ol> <li> <a href="https://owasp.org/www-community/attacks/Session_fixation" target="_blank">OWASP Session Fixation</a> \
```
preetkaran20: @CHI7440 can you please add more details to description like 

Session Management Vulnerabilities occur when an application fails to securely create, maintain, validate, rotate, or terminate user sessions. Since session tokens act as proof of authentication, weaknesses in session handling can allow attackers to impersonate legitimate users and gain unauthorized access to protected resources.
Common issues include session fixation, predictable or exposed session identifiers, insecure cookie configurations, missing session rotation after login, insufficient session expiration, improper logout invalidation, and lack of concurrent session control. These vulnerabilities can lead to session hijacking, account takeover, privilege escalation, and long-term unauthorized access to sensitive application functionality.

preetkaran20:
image
I think showing response json is not needed, nor we need show profile button. I am thinking like say facebook, when we login, it just shows the profile to us. So when user logs in, we should show a simple table about the user like username, email etc.

src/main/resources/i18n/messages.properties
```diff
@@ -472,6 +472,51 @@ AUTHENTICATION_VULNERABILITY_LEVEL_10_HINT_3=Because the cost is only 4, a commo
...
+SESSION_MANAGEMENT_LEVEL4_HINT_1=Cookies are cleared on logout, but the server-side session remains active. Reusing the old session token will allow access to the authenticated profile.
+SESSION_MANAGEMENT_LEVEL4_HINT_2=Server side session is not invalidated on logout. You can try to reuse the old session token after logging out to access the profile.
+SESSION_MANAGEMENT_LEVEL4_HINT_3=Log in, log out, and then try to access the profile with the old session token. You should still be able to access the profile.
+SESSION_MANAGEMENT_LEVEL4_PAYLOAD_DESCRIPTION=Log in to the application, note the session token, log out, enter the token in text box and click on check profile. You should still be able to access the profile with the old token.
+SESSION_MANAGEMENT_LEVEL4_PAYLOAD_VALUE=Log in first time -> session token = <random_token>, log out, enter <random_token> in the text box and click on check profile.
+SESSION_MANAGEMENT_LEVEL5_CHALLENGE=Attempt to log in with incorrect credentials multiple times. The application allows repeated login attempts without delay or lockout.
```
preetkaran20: I think challenge is to let user know what to do like can you find the users password? and then hint should be that you can bruteforce and findout the password for the user.

preetkaran20: we need to think about challenges from this perspective where we challenge user. I am unable to find how we can do this for other levels. can you please give a thought on this?

preetkaran20 (changes requested):
I think there is no need to show anything to the user,. the UI interface is simple login, logout and login fetches user's profile. I think user can manipulate browser developer tools and learn. Our hints, challenge card will tell user what to do.

Image

src/main/resources/i18n/messages.properties
```diff
@@ -472,6 +472,51 @@ AUTHENTICATION_VULNERABILITY_LEVEL_10_HINT_3=Because the cost is only 4, a commo
...
+PREDICTABLE_SESSION_ID_DESCRIPTION=The application generates session IDs using a predictable counter instead of secure random values. Attackers can guess nearby session tokens and access active sessions.
+OBSCURED_PREDICTABLE_SESSION_ID_DESCRIPTION=The application generates session IDs using a predictable counter and a weak encoding methods. Attackers can predict or brute-force session tokens to access active sessions.
+MISSING_LOGOUT_INVALIDATION_DESCRIPTION=The application clears the browser cookie during logout but does not destroy the server-side session. Reusing the old session token still allows access to the authenticated profile.
+NO_LOGIN_RATE_LIMITING_DESCRIPTION=The application allows repeated login attempts without delay, lockout, throttling, or CAPTCHA. Attackers can keep guessing credentials without restriction.
+
+#Challenge Cards for Session Management Vulnerabilities
+SESSION_MANAGEMENT_LEVEL1_CHALLENGE=Log in with a session token you supply before authentication. The application does not regenerate the session token after login.
```
preetkaran20: I think we can have challenge like:

You're an attacker with 5 minutes of physical access to the victim's computer — but you don't know their username or password. Your goal: gain access to their authenticated session once they log in, without ever learning their credentials.
Challenge: Using only the 5 minutes of access, plant something on the victim's browser that will let you hijack their session the moment they log in.

Hints (progressive)
1. You don't need their password if you can get them to use a session token you already control.
2. VAPP_SESSION_ID cookie to a value of your choosing — note it down somewhere you can access later (write it down, or set it to something memorable).
3. Walk away. When the victim logs in, check the application using the same session ID — if you're suddenly viewing their profile, you've confirmed the session was never rotated after login.
4. say victim Logins (for now you can login as victim) 
5. then call the login api again and you can get the user profile details without needing user login details.

src/main/resources/i18n/messages.properties
```diff
@@ -472,6 +472,51 @@ AUTHENTICATION_VULNERABILITY_LEVEL_10_HINT_3=Because the cost is only 4, a commo
...
+SESSION_MANAGEMENT_LEVEL1_HINT_1=The application accepts a session token from the client before login and continues to use that same token after successful authentication.
+SESSION_MANAGEMENT_LEVEL1_HINT_2=Use the console to intercept the login request and modify the session token to a value you control.
+SESSION_MANAGEMENT_LEVEL1_HINT_3=Set the cookie for level1 by using document.cookie to set a session token in the browser before logging in.
+SESSION_MANAGEMENT_LEVEL1_PAYLOAD_DESCRIPTION=Just write document.cookie = "VAPP_SESSION_LEVEL1=your_custom_value" in the browser console before logging in. After login, check if the session token remains the same.
+SESSION_MANAGEMENT_LEVEL1_PAYLOAD_VALUE=document.cookie = "VAPP_SESSION_LEVEL1=your_custom_value"
+SESSION_MANAGEMENT_LEVEL2_CHALLENGE=Log in and observe the session token. Then log out and Log in again. The session token is predictable and increments by 1 each time.
```
preetkaran20: you can build challenegs like a real user challenge. this is a good way to make users learn about vulnerabilities.

src/main/resources/scripts/SessionManagement/db/data.sql
```diff
@@ -0,0 +1,2 @@
...
+INSERT INTO session_users VALUES (2, 'session_user', 'password123', 'USER');
```
preetkaran20: I thought about this. i think we can have 1 attacker user with a simple password and multiple victim users with different passwords for each level. 

Level 5 can have victim user with hard password which attacker will try to guess it. i think other levels can also have one victim with similarly hard password so they can have 2 challenges one like session fixation for level 1 and one as bruteforce for level 1 as well. 

i.e. level 1, level 2, level 3 and so on all have bruteforce issues. Level 6 will not have bruteforce applicable.

Also can we make the structure of table as:

CREATE TABLE session_users (
    id INT PRIMARY KEY,
    username VARCHAR(50) NOT NULL,
    password VARCHAR(255) NOT NULL,
    role VARCHAR(20) NOT NULL,
    level INT,
    is_attacker BOOLEAN DEFAULT FALSE
);

CHI7440:
Unfortunately, while syncing my fork with the upstream repository, I accidentally used GitHub's **Sync fork → Discard commits** option, which reset my branch and caused GitHub to automatically close this PR.

I've recovered the commits and recreated this contribution in **#680**.

I've already carried over the implemented review changes, and it would be helpful if we continue discussions in the new PR

I sincerely apologize for the inconvenience, and thank you for your patience.

PR #708, opened 2026-07-23
feat: add challenge mode to Error-Based and Union-Based SQL Injection

### LINKED ISSUE #561: Updating sql injection
[SQL Injection](https://github.com/SasanLabs/VulnerableApp/tree/master/src/main/java/org/sasanlabs/service/vulnerability/sqlInjection)
[Error Based SQLi](https://github.com/SasanLabs/VulnerableApp/blob/master/src/main/java/org/sasanlabs/service/vulnerability/sqlInjection/ErrorBasedSQLInjectionVulnerability.java)
[Union Based SQLi](https://github.com/SasanLabs/VulnerableApp/blob/master/src/main/java/org/sasanlabs/service/vulnerability/sqlInjection/UnionBasedSQLInjectionVulnerability.java)
[Blind SQLi](https://github.com/SasanLabs/VulnerableApp/blob/master/src/main/java/org/sasanlabs/service/vulnerability/sqlInjection/BlindSQLInjectionVulnerability.java)

Mubashir78:
## Summary

Add `@ChallengeCard` annotations to all unsecure levels in `ErrorBasedSQLInjectionVulnerability` and `UnionBasedSQLInjectionVulnerability`, following the established pattern from `BlindSQLInjectionVulnerability`.

### `ErrorBasedSQLInjectionVulnerability.java`
- Added `@ChallengeCard` annotations to 4 unsecure levels (Level1-Level4)
- Each includes challenge text, 2 hints, and payload description/value

### `UnionBasedSQLInjectionVulnerability.java`
- Added `@ChallengeCard` annotations to 2 unsecure levels (Level1-Level2)
- Same structure as above

### `messages.properties`
- Added 30 new message keys for challenge text, hints, and payloads

Closes #561

## Summary by CodeRabbit

* **New Features**
  * Added challenge cards for **Error-Based SQL Injection** levels 1–4.
  * Added challenge cards for **Union-Based SQL Injection** levels 1–2.
  * Each card includes challenge text, progressive hints, and an example payload for the corresponding level.
* **Localization**
  * Added localized (i18n) labels and guidance for the new SQL injection challenge modes, including payload descriptions and examples.

### DIFF
--- a/src/main/resources/i18n/messages.properties
+++ b/src/main/resources/i18n/messages.properties
@@ -292,18 +292,18 @@ ERROR_SQL_INJECTION_VULNERABILITY_LEVEL2_CHALLENGE=Extract all car information w
 ERROR_SQL_INJECTION_VULNERABILITY_LEVEL2_HINT1=The input is placed inside single quotes in the query: WHERE id='your_input'.
 ERROR_SQL_INJECTION_VULNERABILITY_LEVEL2_HINT2=Close the opening single quote first, then inject your own SQL condition, and comment out the rest.
 ERROR_SQL_INJECTION_VULNERABILITY_LEVEL2_PAYLOAD_DESCRIPTION=Payload that breaks out of the single-quoted string and injects a boolean condition.
-ERROR_SQL_INJECTION_VULNERABILITY_LEVEL2_PAYLOAD_VALUE=' OR '1'='1 --
+ERROR_SQL_INJECTION_VULNERABILITY_LEVEL2_PAYLOAD_VALUE=' OR '1'='1' -- 
 
-ERROR_SQL_INJECTION_VULNERABILITY_LEVEL3_CHALLENGE=Extract car information when single quotes are stripped from the input before query construction.
+ERROR_SQL_INJECTION_VULNERABILITY_LEVEL3_CHALLENGE=Analyze the effectiveness of the naive single-quote stripping used to protect this endpoint.
 ERROR_SQL_INJECTION_VULNERABILITY_LEVEL3_HINT1=The application removes all single quote characters from the input using replaceAll before building the query.
-ERROR_SQL_INJECTION_VULNERABILITY_LEVEL3_HINT2=Since single quotes are stripped, try using a double-encoding or alternate bypass to inject SQL without relying on single quotes.
-ERROR_SQL_INJECTION_VULNERABILITY_LEVEL3_PAYLOAD_DESCRIPTION=Payload that bypasses single-quote stripping to inject SQL.
+ERROR_SQL_INJECTION_VULNERABILITY_LEVEL3_HINT2=Because all single quotes are stripped, standard quote-based injection is not possible. The input remains a quoted literal in the final query.
+ERROR_SQL_INJECTION_VULNERABILITY_LEVEL3_PAYLOAD_DESCRIPTION=The single-quote stripping prevents standard injection. The input becomes a harmless quoted literal.
 ERROR_SQL_INJECTION_VULNERABILITY_LEVEL3_PAYLOAD_VALUE=1 OR 1=1 --
 
-ERROR_SQL_INJECTION_VULNERABILITY_LEVEL4_CHALLENGE=Bypass a PreparedStatement that still constructs the query by concatenating input into the SQL string before preparing it.
-ERROR_SQL_INJECTION_VULNERABILITY_LEVEL4_HINT1=Although a PreparedStatement object is created, the query string itself is built by concatenation with user input.
-ERROR_SQL_INJECTION_VULNERABILITY_LEVEL4_HINT2=Because the user input is embedded in the query text rather than bound as a parameter, the same injection techniques apply despite using PreparedStatement.
-ERROR_SQL_INJECTION_VULNERABILITY_LEVEL4_PAYLOAD_DESCRIPTION=Payload that exploits concatenated input in a misused PreparedStatement query.
+ERROR_SQL_INJECTION_VULNERABILITY_LEVEL4_CHALLENGE=Analyze the misused PreparedStatement that strips single quotes and concatenates input into the query string.
+ERROR_SQL_INJECTION_VULNERABILITY_LEVEL4_HINT1=A PreparedStatement object is created, but the query string is built by concatenation with user input rather than using parameterized binding.
+ERROR_SQL_INJECTION_VULNERABILITY_LEVEL4_HINT2=Single quotes are stripped from input before concatenation. While this prevents standard injection, the code pattern is still insecure and the PreparedStatement provides no actual protection.
+ERROR_SQL_INJECTION_VULNERABILITY_LEVEL4_PAYLOAD_DESCRIPTION=The single-quote stripping and misused PreparedStatement prevent standard injection. The input becomes a harmless quoted literal.
 ERROR_SQL_INJECTION_VULNERABILITY_LEVEL4_PAYLOAD_VALUE=1 OR 1=1 --
 
 # Union-Based SQL Injection Challenge Mode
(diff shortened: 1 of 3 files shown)


Mubashir78:
@coderabbitai

Mubashir78:
@coderabbitai Review latest changes pushed on this PR

Mubashir78:
@coderabbitai you need to review the latest commit, not the commit you have already reviewed.

src/main/resources/i18n/messages.properties
```diff
@@ -281,6 +281,44 @@ UNION_SQL_INJECTION_URL_PARAM_REMOVES_SINGLE_QUOTE_WITH_SINGLE_QUOTE_APPENDED_TO
 BLIND_SQL_INJECTION_URL_PARAM_APPENDED_DIRECTLY_TO_QUERY=Query param is directly appended to the Query hence Query can be manipulated.
 BLIND_SQL_INJECTION_URL_PARAM_WRAPPED_WITH_SINGLE_QUOTE_APPENDED_TO_QUERY=Query param is wrapped around "'" and then appended to the SQL Query hence query can be manipulated.
 
+# Error-Based SQL Injection Challenge Mode
+ERROR_SQL_INJECTION_VULNERABILITY_LEVEL1_CHALLENGE=Extract all car information by exploiting error-based SQL injection where the input is appended directly to the query without any wrapping.
```
preetkaran20: these doesn't seems right challenges. I suggest looking at Blind SQL Injection to see how we can really use SQL Injection to extract databases and other details

preetkaran20 (changes requested):

Mubashir78:
- Updated the Error-Based and Union-Based SQL Injection challenges to ask for real data instead of just `Extract all car information`.
- Challenges now target the `admin_sqli` password in `auth_users`, matching the approach taken by the existing Blind SQL Injection challenges.
- Levels 3 and 4 now describe why the single-quote stripping makes standard injection impractical and focus on analyzing the remaining insecure pattern.

src/main/resources/i18n/messages.properties
```diff
@@ -281,6 +281,44 @@ UNION_SQL_INJECTION_URL_PARAM_REMOVES_SINGLE_QUOTE_WITH_SINGLE_QUOTE_APPENDED_TO
...
+ERROR_SQL_INJECTION_VULNERABILITY_LEVEL2_PAYLOAD_DESCRIPTION=Error-based SQL injection with single-quote escape that reads the admin_sqli password through a forced type conversion error.
+ERROR_SQL_INJECTION_VULNERABILITY_LEVEL2_PAYLOAD_VALUE=' AND CAST((SELECT password FROM auth_users WHERE username='admin_sqli') AS int)=1 -- 
+
+ERROR_SQL_INJECTION_VULNERABILITY_LEVEL3_CHALLENGE=Single quotes are stripped from your input before building the query, making standard injection impossible. Analyze why this endpoint is still not secure despite this filter.
+ERROR_SQL_INJECTION_VULNERABILITY_LEVEL3_HINT1=The application removes all single quote characters from the input using replaceAll before building the query.
+ERROR_SQL_INJECTION_VULNERABILITY_LEVEL3_HINT2=Although single quotes are stripped, using PreparedStatement with concatenation is the real issue. Without parameterized binding, other injection vectors may still be exploitable.
+ERROR_SQL_INJECTION_VULNERABILITY_LEVEL3_PAYLOAD_DESCRIPTION=The single-quote stripping prevents standard injection but the concatenation pattern remains insecure.
+ERROR_SQL_INJECTION_VULNERABILITY_LEVEL3_PAYLOAD_VALUE=1 OR 1=1 --
```
preetkaran20, line 301: why not have something similar as challenge and payload for this level too as level 1 and level 2?

src/main/resources/i18n/messages.properties
```diff
@@ -281,6 +281,44 @@ UNION_SQL_INJECTION_URL_PARAM_REMOVES_SINGLE_QUOTE_WITH_SINGLE_QUOTE_APPENDED_TO
...
+
+ERROR_SQL_INJECTION_VULNERABILITY_LEVEL4_CHALLENGE=A PreparedStatement is used but the query is built by concatenating stripped input. Understand why this provides no real protection against injection.
+ERROR_SQL_INJECTION_VULNERABILITY_LEVEL4_HINT1=A PreparedStatement object is created, but the query string is built by concatenation with user input rather than using parameterized binding.
+ERROR_SQL_INJECTION_VULNERABILITY_LEVEL4_HINT2=Single quotes are stripped from input before concatenation. However, the PreparedStatement offers zero protection because the injection happens before the query is prepared, not through parameterized placeholders.
+ERROR_SQL_INJECTION_VULNERABILITY_LEVEL4_PAYLOAD_DESCRIPTION=The misused PreparedStatement provides no protection since concatenation happens before preparation.
+ERROR_SQL_INJECTION_VULNERABILITY_LEVEL4_PAYLOAD_VALUE=1 OR 1=1 --
```
preetkaran20, line 307: same as above one.  this challenge should also be related to retrieving some data from other tables.

src/main/resources/i18n/messages.properties
```diff
@@ -281,6 +281,44 @@ UNION_SQL_INJECTION_URL_PARAM_REMOVES_SINGLE_QUOTE_WITH_SINGLE_QUOTE_APPENDED_TO
...
+ERROR_SQL_INJECTION_VULNERABILITY_LEVEL1_HINT1=The input parameter is concatenated directly into the SQL query string without quotes or escaping. This lets you break the query and control what appears in the error output.
+ERROR_SQL_INJECTION_VULNERABILITY_LEVEL1_HINT2=CAST or CONVERT a value from auth_users to an incompatible type in a subquery. The database will helpfully include that value in the error message when the type conversion fails.
+ERROR_SQL_INJECTION_VULNERABILITY_LEVEL1_PAYLOAD_DESCRIPTION=Error-based SQL injection that forces the database to display admin_sqli's password through a type conversion error.
+ERROR_SQL_INJECTION_VULNERABILITY_LEVEL1_PAYLOAD_VALUE=1 AND CAST((SELECT password FROM auth_users WHERE username='admin_sqli') AS int)=1
+
+ERROR_SQL_INJECTION_VULNERABILITY_LEVEL2_CHALLENGE=The parameter is now wrapped in single quotes. Use the same error-based technique to extract the admin_sqli password from the auth_users table.
```
preetkaran20, line 291: as user has already retrieved password from level 1, i think in this challenges, we can ask user to find out all the tables or may be password for any other user. Scripts are there under https://github.com/SasanLabs/VulnerableApp/blob/master/src/main/resources/scripts

Example:
https://github.com/SasanLabs/VulnerableApp/blob/master/src/main/resources/scripts/Authentication/db/data.sql
https://github.com/SasanLabs/VulnerableApp/blob/master/src/main/resources/scripts/SessionManagement/db/data.sql#L2

src/main/resources/i18n/messages.properties
```diff
@@ -281,6 +281,44 @@ UNION_SQL_INJECTION_URL_PARAM_REMOVES_SINGLE_QUOTE_WITH_SINGLE_QUOTE_APPENDED_TO
 BLIND_SQL_INJECTION_URL_PARAM_APPENDED_DIRECTLY_TO_QUERY=Query param is directly appended to the Query hence Query can be manipulated.
 BLIND_SQL_INJECTION_URL_PARAM_WRAPPED_WITH_SINGLE_QUOTE_APPENDED_TO_QUERY=Query param is wrapped around "'" and then appended to the SQL Query hence query can be manipulated.
 
+# Error-Based SQL Injection Challenge Mode
+ERROR_SQL_INJECTION_VULNERABILITY_LEVEL1_CHALLENGE=Using only the error messages returned by the database, discover the password stored for admin_sqli in the auth_users table. The application never displays table data directly - only errors.
```
preetkaran20, line 285: Ask for any other user from https://github.com/SasanLabs/VulnerableApp/blob/master/src/main/resources/scripts/Authentication/db/data.sql

src/main/resources/i18n/messages.properties
```diff
@@ -281,6 +281,44 @@ UNION_SQL_INJECTION_URL_PARAM_REMOVES_SINGLE_QUOTE_WITH_SINGLE_QUOTE_APPENDED_TO
...
+
+ERROR_SQL_INJECTION_VULNERABILITY_LEVEL4_CHALLENGE=A PreparedStatement is used but the query is built by concatenating stripped input. Understand why this provides no real protection against injection.
+ERROR_SQL_INJECTION_VULNERABILITY_LEVEL4_HINT1=A PreparedStatement object is created, but the query string is built by concatenation with user input rather than using parameterized binding.
+ERROR_SQL_INJECTION_VULNERABILITY_LEVEL4_HINT2=Single quotes are stripped from input before concatenation. However, the PreparedStatement offers zero protection because the injection happens before the query is prepared, not through parameterized placeholders.
+ERROR_SQL_INJECTION_VULNERABILITY_LEVEL4_PAYLOAD_DESCRIPTION=The misused PreparedStatement provides no protection since concatenation happens before preparation.
+ERROR_SQL_INJECTION_VULNERABILITY_LEVEL4_PAYLOAD_VALUE=1 OR 1=1 --
+
+# Union-Based SQL Injection Challenge Mode
+UNION_SQL_INJECTION_VULNERABILITY_LEVEL1_CHALLENGE=Using a UNION-based injection, extract the username and password for admin_sqli from the auth_users table. The input is appended directly with no wrapping.
```
preetkaran20, line 310: we should not use same challenge for each level. please think of something else. you can start the app and access the database and write the challenges. Accessing ways for database: https://github.com/SasanLabs/VulnerableApp/tree/master#connecting-to-embedded-h2-database

pereiravp:
Hey @Mubashir78, was also working on #561 and only saw this after. Sorry for the overlap, should have checked open PRs first.

Ran your payloads against a live instance before deciding what to do. Level 1 works well and leaks the real password through the CAST error, nice technique, better than what I had. The other five don't reproduce for me though.

Level 2 (' AND CAST(...)=1 -- ) fails because id='' breaks the INT conversion before the AND even runs.

Level 3 and 4 (1 OR 1=1 --) give the same generic conversion error as any bad input, quotes stripped or not.

Union Level 1 (1 UNION SELECT...) doesn't work because id=1 also matches the real Audi row, and since it only reads the first result, the union row never surfaces. Needs a nonexistent id like -1.

Union Level 2 hits the same id='' issue as Level 2, and since UnionBasedSQLInjectionVulnerability isn't wrapped in a try/catch it comes back as a raw 500.

Opened #761 with different payloads for the same 6 levels, verified end to end. Couldn't find a real bypass for levels 3/4 at all, tried quote variants, backslash, unicode look-alikes, dollar quoting, so I built those around the verbose error leaking the table and engine instead.

Not trying to step on this, just want whichever one gets merged to actually work. Happy to fold your Level 1 idea into either PR.

PR #721, opened 2026-08-08
Add secure Level 13 implementation for PathTraversal (#313)

### LINKED ISSUE #313: Adding Secure Implementation for PathTraversal Vulnerability
**Describe the enhancement **
Path Traversal Vulnerability doesn't have a secure implementation so we need to add the Secure implementation. This vulnerability, we need to also validate other levels for PathTraversal vulnerability that they are exploitable or not. 

Glimpse of Path traversal vulnerability:
image

** How to Fix **
Add a new Level in https://github.com/SasanLabs/VulnerableApp/blob/master/src/main/java/org/sasanlabs/service/vulnerability/pathTraversal/PathTraversalVulnerability.java similar to how we added the other levels which has all the validations properly implemented.
Adding Variant as secure to the annotation of the level which is VulnerableAppRequestMapping.

nirupam-06:
Closes #313

Adds a new secure Level 13 for PathTraversal, marked `variant = Variant.SECURE`.

**Why this is actually secure, not just labeled secure:**
Levels 7-12 all validate the filename by checking whether it *contains* an allowed name
(`ALLOWED_FILE_NAMES.stream().anyMatch(name -> queryFileName.contains(name))`), combined
with various blacklist checks for `..` and `%2f`. This is bypassable with a null-byte
payload like `../../../../etc/passwd%00UserInfo.json`, since the raw string still
*contains* "UserInfo.json" even though the resolved path points elsewhere.

Level 13 instead does an exact-match allow-list check
(`ALLOWED_FILE_NAMES.contains(fileName)`), with no blacklist logic needed at all.
Since the filename must equal "UserInfo.json" or "OwaspAppInfo.json" exactly, path
traversal sequences, URL encoding, and the null-byte trick are all rejected by
construction.

**Testing:**
- Added `testGetVulnerablePayloadLevel13WithNullFileName`,
  `testGetVulnerablePayloadLevel13WithWrongFileName`,
  `testGetVulnerablePayloadLevel13BlocksNullByteBypass`, and
  `testGetVulnerablePayloadLevel13` to PathTraversalTest.java.
- The bypass test specifically uses the exact null-byte payload that defeats
  Levels 7-12, confirming Level 13 rejects it.
- Full existing test suite still passes (`./gradlew test` → BUILD SUCCESSFUL),
  confirming Levels 1-12 remain

## Summary by CodeRabbit

* **New Features**
  * Added a level-13 path traversal scenario with secure handling for approved filenames.
  * Requests with missing, traversal-based, or null-byte filenames are rejected.
  * Valid approved filenames continue to return successfully.

* **Tests**
  * Added coverage for invalid filenames, traversal attempts, null-byte bypasses, and valid requests.


preetkaran20 (approved):
LGTM Thanks for the PR @nirupam-06

PR #733, opened 2026-08-10
Add GET /scanner/sast and GET /scanner/dast, deprecate bare GET /scanner (#725)

### LINKED ISSUE #725: Add REST /scanner/sast endpoint; deprecate bare /scanner in favor of /scanner/dast
Currently `expected_issues.csv` is a checked-in file for SAST tools. Move this to a REST
endpoint so the comparator (and cross-app aggregation) can consume it the same way as DAST
data, and align naming with the new `/scanner/dast` + `/scanner/sast` convention being rolled
out farm-wide.

**Scope:**
- Add `GET /scanner/sast` returning the same data currently in `expected_issues.csv`
  (filePath, line, cwe, type) as JSON
- Add `GET /scanner/dast` serving the same response as today's bare `GET /scanner`
- Keep bare `GET /scanner` live and functional, marked deprecated in docs/response headers —
  do not remove yet; existing ZAP addons and external scanners may depend on it directly

vianbas:
Closes #725.

Three endpoints after this change. `GET /scanner/dast` and `GET /scanner/sast` are new. The bare `GET /scanner` stays live, serves the same body it always did, and is marked deprecated: over HTTP through the response headers, and in Java through `@Deprecated(forRemoval = true)` on the method.

This follows your three answers on the issue: `numberOfSources` is in the JSON, `/scanner/sast` sits under `@Profile("public")` next to the other read-only ground-truth endpoints, and the CSV moved onto the classpath and is read and cached once at startup.

**What the endpoints return**

Measured against a build of this branch:

| Request | Result |
|---|---|
| `GET /scanner` | 200, body unchanged, plus a `Deprecation` and a `Link` header |
| `GET /scanner/dast` | 200, byte-for-byte the same body, no deprecation headers |
| `GET /scanner/sast` | 200, 154 objects, fields `cwe`, `filePath`, `line`, `numberOfSources`, `type` |

The two headers on the bare endpoint, as returned by a branch build on port 9099:

```
Deprecation: true
Link: <http://localhost:9099/VulnerableApp/scanner/dast>; rel="successor-version"
```

The `Link` value is built from the incoming request, so it names the host the caller actually reached rather than a fixed one.

One SAST object:

```json
{
  "cwe": "CWE-89",
  "filePath": "src/main/java/org/sasanlabs/service/vulnerability/authentication/AuthenticationVulnerability.java",
  "line": 40,
  "numberOfSources": 1,
  "type": "SQL Injection"
}
```

`type` rather than the Java property name `vulnerabilityType`, matching the name `Finding` already uses on the scanner-facing side. SasanLabs/VulnerableApp-php#20 and SasanLabs/VulnerableApp-jsp#17 are open issues asking for the same two endpoints, so the shape that lands here is probably the one they take. A test asserts `type` is present and `vulnerabilityType` is not.

**Ground truth on the classpath**

`scanner/sast/expectedIssues.csv` moved to `src/main/resources/scanner/sast/expectedIssues.csv`, byte for byte, and now ships inside the jar. The Jib `extraDirectories` block that copied `scanner/` into the image is gone, since the CSV needs no special packaging now. I read that as the patching you were referring to; tell me if you meant something else.

`CsvExpectedIssuesProvider` accepts either form. `benchmark.sast.ground-truth.path` defaults to `classpath:scanner/sast/expectedIssues.csv`; drop the prefix and it reads from the filesystem exactly as before, relative to the working directory or absolute. Existing overrides keep working.

The file is parsed once and cached. Warming happens in `@PostConstruct`, so the 154 rows are parsed on the `main` thread before Tomcat starts listening, and no request ever pays for a parse.

What the move buys, measured from a directory that is not the project root: a `master` build answers `POST /scanner/benchmark` with `scanType=SAST` as HTTP 500, while a build of this branch answers `totalExpected=154` and serves `/scanner/sast` normally. That working-directory dependency is what made the CSV awkward to expose as an endpoint.

**Two things worth your call**

*Startup now fails on an unreadable ground truth.* Reading at startup means a path that cannot be read aborts the context. That matches how `EmbeddedLDAPConfig` treats its own startup work, and on the default the file ships inside the jar, so this is only reachable when someone overrides the property. The message names the property so the cause is obvious:

```
IllegalStateException: Could not read the SAST ground truth from 'classpath:typo.csv'.
  Point benchmark.sast.ground-truth.path at a readable CSV, or prefix it with
  'classpath:' to read one bundled in the jar.
Caused by: java.io.FileNotFoundException: class path resource [typo.csv] cannot be
  opened because it does not exist
```

I can make it log the error and keep booting instead.

*`Deprecation: true` follows the expired draft, not the finished RFC.* `draft-dalal-deprecation-header-03` defined `Deprecation = IMF-fixdate / "true"`; that draft has expired and been replaced, and RFC 9745 makes the value a Date, for example `Deprecation: @1688169599`. RFC 9745 also registers `rel="deprecation"` for human-readable documentation, while `successor-version` is registered separately by RFC 5829 for "the successor version in the version history". `true` is what I proposed on the issue so that is what I built, but the applications copying this will copy whichever one lands. Tell me if you would rather it sent a date.

**Where else `/scanner` was still named**

Deprecating an endpoint the application itself keeps calling would not mean much, so the places still naming it moved:

- `benchmark.dast.ground-truth.url`, the self-call the DAST comparator makes by default
- the "Scanner Endpoint" link on the landing page
- the `DastBenchmarkStrategy` javadoc
- `docs/DesignDocumentation.md`

The readiness probe in `.github/workflows/zap-benchmark.yml` is deliberately left on the bare `/scanner`. That workflow starts `sasanlabs/owasp-vulnerableapp:latest` from the registry rather than a build of the branch, so pointing it at `/scanner/dast` would poll an endpoint the published image does not have and fail the job on every run until a new image ships. It is the version-skewed caller the bare endpoint is being kept alive for. Worth moving once the image is republished.

**Docs**

`docs/HOW-TO-USE-VulnerableApp.md` now points DAST tools at `/scanner/dast`, describes `/scanner/sast` alongside the CSV, and lists the JSON field names. Its old field list named a `WASC` column the CSV does not have, so that is corrected too.

`benchmarks/README.md` described the SAST ground truth as a file read relative to the working directory and listed the old value as the default for `benchmark.sast.ground-truth.path`. Both are updated, along with the `benchmark.dast.ground-truth.url` row and the two places naming `/scanner` as the DAST source.

**Tests**

`VulnerableAppRestControllerTest` runs through `MockMvcBuilders.standaloneSetup`, the way `BenchmarkControllerTest` does, so the request mappings, the JSON field names and the response headers are all under test rather than just the method bodies. `CsvExpectedIssuesProviderTest` gains cases for the `classpath:` prefix, a missing classpath resource, the cache, and the startup failure message. The `@Deprecated` marker is asserted by reflection, since nothing a request can observe changes when it is removed.

I checked the suite by breaking things on purpose and confirming a test goes red. Ten changes, ten caught:

| Broken on purpose | Tests that failed |
|---|---:|
| `@JsonProperty("type")` removed | 1 |
| `/scanner/sast` path typo | 2 |
| `/scanner/dast` path typo | 3 |
| `Deprecation` header removed | 1 |
| `@Deprecated(forRemoval = true)` removed | 1 |
| `Link` header removed | 2 |
| `Link` pointed back at `/scanner` | 2 |
| cache never stored | 2 |
| startup failure swallowed | 1 |
| `classpath:` prefix not recognised | 1 |

**Checks**

- `./gradlew clean build` on JDK 17: 447 tests, 0 failures, `spotlessCheck` clean. Every commit builds and passes its tests on its own, with one exception: the merge commit that brings in #737 does not compile against the test that PR added, and the commit straight after it is the fix.
- Fetched all three endpoints from a build of this branch. `/scanner` and `/scanner/dast` bodies hash identically; `/scanner/sast` returns 154 objects with exactly the five field names above.
- Ran a build of `master` and a build of this branch on the same port and compared them. `/scanner` returns the same 18,805 bytes over 115 entries, with the same SHA-256, from both, and `/scanner/metadata` and `sitemap.xml` match as well. The only difference on `/scanner` is the two headers, which `master` does not send.
- The CSV is byte identical to the one on `master` (same SHA-256) and is present in both `VulnerableApp-1.0.0.jar` and `VulnerableApp-plain.jar`.
- `POST /scanner/benchmark` still works both ways off the moved ground truth: `scanType=SAST` reports `totalExpected=154`, `scanType=DAST` reports `totalExpected=153` against the new default URL, the same figure it reported against the old one.
- Started the jar under `SPRING_PROFILES_ACTIVE=public`, with the `DB_ADMIN_*`, `DB_APP_PASSWORD`, `SMTP_*` and `APP_BASE_URL` values `docker-compose.prod.yml` passes, since `application-public.properties` reads them. `/scanner`, `/scanner/dast` and `/scanner/sast` all answer 200, and `/scanner/benchmark` stays 404 because it lives under `unsafe`. The same run against `master` answers 200, 404, 404, 404.
- The startup log shows the CSV parsed once, on `main`, before Tomcat starts. Four later requests to `/scanner/sast` add no further parse.

**Known gap**

There is no Spring context test in the repository, so nothing asserts the *default value* of `benchmark.sast.ground-truth.path`. I verified the classpath default by running the jar, but if someone changed the default back to a filesystem path, `./gradlew build` would stay green and only a running app would show it. Closing that would mean standing up a full context, LDAP and mail included, which felt out of proportion here. I will add it if you would rather have that covered.

## Summary by CodeRabbit

* **New Features**
  * Added dedicated `/scanner/dast` and `/scanner/sast` endpoints.
  * Added JSON responses for SAST expected issues.
  * Added deprecation and successor-link headers to the legacy `/scanner` endpoint.
  * SAST issue data now loads from bundled resources by default, with filesystem paths supported.

* **Documentation**
  * Updated scanner endpoint guidance, configuration examples, and response field documentation.

* **Bug Fixes**
  * Improved startup validation and caching for SAST issue data.


src/main/java/org/sasanlabs/controller/VulnerableAppRestController.java
```diff
@@ -89,23 +108,63 @@ public List<AllEndPointsResponseBean> allEndPointsJsonResponse()
      */
     @GetMapping
     @RequestMapping("/scanner")
-    public List<ScannerResponseBean> getScannerRelatedInformation(HttpServletRequest request)
+    public ResponseEntity<List<ScannerResponseBean>> getScannerRelatedInformation(
```
preetkaran20, line 112: mark this method as deprecated as well via annotation. Mention that it will be deprecated next year in 2027 September.

src/main/java/org/sasanlabs/controller/VulnerableAppRestController.java
```diff
@@ -89,23 +108,63 @@ public List<AllEndPointsResponseBean> allEndPointsJsonResponse()
...
+        return new StringBuilder()
+                .append(request.getScheme()) // http or https
+                .append("://")
+                .append(request.getServerName()) // actual hostname/IP
+                .append(FrameworkConstants.COLON)
+                .append(request.getServerPort()) // actual port
+                .append(FrameworkConstants.SLASH)
+                .append(FrameworkConstants.VULNERABLE_APP)
+                .append(FrameworkConstants.SLASH)
+                .toString();
```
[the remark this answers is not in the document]
vianbas: Verified this locally against `cad4d1a`. The behaviour is real, but it predates
this PR and fixing it properly means touching code this PR does not modify.

**It reproduces.** I added a throwaway test that drives the endpoints through a
non-default context path (`MockMvc` standalone setup, `.contextPath("/customCtx")`)
and printed what the controller builds. Two runs, identical output:

```
appUrl         = https://10.0.0.5:443/VulnerableApp/
getContextPath = /customCtx
requestURI     = /customCtx/scanner
Link header    = <https://10.0.0.5:443/VulnerableApp/scanner/dast>; rel="successor-version"
sitemap <loc>  = https://10.0.0.5:443/VulnerableApp/SQLInjection/LEVEL_1
```

Served under `/customCtx`, all three generated URLs point at `/VulnerableApp/`.
Both halves of your finding hold, the scanner URLs and the sitemap.

**`applicationUrl()` is an extraction, not new logic.** At the merge base
`f15c0b9`, `getScannerRelatedInformation` already assembled the identical string
inline:

```java
        String appUrl =
                new StringBuilder()
                        .append(scheme)
                        .append("://")
                        .append(serverName)
                        .append(FrameworkConstants.COLON)
                        .append(serverPort)
                        .append(FrameworkConstants.SLASH)
                        .append(FrameworkConstants.VULNERABLE_APP)
                        .append(FrameworkConstants.SLASH)
                        .toString();
```

The diff of this file contains exactly two lines mentioning `VULNERABLE_APP`:
one removed from that inline block, one added to the helper.

**The sitemap builder is untouched.** The last line this PR changes in
`VulnerableAppRestController.java` is 167. `sitemapForPassiveScanners` runs from
line 195 to line 237, so the whole tail of the file is unchanged.

**And there is a third site with the same constant**, also untouched:
`DastBenchmarkStrategy` line 49, `private static final String CONTEXT_PATH =
"/VulnerableApp"`, used by `normalizeUrl` to reduce ground-truth and
scanner-finding URLs to a common key. It sits at the same line with the same
text at `f15c0b9`; this PR changes one javadoc line in that file and nothing
else. I ran `normalizeUrl` directly against `/customCtx` inputs, twice, same
result:

```
/customCtx/SQLInjection/LEVEL_1  ->  /customCtx/SQLInjection/LEVEL_1
/SQLInjection/LEVEL_1            ->  /SQLInjection/LEVEL_1
```

Under `/VulnerableApp` those two forms collapse to one key, which is what
`urlNormalization_acceptsAbsoluteRelativeAndContextStrippedForms` asserts. Under
`/customCtx` they no longer do. Whether that changes a real benchmark result
depends on which URL form the scanner emits, which I have not tested. Worth
folding into the same fix either way.

**What this PR does add** is the `Link: <…>; rel="successor-version"` header,
which carries the same hardcoded base. That is a new surface for an old bug, so
if you want it handled here I will not argue.

**Why I would rather split it.** The proper fix spans three sites across two
subsystems, two of which this PR never touched, and needs tests for a deployment
mode nothing in the repo exercises: `server.servlet.context-path` is set to
`/VulnerableApp` in `application.properties` and is not overridden by
`application-public.properties`, `application-unsafe.properties`, any
`docker-compose*.yml`, `Dockerfile.base`, or any workflow. The bug is latent for
every shipped configuration and bites only an operator who overrides the context
path. This PR implements #725, is already +452/-64 across 14 files, and has not
had a human review yet.

I checked the tracker and found no existing issue for it, so I opened
#736 with the reproduction run against `master` rather than this branch.
@preetkaran20 if you would rather it were fixed in this PR, say so and I will
fold it in.

docs/HOW-TO-USE-VulnerableApp.md
```diff
@@ -42,8 +42,8 @@ There are 2 ways in which this project can be built and used:
...
-   2. For SAST, we have added a [ExpectedIssues.csv](https://github.com/SasanLabs/VulnerableApp/blob/master/scanner/sast/expectedIssues.csv) file which has vulnerabilities and their line numbers which SAST tools can use to evaluate themselves.
+   1. For DAST, VulnerableApp exposes an endpoint `http://<baseurl>/VulnerableApp/scanner/dast` which provides information about all the Vulnerabilities present in the VulnerableApp which DAST tools can leverage to evaluate themselves. The bare `http://<baseurl>/VulnerableApp/scanner` endpoint serves the same response and stays available, but it is deprecated: it returns a `Deprecation: true` header and a `Link` header pointing at `/scanner/dast`. New integrations should use `/scanner/dast`.
+   2. For SAST, VulnerableApp exposes an endpoint `http://<baseurl>/VulnerableApp/scanner/sast` which returns the expected issues as JSON — the vulnerabilities and their line numbers which SAST tools can use to evaluate themselves. The same data is the checked-in [ExpectedIssues.csv](https://github.com/SasanLabs/VulnerableApp/blob/master/src/main/resources/scanner/sast/expectedIssues.csv) file.
```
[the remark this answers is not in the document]
vianbas: I do not think this one holds. The case difference is in the link text, not the
link target.

Lines 46 and 69 at `cad4d1a` both link like this:

```
[ExpectedIssues.csv](https://github.com/SasanLabs/VulnerableApp/blob/master/src/main/resources/scanner/sast/expectedIssues.csv)
```

The target is already `expectedIssues.csv`, lowercase, matching both the tracked
file `src/main/resources/scanner/sast/expectedIssues.csv` and
`benchmark.sast.ground-truth.path=classpath:scanner/sast/expectedIssues.csv`
(`application.properties:41`). `ExpectedIssues.csv` is the visible label of the
markdown link and never reaches a filesystem, so there is nothing case-sensitive
to break.

For context on why these two lines changed at all: this PR moves the CSV from
`scanner/sast/expectedIssues.csv` to `src/main/resources/scanner/sast/expectedIssues.csv`
so it ships on the classpath (`git diff -M` reports it as a rename with zero
content change), and both links were updated to follow it.

One real caveat, though not the one flagged: these are `blob/master` links, so
they 404 until this PR merges, because master still has the file at the old root
path. I checked both URLs just now, the new target returns 404 and the old root
path returns 200. It resolves on merge. Happy to switch them to commit-pinned
permalinks instead if reviewers prefer.

preetkaran20 (approved):
LGTM. A small cosmetic change to add `@deprecated` annotation

src/main/java/org/sasanlabs/controller/VulnerableAppRestController.java
vianbas: Added in the latest commit. Quoting it here since the hunk above is from before the change:

```java
     * @deprecated use {@code /scanner/dast}, which serves the same body under the name every
     *     application uses. This path stays until September 2027 so existing scanners keep working;
     *     its responses carry a {@code Deprecation} header and a {@code Link} header naming the
     *     successor.
     */
    @Deprecated(forRemoval = true)
    @GetMapping
    @RequestMapping("/scanner")
```

A reflection test asserts it, since no request-driven test can see an annotation.

One thing I want to get right, because it lands in two places. "it will be deprecated next year in 2027 September" reads two ways for me:

- September 2027 is when `/scanner` goes away. That is what I committed, and the HTTP counterpart would be a `Sunset` header (RFC 8594), which this branch does not send yet.
- September 2027 is when the deprecation takes effect, so it is not deprecated today. Then the javadoc wording is wrong, and `Deprecation` should carry that future date rather than `true`, which also settles the open question in the description.

Which did you mean? I will leave it on the first unless you say otherwise.

preetkaran20: Yes I mean first one.

preetkaran20:
@vianbas  thanks a lot for the PR. Merging it now.

PR #737, opened 2026-08-13
Fix scanner, sitemap and benchmark URL handling ignoring server.servlet.context-path (#736)

### LINKED ISSUE #736: Generated scanner and sitemap URLs ignore server.servlet.context-path
**Describe the bug**

`VulnerableAppRestController` hardcodes `FrameworkConstants.VULNERABLE_APP` into
the URLs it hands to scanners, instead of deriving the path from
`HttpServletRequest.getContextPath()`. Deployed under any context path other than
`/VulnerableApp`, those URLs point outside the deployment.

Two sites on `master` (`766890c`), both in `VulnerableAppRestController.java`:

1. `getScannerRelatedInformation` (line 92), which builds the `appUrl` inline and
   appends the constant at line 105. Serves `GET /scanner`.
2. `sitemapForPassiveScanners` (line 137), which appends the same constant at
   line 164. Serves `GET /sitemap.xml`.

A third site holds the same literal from the other direction:
`DastBenchmarkStrategy` line 49, `private static final String CONTEXT_PATH =
"/VulnerableApp"`, used by `normalizeUrl` to reduce ground-truth and
scanner-finding URLs to a common key.

**To Reproduce**

I could not do this through the UI, so I drove it from a test. Throwaway probe on
a clean checkout of `master` (`766890c`), `MockMvc` standalone setup, requests
built with `.contextPath("/customCtx")` against
`https://10.0.0.5:443/customCtx/scanner` and `.../customCtx/sitemap.xml`. Run
twice, identical both times:

```
appUrl      = https://10.0.0.5:443/VulnerableApp/
sitemap loc = https://10.0.0.5:443/VulnerableApp/SQLInjection/LEVEL_1
```

The request arrives under `/customCtx` and the request's context path is
`/customCtx`, yet both generated URLs still say `/VulnerableApp/`. This is MockMvc
rather than a real container, but `getContextPath()` is the value a fix would
read, and it is already correct there.

For site 3, running `normalizeUrl` directly with `/customCtx` inputs gives:

```
/customCtx/SQLInjection/LEVEL_1  ->  /customCtx/SQLInjection/LEVEL_1
/SQLInjection/LEVEL_1            ->  /SQLInjection/LEVEL_1
```

Under `/VulnerableApp` those two forms collapse to one key, which is what
`urlNormalization_acceptsAbsoluteRelativeAndContextStrippedForms` asserts. Under
any other context path they do not. Whether that changes a real benchmark result
depends on which URL form the scanner emits, which I have not tested.

**Expected behavior**

Generated URLs stay inside the context path the application is actually served
under. Served at `/customCtx`, `GET /scanner` should hand back
`https://host:443/customCtx/...` and `GET /sitemap.xml` should emit `<loc>`
entries under the same prefix.

**Additional context**

Latent for everything in the repo. `server.servlet.context-path=/VulnerableApp`
is set in `application.properties` and is not overridden by
`application-public.properties`, `application-unsafe.properties`, any
`docker-compose*.yml`, `Dockerfile.base`, or any GitHub workflow, so the
hardcoded value happens to match. It breaks for anyone who overrides the context
path. `application.properties` already anticipates a multi-app deployment behind
VulnerableApp-facade, which looks like the likeliest way to hit this, though I
have not tested a facade deployment.

Raised by CodeRabbit on #733 and filed separately because the code predates that
PR. In #733 the inline block of site 1 is extracted into a helper named
`applicationUrl()` with the same construction, and that PR also adds a `Link:
<...>; rel="successor-version"` header carrying the same base. Sites 2 and 3 are
byte-identical between `master` and that PR's merge base.

On a fix: sites 1 and 2 are straightforward, derive the segment from
`request.getContextPath()` and keep the current output when it is empty. Site 3
is less mechanical, because `CONTEXT_PATH` is `static final` and `normalizeUrl` is
a static method that `DastBenchmarkStrategyTest` calls statically, so making the
value configurable means making both non-static and reworking that test. It may
deserve its own commit. I have not attempted it.

Worth adding a test that serves under a non-default context path and asserts the
scanner URL and the sitemap `<loc>` both stay inside it.

Yes, happy to pick it up.

vianbas:
Fixes #736.

## What was wrong

`VulnerableAppRestController` built the URLs it hands to scanners by appending
`FrameworkConstants.VULNERABLE_APP` to the scheme, host and port, so those URLs
carried `/VulnerableApp/` whatever context path the request actually arrived
under. Two sites on `master` (`766890c`): `getScannerRelatedInformation`
(line 92, constant appended at line 105) and `sitemapForPassiveScanners`
(line 137, constant appended at line 164).

`DastBenchmarkStrategy` held the same literal from the other direction:
`private static final String CONTEXT_PATH = "/VulnerableApp"` at line 49, which
`normalizeUrl` strips to reduce scanner findings and ground-truth entries to a
common key.

`server.servlet.context-path` is set in exactly one place in the repository,
`application.properties:2`. I grepped every spelling of the key,
case-insensitively, across all files tracked at `766890c`: 10 matches, none of
them in `application-public.properties`, `application-unsafe.properties`, the
four `docker-compose*.yml` files, `Dockerfile.base`, or any of the seven
workflows. The hardcoded value therefore matches every configuration the
repository ships, and diverges for anyone who overrides it.

## The fix

One commit per site, and a third that pins the shipped deployment in the
controller test.

**Controller.** Both methods already receive the `HttpServletRequest`, so the
deployment's own context path is one call away and no signature changes. An
empty context path (a root deployment) yields `scheme://host:port/`.

**Benchmark.** The class is already a `@Service` taking `@Value` config, so the
context path arrives the same way, and `normalizeUrl` takes it as a parameter.
A null or empty context path strips nothing.

### A real deployment, before and after

I built the boot jar from `766890c` and from this branch and ran both on the
embedded Tomcat. Counts below are from parsing each response, `url` fields for
`/scanner` and `<loc>` elements for `/sitemap.xml`.

Started with `--server.servlet.context-path=/customCtx`, both jars log:

```
Tomcat started on port(s): 9090 (http) with context path '/customCtx'
```

| Jar | `/scanner` | `/sitemap.xml` |
|---|---|---|
| `766890c` | 153 entries, all 153 on `http://localhost:9090/VulnerableApp` | 164 entries, all 164 on `http://localhost:9090/VulnerableApp` |
| this branch | 153 entries, all 153 on `http://localhost:9090/customCtx` | 164 entries, all 164 on `http://localhost:9090/customCtx` |

Taking one URL that `766890c` emitted and fetching it back from the same
running instance:

| URL | Status |
|---|---|
| `http://localhost:9090/VulnerableApp/JWTVulnerability/LEVEL_1` | 404 |
| `http://localhost:9090/customCtx/JWTVulnerability/LEVEL_1` | 200 |

This branch started with an empty context path,
`--server.servlet.context-path=`, logs:

```
Tomcat started on port(s): 9090 (http) with context path ''
```

`/scanner` returns 153 entries and `/sitemap.xml` 164, all on
`http://localhost:9090` with no context segment, for example
`http://localhost:9090/JWTVulnerability/LEVEL_1`, which answers 200. Neither
response contains a doubled slash or the string `VulnerableApp`.

### The shipped configuration is unchanged

Both jars started with no flag, so both under `/VulnerableApp`. Each response
is byte-identical between the two jars:

| Response | sha256, first 16 hex |
|---|---|
| `/scanner` | `1739486ea7348e10` |
| `/sitemap.xml` | `c5dd89f486c5ff33` |

This branch's jar starting cleanly, under all three context paths, is also what
checks that the new `@Value("${server.servlet.context-path:}")` constructor
parameter wires. The repository has no `@SpringBootTest`, so no test starts the
Spring context.

### Controller tests, against unmodified master

The new test class copied onto a clean `766890c` worktree with no other change.
Three of its four cases fail:

```
scannerUrlIsBuiltFromTheRequestContextPath
org.opentest4j.AssertionFailedError: [the scanner base URL must sit under the served context path]
Expecting: <"https://10.0.0.5:443/VulnerableApp/"> to be equal to: <"https://10.0.0.5:443/customCtx/"> but was not.

sitemapLocEntriesAreBuiltFromTheRequestContextPath
java.lang.AssertionError: [every <loc> must sit under the served context path]
Expecting: <"... <loc> https://10.0.0.5:443/VulnerableApp/SQLInjection/LEVEL_1 </loc> ...">
to contain: <"https://10.0.0.5:443/customCtx/SQLInjection/LEVEL_1">

rootDeploymentGetsUrlsWithNoContextSegment
org.opentest4j.AssertionFailedError:
Expecting: <"https://10.0.0.5:443/VulnerableApp/"> to be equal to: <"https://10.0.0.5:443/"> but was not.
```

The sitemap message is abridged at the ellipses. Otherwise these are the
assertion messages as JUnit reported them, with line breaks added and repeated
spaces collapsed. All four cases pass on this branch.

The fourth, `defaultContextPathStillProducesTheVulnerableAppSegment`, passes on
`766890c` as well, and that is its purpose: it serves the controller under the
shipped `/VulnerableApp` and pins the URLs that deployment already receives, so
the two cases above cannot stay green while the default quietly changes. There
is no test class for this controller on `master`.

### Benchmark normalisation, against unmodified master

`normalizeUrl` on `766890c`, called directly from a throwaway test:

```
/VulnerableApp/SQLInjection/LEVEL_1                       ->  /SQLInjection/LEVEL_1
http://localhost:9090/VulnerableApp/SQLInjection/LEVEL_1  ->  /SQLInjection/LEVEL_1
/SQLInjection/LEVEL_1                                     ->  /SQLInjection/LEVEL_1
/customCtx/SQLInjection/LEVEL_1                           ->  /customCtx/SQLInjection/LEVEL_1
https://10.0.0.5:443/customCtx/SQLInjection/LEVEL_1       ->  /customCtx/SQLInjection/LEVEL_1
```

Under `/VulnerableApp` all three forms of the same endpoint reduce to one key.
Under `/customCtx` the absolute and the context-relative form reduce to
`/customCtx/SQLInjection/LEVEL_1` and `/SQLInjection/LEVEL_1`, two keys that no
longer match.

The constant also disagreed with its own configuration:
`benchmark.dast.ground-truth.url` is built from
`${server.servlet.context-path:/VulnerableApp}` (`application.properties:46`),
so the ground-truth URL follows whatever context path is configured while the
constant did not.

### Test run

`./gradlew clean build --no-daemon`, the command this repository's CI runs, on
JDK 17: run twice, `BUILD SUCCESSFUL` both times, 38 classes, 432 tests,
0 failures, 0 errors, 2 skipped. At `766890c`, `./gradlew clean test` gives
37 classes and 427 tests, so this PR adds one test class and five tests. The
2 skipped are pre-existing, both in `Http3xxStatusCodeBasedInjectionTest`.

## Correction to the issue

In #736 I wrote that site 3 means "making both non-static and reworking that
test". That was wrong, and I am correcting it here rather than editing the
issue. `normalizeUrl` stays `static` and package-private and takes the context
path as a second parameter, so the existing assertions in
`urlNormalization_acceptsAbsoluteRelativeAndContextStrippedForms` keep their
cases and gain one argument each.

## What I did not test

- A real scanner run. Whether the benchmark key divergence changes a reported
  number depends on which URL form the scanner emits, which I have not measured.
- A VulnerableApp-facade deployment.
- A deployment behind a reverse proxy that rewrites the path.

## Overlap with #733

#733 changes both of the main-source files this PR touches. Merging this branch
onto its head (`cad4d1a`) conflicts in one of them,
`VulnerableAppRestController.java`; `DastBenchmarkStrategy.java` merges cleanly,
since #733 changes only a javadoc line there. Whichever lands second needs the
rebase, and I am happy to do it on this side either way. In #733 the block this
PR edits is extracted into a helper, `applicationUrl(request)`, so on that side
the same change moves into the helper, which also feeds the
`Link: <...>; rel="successor-version"` header that PR adds.

`FrameworkConstants.VULNERABLE_APP` stays. `UnrestrictedFileUpload.java:123`
still uses it.

## Summary by CodeRabbit

* **Bug Fixes**
  * Scanner and sitemap URLs now correctly use the application’s configured context path.
  * Applications deployed at the root path or under a custom path now generate valid URLs.
  * Removed assumptions tied to the default `/VulnerableApp` deployment path.

* **Tests**
  * Added coverage for custom context paths, root deployments, and missing context paths.

### DIFF
--- a/src/main/java/org/sasanlabs/benchmark/service/DastBenchmarkStrategy.java
+++ b/src/main/java/org/sasanlabs/benchmark/service/DastBenchmarkStrategy.java
@@ -46,7 +46,6 @@ public class DastBenchmarkStrategy implements BenchmarkStrategy {
 
     private static final Logger LOGGER = LogManager.getLogger(DastBenchmarkStrategy.class);
 
-    private static final String CONTEXT_PATH = "/VulnerableApp";
     private static final String UNSECURE_VARIANT = "UNSECURE";
     private static final String KEY_SEPARATOR = "::";
     private static final String AXIS_TYPE = "TYPE";
@@ -57,11 +56,22 @@ public class DastBenchmarkStrategy implements BenchmarkStrategy {
     private final RestTemplate restTemplate;
     private final String groundTruthUrl;
 
+    /**
+     * The context path this instance is served under, used to reduce scanner findings and
+     * ground-truth entries to a common key. Read from configuration rather than hardcoded: {@code
+     * benchmark.dast.ground-truth.url} already follows {@code server.servlet.context-path}, so a
+     * fixed value here would disagree with the very URLs it is comparing. Empty for a root
+     * deployment, which strips nothing.
+     */
+    private final String contextPath;
+
     public DastBenchmarkStrategy(
             RestTemplate restTemplate,
-            @Value("${benchmark.dast.ground-truth.url}") String groundTruthUrl) {
+            @Value("${benchmark.dast.ground-truth.url}") String groundTruthUrl,
+            @Value("${server.servlet.context-path:}") String contextPath) {
         this.restTemplate = restTemplate;
         this.groundTruthUrl = groundTruthUrl;
+        this.contextPath = contextPath;
     }
 
     @Override
@@ -81,7 +91,7 @@ public class DastBenchmarkStrategy implements BenchmarkStrategy {
             if (finding == null || finding.getUrl() == null) {
                 continue;
             }
-            String url = normalizeUrl(finding.getUrl());
+            String url = normalizeUrl(finding.getUrl(), contextPath);
             List<String> keys = keysForFinding(url, finding);
             if (keys.isEmpty()) {
                 continue;
@@ -161,7 +171,7 @@ public class DastBenchmarkStrategy implements BenchmarkStrategy {
             if (!UNSECURE_VARIANT.equalsIgnoreCase(entry.getVariant())) {
                 continue;
             }
-            String url = normalizeUrl(entry.getUrl());
+            String url = normalizeUrl(entry.getUrl(), contextPath);
             String method =
                     (entry.getRequestMethod() != null)
                             ? entry.getRequestMethod().name()
@@ -283,10 +293,11 @@ public class DastBenchmarkStrategy implements BenchmarkStrategy {
      * Normalises a URL to a context-path-stripped path so absolute ground-truth URLs and relative
      * scanner-finding URLs compare equal.
      *
-     * <p>Strips scheme + host + port, the {@code /VulnerableApp} context path, the query string,
-     * and a trailing slash. Leaves the path case unchanged.
+     * <p>Strips scheme + host + port, the context path the application is served under, the query
+     * string, and a trailing slash. Leaves the path case unchanged. An empty or {@code null} {@code
+     * contextPath} strips no prefix, leaving a root deployment's paths unchanged.
      */
-    static String normalizeUrl(String url) {
+    static String normalizeUrl(String url, String contextPath) {
         if (url == null) {
             return "";
         }
@@ -307,10 +318,12 @@ public class DastBenchmarkStrategy implements BenchmarkStrategy {
             s = s.substring(0, zapSQLPayloadIdx);
         }
 
-        if (s.equals(CONTEXT_PATH)) {
-            s = "/";
-        } else if (s.startsWith(CONTEXT_PATH + "/")) {
-            s = s.substring(CONTEXT_PATH.length());
+        if (contextPath != null && !contextPath.isEmpty()) {
+            if (s.equals(contextPath)) {
+                s = "/";
+            } else if (s.startsWith(contextPath + "/")) {
+                s = s.substring(contextPath.length());
+            }
         }
         if (s.length() > 1 && s.endsWith("/")) {
             s = s.substring(0, s.length() - 1);
--- a/src/main/java/org/sasanlabs/controller/VulnerableAppRestController.java
+++ b/src/main/java/org/sasanlabs/controller/VulnerableAppRestController.java
@@ -94,6 +94,11 @@ public class VulnerableAppRestController {
         String scheme = request.getScheme(); // http or https
         String serverName = request.getServerName(); // actual hostname/IP
         int serverPort = request.getServerPort(); // actual port
+        // The deployment's own context path, not a hardcoded one:
+        // `server.servlet.context-path` is configurable, so a fixed `/VulnerableApp`
+        // describes this deployment only while that default is in force. Empty for a
+        // root deployment.
+        String contextPath = request.getContextPath();
         String appUrl =
                 new StringBuilder()
                         .append(scheme)
@@ -101,8 +106,7 @@ public class VulnerableAppRestController {
                         .append(serverName)
                         .append(FrameworkConstants.COLON)
                         .append(serverPort)
-                        .append(FrameworkConstants.SLASH)
-                        .append(FrameworkConstants.VULNERABLE_APP)
+                        .append(contextPath)
                         .append(FrameworkConstants.SLASH)
                         .toString();
         return getAllSupportedEndPoints.getScannerRelatedEndPointInformation(appUrl);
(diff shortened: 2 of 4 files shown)


preetkaran20 (approved):
LGTM. Thanks for this PR. Really appreciate finding out the issue and fixing it.

PR #738, opened 2026-08-14
feat: add CSRF vulnerability lab (Fixes #713)

### LINKED ISSUE #713: Add CSRF Vulnerability Lab (Authenticated) — Challenge-Mode, Header-Based Token Pattern for React SPA
# Summary
Add a new Cross-Site Request Forgery (CSRF) vulnerability lab to VulnerableApp.
The lab should demonstrate how CSRF affects modern React + REST API applications where authenticated users perform state-changing actions through API calls.
The implementation should follow VulnerableApp's Challenge Mode approach: users should receive an objective and understand/craft the attack themselves instead of using a pre-built "execute exploit" button.

## Use Case Story
A user is logged into VulnerableApp and manages account settings through the React frontend.

The application exposes REST APIs for actions such as:
changing account email address/updating profile information

An attacker attempts to trick the logged-in user into performing an unintended action.

Because browsers automatically attach authentication cookies to requests, the server must verify that state-changing requests originate from the legitimate application.

## Challenge Mode Flow

Each level should provide a challenge objective.

Example:
You are the attacker. The victim is already logged in. Create a request that changes the victim's account email to attacker@example.com without their knowledge.

The user should:
1. Inspect the legitimate API request using browser developer tools.
2. Understand the required request structure.
3. Create the attack payload based on the vulnerability in that level.
4. Execute the request.

The objective is to teach the vulnerability mechanics, not just demonstrate a successful attack.

### React and CSRF Token Handling

The implementation should follow a pattern suitable for a React SPA.

Recommended approach:
CSRF token is delivered during the existing authentication flow.
The token is kept in React runtime memory.
State-changing API requests include the token through a custom header such as X-CSRF-Token.
Backend validates the token before processing sensitive operations.

Avoid introducing unnecessary endpoints only for token retrieval unless required by the implementation.

## Level 1 — No CSRF Protection
Challenge
Change the victim's account email using a crafted request.

The backend accepts state-changing requests only based on the authenticated session cookie.
No CSRF validation is performed.

### Learning Objective
Authentication does not prevent CSRF attacks.

## Level 2 — CSRF Token Present but Not Validated
Challenge
A CSRF token is included in requests. Find why the attack still succeeds.

The application generates and sends a CSRF token, but the backend does not correctly validate it.

Examples:
validation logic missing
validation result ignored
token compared incorrectly

A CSRF token only protects the application when it is validated server-side.

## Level 3 — Weak CSRF Token Generation
Challenge
The application validates the CSRF token, but the token generation mechanism is weak.

The token can be predicted because it is generated using insecure or guessable values.
Examples:
sequential values
timestamps
predictable user information
weak random generation

CSRF tokens must be unpredictable and generated securely.

## Level 4 — Unsafe State-Changing GET Request
Challenge
The normal POST request is protected. Find another way to trigger the action.

Vulnerability

A state-changing operation is incorrectly exposed through HTTP GET.

Example:

GET /change-email?email=attacker@example.com

GET requests should not modify server-side state.

## Level 5 — Secure CSRF Protection
Challenge
Previous attacks no longer work. Understand why.

Protection

Implement a secure CSRF defense:

secure token generation
server-side token validation
correct request header validation
appropriate cookie configuration such as SameSite attributes

Learning Objective
CSRF protection must be enforced by the backend and cannot rely only on frontend behavior.

Optional Advanced Scenario — Mailpit Attack Delivery
As an optional enhancement, integrate Mailpit to demonstrate a realistic attack delivery flow.

Example:

Attacker creates a malicious link.
Victim receives the link through email.
Victim clicks the link.
The CSRF attack is triggered.

This demonstrates how CSRF attacks are commonly delivered while keeping the core vulnerability focused on request validation.

## Technical Requirements

- Follow existing VulnerableApp LEVEL_X structure.
- Ensure one level db state change doesnt impact other levels
- Use React frontend and REST APIs.
- Add a realistic state-changing operation for demonstration.
- Keep vulnerable and secure implementations isolated per level.
- Provide clear verification steps for each level.

mmustafasenoglu:
Fixes #713

## What

Adds a full CSRF (Cross-Site Request Forgery, CWE-352) lab with five progressive levels, following the same structure as the existing labs (IDOR, XSS, etc.):

- **Level 1** – No CSRF protection at all: the email change endpoint only checks the session cookie.
- **Level 2** – The server issues a CSRF token after login, but the change endpoint reads the `X-CSRF-Token` header and never validates it.
- **Level 3** – The token is validated, but it is generated as `username:epochSeconds`, so an attacker can forge a valid token within the 2 minute acceptance window.
- **Level 4** – The POST endpoint is properly protected, but the same state-changing operation is also exposed over GET, which works as a real browser attack (top-level navigation carries the Lax cookie).
- **Level 5** – Secure implementation: unpredictable server-side token validated on every state change, session cookie marked `SameSite=Lax`, no state change over GET.

## Implementation details

- Each level has its own isolated user table (`csrf_accounts_l1..l5`) and session cookie (`csrf_session_l1..l5`) so that solving one level does not break the state of the others. Two users per table: `victim`/`Victim@123` and `attacker`/`Attacker@123`.
- Passwords are stored as salted SHA-256 hashes, reusing the existing `PasswordHashingUtils`.
- Every level ships a challenge card with three hints plus a payload description, the attack vector is annotated with `@AttackVector` and the benchmark CSV is updated with the CWE-352 expected issues.
- UI follows the existing vanilla JS template pattern (`static/templates/CSRFVulnerability/LEVEL_X/CSRF.{html,js,css}`). The issue mentions a React SPA; this repo currently has no React setup and all labs use the plain template approach, so I kept it consistent with the codebase. Happy to adjust if a React migration is preferred.

## How to test

```bash
./gradlew bootRun
```

1. Open `http://localhost:9090/VulnerableApp/CSRFVulnerability`
2. Login as `victim` / `Victim@123` on each level
3. Levels 1–3 can be exploited with a curl request that carries the victim's session cookie (hints show the exact commands)
4. Level 4 works by opening the GET link in a new tab while logged in
5. Level 5 rejects requests without a valid token and the cookie is `SameSite=Lax`

## Summary by CodeRabbit

* **New Features**
  * Added a five-level Cross-Site Request Forgery (CSRF) training lab.
  * Covers missing protection, invalid and predictable tokens, unsafe state-changing requests, and secure token validation.
  * Added login, account, email-update, token, and attack-demonstration interactions for each level.
  * Added seeded accounts and responsive pages with status feedback, hints, and example requests.
  * Secure demonstrations include protected cookies and server-side token validation.

* **Documentation**
  * Added challenge descriptions, attack-vector explanations, payload guidance, and progressive-level instructions.

### DIFF
--- a/scanner/sast/expectedIssues.csv
+++ b/scanner/sast/expectedIssues.csv
@@ -153,3 +153,7 @@ CWE-79,Reflected XSS,src/main/java/org/sasanlabs/service/vulnerability/xss/refle
 CWE-79,Reflected XSS,src/main/java/org/sasanlabs/service/vulnerability/xss/reflected/XSSWithHtmlTagInjection.java,73,1
 CWE-611,Xxe,src/main/java/org/sasanlabs/service/vulnerability/xxe/XXEVulnerability.java,64,1
 CWE-611,Xxe,src/main/java/org/sasanlabs/service/vulnerability/xxe/XXEVulnerability.java,137,1
+CWE-352,Cross-Site Request Forgery,src/main/java/org/sasanlabs/service/vulnerability/csrf/CSRFVulnerability.java,85,1
+CWE-352,Cross-Site Request Forgery,src/main/java/org/sasanlabs/service/vulnerability/csrf/CSRFVulnerability.java,137,1
+CWE-352,Cross-Site Request Forgery,src/main/java/org/sasanlabs/service/vulnerability/csrf/CSRFVulnerability.java,194,1
+CWE-352,Cross-Site Request Forgery,src/main/java/org/sasanlabs/service/vulnerability/csrf/CSRFVulnerability.java,266,1
--- a/src/main/java/org/sasanlabs/configuration/VulnerableAppConfiguration.java
+++ b/src/main/java/org/sasanlabs/configuration/VulnerableAppConfiguration.java
@@ -145,6 +145,8 @@ public class VulnerableAppConfiguration {
         populator.addScript(new ClassPathResource("scripts/CryptographicFailures/db/schema.sql"));
         populator.addScript(new ClassPathResource("scripts/SessionManagement/db/schema.sql"));
         populator.addScript(new ClassPathResource("scripts/SessionManagement/db/data.sql"));
+        populator.addScript(new ClassPathResource("scripts/CSRF/db/schema.sql"));
+        populator.addScript(new ClassPathResource("scripts/CSRF/db/data.sql"));
         populator.setSeparator(";");
 
         DataSourceInitializer initializer = new DataSourceInitializer();
--- a/src/main/java/org/sasanlabs/service/vulnerability/csrf/CSRFAccount.java
+++ b/src/main/java/org/sasanlabs/service/vulnerability/csrf/CSRFAccount.java
@@ -1,0 +1,39 @@
+package org.sasanlabs.service.vulnerability.csrf;
+
+/** Represents a CSRF lab account. */
+public class CSRFAccount {
+
+    private int id;
+    private String username;
+    private String email;
+
+    public CSRFAccount(int id, String username, String email) {
+        this.id = id;
+        this.username = username;
+        this.email = email;
+    }
+
+    public int getId() {
+        return id;
+    }
+
+    public void setId(int id) {
+        this.id = id;
+    }
+
+    public String getUsername() {
+        return username;
+    }
+
+    public void setUsername(String username) {
+        this.username = username;
+    }
+
+    public String getEmail() {
+        return email;
+    }
+
+    public void setEmail(String email) {
+        this.email = email;
+    }
+}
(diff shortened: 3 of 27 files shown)


mmustafasenoglu:
Addressed all three findings in the latest pushes: getSession now shares a deterministic isExpired() helper with the evictor (tested with an expired session read, no wall-clock waits), the Level 5 cookie tests capture the header and assert SameSite=Lax / Secure, and I also fixed the Level 3 tests for the 4-arg signature which would have failed compilation. Couldn't run Gradle locally (the Jib block fails evaluation in this environment, unrelated to the change), so I compiled the touched sources plus all CSRF tests with javac and ran them via the JUnit launcher: 60 passed. Spotless formatting was matched by hand to google-java-format AOSP (import order, 100-col); CI will confirm.

mmustafasenoglu:
Both addressed: the Level 3 label now escapes the token-format brackets so it renders as text, and the Level 4 payload is now a standalone attacker-hosted page (absolute URL, matching the convention of the other payloads) with the demo copy updated to match. Vulnerable endpoint and demo behavior unchanged.

mmustafasenoglu:
Good catch — Level 3 really does need it since the whole attack rides a cross-site form POST. The session cookie there now carries an explicit SameSite=None with Secure (localhost stays working as a trustworthy origin), reusing the same ResponseCookie pattern as Level 5. Levels 1-2 keep the plain cookie since those demos run through curl. Added a test asserting both attributes, all 61 CSRF tests pass locally.

PR #758, opened 2026-08-15
Add CSRF vulnerability lab with 5 challenge levels

### LINKED ISSUE #713: Add CSRF Vulnerability Lab (Authenticated) — Challenge-Mode, Header-Based Token Pattern for React SPA

mmustafasenoglu:
Fixes #713

Adds a CSRF vulnerability lab following the Challenge Mode approach. Each level teaches a different aspect of CSRF defense through hands-on objectives.

Changes:
- `VulnerabilityType.java`: Added `CROSS_SITE_REQUEST_FORGERY` enum (CWE-352)
- `CrossSiteRequestForgeryVulnerability.java`: New controller with 5 levels
- `CrossSiteRequestForgeryVulnerabilityPayload.properties`: Attack vector payloads
- `messages_en_US.properties`: i18n labels, hints, and challenge descriptions
- `templates/CrossSiteRequestForgeryVulnerability/LEVEL_1/`: HTML/JS/CSS UI

Levels:
- Level 1: No CSRF protection — only session cookie required
- Level 2: Token present but never validated server-side
- Level 3: Weak token generation (predictable timestamp)
- Level 4: State-changing operation exposed via GET
- Level 5: Secure CSRF defense (random token + validation + SameSite)

## Summary by CodeRabbit

* **New Features**
  * Added a CSRF vulnerability learning lab with five progressive levels, from missing protection to secure token validation.
  * Added login, account lookup, email-change exercises, CSRF token handling, session management, and SameSite cookie demonstrations.
  * Added dedicated interfaces, guidance, hints, attack scenarios, and example requests for each level.
  * Added sample accounts and supporting vulnerability classification metadata.

* **Tests**
  * Added comprehensive coverage for accounts, sessions, authentication, token validation, and all five lab levels.

### DIFF
--- a/src/main/java/org/sasanlabs/service/vulnerability/csrf/CSRFVulnerability.java
+++ b/src/main/java/org/sasanlabs/service/vulnerability/csrf/CSRFVulnerability.java
@@ -1,5 +1,6 @@
 package org.sasanlabs.service.vulnerability.csrf;
 
+import java.util.function.BiFunction;
 import javax.servlet.http.Cookie;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
@@ -12,8 +13,10 @@ import org.sasanlabs.internal.utility.annotations.VulnerableAppRestController;
 import org.sasanlabs.service.vulnerability.bean.GenericVulnerabilityResponseBean;
 import org.sasanlabs.vulnerability.types.VulnerabilityType;
 import org.springframework.context.annotation.Profile;
+import org.springframework.http.HttpHeaders;
 import org.springframework.http.HttpStatus;
 import org.springframework.http.ResponseEntity;
+import org.springframework.http.ResponseCookie;
 import org.springframework.web.bind.annotation.CookieValue;
 import org.springframework.web.bind.annotation.RequestHeader;
 import org.springframework.web.bind.annotation.RequestMethod;
@@ -82,7 +85,7 @@ public class CSRFVulnerability {
     public ResponseEntity<GenericVulnerabilityResponseBean<Object>> level1ChangeEmail(
             @CookieValue(value = "csrf_session_l1", required = false) String sessionToken,
             @RequestParam String email) {
-        return changeEmail(1, sessionToken, email, null, true);
+        return changeEmail(1, sessionToken, email, null, null);
     }
 
     @VulnerableAppRequestMapping(
@@ -134,7 +137,7 @@ public class CSRFVulnerability {
             @RequestParam String email) {
         // The token is read from the request but its value is never compared
         // against anything - the state change succeeds with any (or no) token.
-        return changeEmail(2, sessionToken, email, csrfToken, true);
+        return changeEmail(2, sessionToken, email, csrfToken, null);
     }
 
     @VulnerableAppRequestMapping(
@@ -184,23 +187,12 @@ public class CSRFVulnerability {
             @CookieValue(value = "csrf_session_l3", required = false) String sessionToken,
             @RequestHeader(value = CSRF_TOKEN_HEADER, required = false) String csrfToken,
             @RequestParam String email) {
-        CSRFSessionService.CSRFLoginSession session =
-                csrfSessionService.getSession(3, sessionToken);
-        if (session == null) {
-            return response(NOT_LOGGED_IN, false, HttpStatus.UNAUTHORIZED);
-        }
-        // The token format is username:epochSeconds, so an attacker who knows
-        // the victim's username can forge a valid token for the current window.
-        if (!csrfSessionService.isValidWeakToken(session, csrfToken)) {
-            return response(INVALID_TOKEN, false, HttpStatus.FORBIDDEN);
-        }
-        CSRFAccount account = csrfSessionService.accountByUsername(3, session.getUsername());
-        if (account == null) {
-            return response(NOT_LOGGED_IN, false, HttpStatus.UNAUTHORIZED);
-        }
-        csrfSessionService.updateEmail(3, account.getId(), email);
-        csrfSessionService.refreshSessionEmail(3, sessionToken);
-        return response(EMAIL_UPDATED, true);
+        return changeEmail(
+                3,
+                sessionToken,
+                email,
+                csrfToken,
+                (session, token) -> csrfSessionService.isValidWeakToken(session, token));
     }
 
     @VulnerableAppRequestMapping(
@@ -250,7 +242,12 @@ public class CSRFVulnerability {
             @CookieValue(value = "csrf_session_l4", required = false) String sessionToken,
             @RequestHeader(value = CSRF_TOKEN_HEADER, required = false) String csrfToken,
             @RequestParam String email) {
-        return changeEmail(4, sessionToken, email, csrfToken, false);
+        return changeEmail(
+                4,
+                sessionToken,
+                email,
+                csrfToken,
+                csrfSessionService::isValidSecureToken);
     }
 
     // The state-changing operation is also exposed over GET without any CSRF
@@ -263,7 +260,7 @@ public class CSRFVulnerability {
     public ResponseEntity<GenericVulnerabilityResponseBean<Object>> level4ChangeEmailViaGet(
             @CookieValue(value = "csrf_session_l4", required = false) String sessionToken,
             @RequestParam String email) {
-        return changeEmail(4, sessionToken, email, null, true);
+        return changeEmail(4, sessionToken, email, null, null);
     }
 
     @VulnerableAppRequestMapping(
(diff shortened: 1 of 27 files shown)


src/main/java/org/sasanlabs/service/vulnerability/csrf/CrossSiteRequestForgeryVulnerability.java
```diff
@@ -0,0 +1,276 @@
...
+
+        String email = allParams.getOrDefault("email", VICTIM_EMAIL);
+        addSecureTokenCookie(response);
+        return ResponseEntity.ok(
+                new GenericVulnerabilityResponseBean<>(
+                        "Email changed to: " + email + " (secure CSRF protection enforced)", true));
+    }
+
+    /** Level 2 initial token endpoint (called by frontend on page load). */
+    @VulnerableAppRequestMapping(
```
preetkaran20: I think we need to have it as `@RequestMapping` annotation directly instead of @VulnerableAppRequestMapping.

src/main/resources/static/templates/CrossSiteRequestForgeryVulnerability/LEVEL_1/CrossSiteRequestForgeryVulnerability.js
```diff
@@ -0,0 +1,158 @@
...
+  var text = attackPayload.textContent;
+  navigator.clipboard.writeText(text).then(function () {
+    document.getElementById("copyPayload").textContent = "Copied!";
+    setTimeout(function () {
+      document.getElementById("copyPayload").textContent = "Copy Payload";
+    }, 2000);
+  });
+});
+
+function getCookie(name) {
```
preetkaran20: we cannot access http only cookie from UI. can you please fix this?

src/main/resources/static/templates/CrossSiteRequestForgeryVulnerability/LEVEL_1/CrossSiteRequestForgeryVulnerability.js
```diff
@@ -0,0 +1,158 @@
...
+  .addEventListener("click", function () {
+    var email = document.getElementById("emailInput").value;
+    if (!email) {
+      email = "victim@example.com";
+    }
+
+    var apiUrl = getUrlForVulnerabilityLevel();
+    var options;
+
+    if (level === "LEVEL_4") {
```
preetkaran20: These If else checks are not maintainable. I would suggest breaking js files for each level if that makes it better and maintainable.

src/main/resources/static/templates/CrossSiteRequestForgeryVulnerability/LEVEL_1/CrossSiteRequestForgeryVulnerability.js
```diff
@@ -0,0 +1,158 @@
+var levelDescriptions = {
```
preetkaran20: Page load is not doing api call to fetch the XSRF tokens for each level. can you please relook at it?

src/main/java/org/sasanlabs/service/vulnerability/csrf/CrossSiteRequestForgeryVulnerability.java
```diff
@@ -0,0 +1,276 @@
...
+            return ResponseEntity.ok(
+                    new GenericVulnerabilityResponseBean<>(
+                            "CSRF validation failed. The request was rejected.", false));
+        }
+
+        String email = allParams.getOrDefault("email", VICTIM_EMAIL);
+        addSecureTokenCookie(response);
+        return ResponseEntity.ok(
+                new GenericVulnerabilityResponseBean<>(
+                        "Email changed to: " + email + " (secure CSRF protection enforced)", true));
```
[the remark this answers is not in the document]
preetkaran20: @mmustafasenoglu can you please relook at this? I tried running it and it is not working. I think as we have react SPA, we should call backend and load the csrf token into the state and then pass it while doing the Api call and that is the right mechanism. 

Ensure, this works in both legacy and modern frontend.

preetkaran20 (commented):
I have looked at the PR and I have concerns regarding showing the email id of the user and then giving update option. I feel directly showing update is not a good UI design and how does user find out that it is actually updated. Please rethink. we have database as well and you can utilize it for storing and manipulating data.

src/main/java/org/sasanlabs/service/vulnerability/csrf/CrossSiteRequestForgeryVulnerability.java
```diff
@@ -0,0 +1,276 @@
...
+            @CookieValue(value = "victim_session", defaultValue = "victim") String session,
+            HttpServletRequest request,
+            HttpServletResponse response) {
+        // Token is received but never actually compared or checked
+        String csrfToken = request.getHeader("X-CSRF-Token");
+        // Intentionally missing: if (csrfToken == null || !csrfToken.equals(storedToken)) { reject
+        // }
+
+        String email = allParams.getOrDefault("email", VICTIM_EMAIL);
+        addVictimCookie(response);
```
preetkaran20: this doesn't seems right. we should update database as well.

src/main/java/org/sasanlabs/service/vulnerability/csrf/CSRFVulnerability.java
```diff
@@ -0,0 +1,448 @@
...
+    public ResponseEntity<GenericVulnerabilityResponseBean<Object>> level1Login(
+            @RequestParam String username,
+            @RequestParam String password,
+            HttpServletRequest request,
+            HttpServletResponse response) {
+        return login(1, username, password, "csrf_session_l1", request, response);
+    }
+
+    @VulnerableAppRequestMapping(
+            value = "LEVEL_1/changeEmail",
```
preetkaran20, line 82: we cannot use VulnerableAppRequestMapping annotation with custom values. You need to use Request mapping which is spring boot annotation for this

preetkaran20: can you please validate the project before marking it ready for review?

PR #720, opened 2026-08-05
feat: add Challenge mode to XSS (#562)

pereiravp:
## Summary
Adds `@ChallengeCard` annotations to the vulnerable and secure levels of `XSSWithHtmlTagInjection`, `XSSInImgTagAttribute` and `PersistentXSSInHTMLTagVulnerability`, following the pattern established in #709 for LDAP Injection, File Upload and SSRF.

- Each level gets a `challengeText`, progressive `hints`, and a documented `payload` (description + value), added to `messages.properties`.
- Purely additive. No changes to existing `@AttackVector` / `@VulnerableAppRequestMapping` annotations or vulnerability logic.
- Covers all 19 levels across the 3 XSS classes (reflected div injection, reflected img attribute, and persistent).

## Test plan
- [x] `./gradlew compileJava` succeeds.
- [x] `./gradlew spotlessJavaApply` reports the changed files as already conforming to the project's Google Java Format style.
- [x] Ran the app locally (`bootRun`), opened Challenge mode, and confirmed the challenge/hint/payload text renders correctly for each of the three modules.
- [x] Verified the documented payloads against the live endpoints (e.g. level 1 of `XSSWithHtmlTagInjection` reflects `<script>alert(1)</script>` unescaped; the null-byte bypass on `PersistentXSSInHTMLTagVulnerability` level 4 lets a raw `` tag through after `%00`).

Closes #562

## Summary by CodeRabbit

* **New Features**
  * Added richer challenge details across multiple XSS scenarios, including level-specific instructions, hints, and payload examples.
  * Expanded guidance for HTML tag injection, image attribute injection, and persistent XSS challenges.
  * Added localized challenge text covering payloads, filtering behavior, escaping, bypass techniques, and secure examples.
  * Improved progression guidance to help learners understand the techniques demonstrated at each level.

PR #759, opened 2026-08-15
feat: add Challenge mode to Path Traversal

### LINKED ISSUE #560: Add Challenge mode to Path Traversal Vulnerability
updating [Path Traversal](https://github.com/SasanLabs/VulnerableApp/tree/master/src/main/java/org/sasanlabs/service/vulnerability/pathTraversal)

## What needs to be done

## Annotation details

It has three fields:

pereiravp:
## Summary

Closes #560. Adds `@ChallengeCard` metadata to all 13 levels of `PathTraversalVulnerability`, following the merged XSS challenge mode work (#720).

## Approach

Every payload verified end-to-end against a running instance, confirmed to return `isValid: true`.

- **Levels 1-5**: progressively tighten a raw-URL blocklist (`../`, then `%2f`, then case-insensitive `%2f`). Encoding the dots (`%2e%2e`) keeps bypassing all of them, since none inspect encoded dots.
- **Level 6**: first level to check the decoded `fileName` for `..`, but never validates *which* file you ask for. `secret.json` sits unprotected next to the intended files.
- **Levels 7-11**: raw-URL blocklist combined with a NUL-byte parsing quirk. The allowlist check runs on the full parameter, but only the part before `%00` is opened.
- **Level 12**: closes the encoding bypass but collapses back to the same `secret.json` trick as level 6, combined with the NUL-byte bypass.
- **Level 13** (secure baseline): exact-equality filename check, nothing left to exploit.

No frontend changes needed.

## Test plan

- [x] `./gradlew compileJava`, `./gradlew test`, `./gradlew spotlessApply` all clean
- [x] All 13 payloads verified against a running instance
- [x] `@ChallengeCard` data confirmed via `/VulnerableApp/allEndPointJson` for every level

## Summary by CodeRabbit

* **New Features**
  * Added challenge metadata to all 13 path-traversal scenarios, including instructions, hints, and payload descriptions.
  * Added multiple challenge cards for selected scenarios, with expanded hints for Levels 6 and 12.

* **Documentation**
  * Updated guidance for encoded-dot traversal, same-directory file access, and credential disclosure.
  * Refined challenge wording and payload examples for Levels 4, 10, and 11.


src/main/resources/i18n/messages.properties
```diff
@@ -1064,3 +1064,82 @@ PERSISTENT_XSS_LEVEL7_HINT2=Even the null-byte-prefix trick that worked on level
...
+PATH_TRAVERSAL_LEVEL3_PAYLOAD_DESCRIPTION=Encoding every dot hides ".." from the raw-URL check entirely; the decoded fileName parameter still resolves to a real ../../ traversal.
+PATH_TRAVERSAL_LEVEL3_PAYLOAD_VALUE=%2e%2e/%2e%2e/application.properties
+
+PATH_TRAVERSAL_LEVEL4_CHALLENGE=This level also blocks the literal "%2f" in the URL. Bypass it too.
+PATH_TRAVERSAL_LEVEL4_HINT1=The new check only rejects the raw substring "%2f" - it is still only inspecting undecoded text, just like the ".." check next to it.
+PATH_TRAVERSAL_LEVEL4_HINT2=The dot-encoding payload from level 3 never uses "%2f" at all, so blocking that string changes nothing for it.
+PATH_TRAVERSAL_LEVEL4_PAYLOAD_DESCRIPTION=Neither the "%2f" check nor the ".." check inspects encoded dots, so the level 3 payload keeps working unchanged.
+PATH_TRAVERSAL_LEVEL4_PAYLOAD_VALUE=%2e%2e/%2e%2e/application.properties
+
+PATH_TRAVERSAL_LEVEL5_CHALLENGE=The "%2f" check is now case-insensitive too, catching %2F as well. Still bypass it.
```
preetkaran20: @pereiravp same feedback as previous sql injection. Please don't link challenges of one level with another.

src/main/resources/i18n/messages.properties
```diff
@@ -1064,3 +1064,82 @@ PERSISTENT_XSS_LEVEL7_HINT2=Even the null-byte-prefix trick that worked on level
 PERSISTENT_XSS_LEVEL7_PAYLOAD_DESCRIPTION=The same null-byte-prefixed script payload that broke level 6 is now fully escaped and rendered as inert text.
 PERSISTENT_XSS_LEVEL7_PAYLOAD_VALUE=%00<script>alert(1)</script>
 BLIND_SQL_INJECTION_VULNERABILITY_LEVEL1_CHALLENGE2_PAYLOAD_VALUE=1 AND (SELECT SUBSTRING(password,1,1) FROM auth_users WHERE username='admin_sqli')='n'
+
+# Path Traversal Challenge Mode
+PATH_TRAVERSAL_LEVEL1_CHALLENGE=Read a file that lives outside the folder this level was meant to serve files from.
```
preetkaran20: This is good, can we make challenge more closed or may be add multiple challenges. one is like reading application.properties then other is reading say etc/passwd file etc.

src/main/resources/i18n/messages.properties
```diff
@@ -1064,3 +1064,82 @@ PERSISTENT_XSS_LEVEL7_HINT2=Even the null-byte-prefix trick that worked on level
...
+PATH_TRAVERSAL_LEVEL10_PAYLOAD_VALUE=%2e%2e/%2e%2e/application.properties%00UserInfo.json
+
+PATH_TRAVERSAL_LEVEL11_CHALLENGE="%2f" is now blocked case-insensitively too, catching %2F as well. Still bypass it.
+PATH_TRAVERSAL_LEVEL11_HINT1=Exactly like level 5, lower-casing the URL before the "%2f" check only closes the uppercase %2F loophole - it still says nothing about encoded dots.
+PATH_TRAVERSAL_LEVEL11_HINT2=The %2e%2e traversal payload never contains "%2f" in any case, so this level is no stronger against it than level 9 was.
+PATH_TRAVERSAL_LEVEL11_PAYLOAD_DESCRIPTION=Case-insensitive %2f matching only closes the slash-encoding loophole; the dot-encoding plus NUL-byte allowlist bypass from level 9 is completely unaffected.
+PATH_TRAVERSAL_LEVEL11_PAYLOAD_VALUE=%2e%2e/%2e%2e/application.properties%00UserInfo.json
+
+PATH_TRAVERSAL_LEVEL12_CHALLENGE=This level checks the fully decoded fileName parameter for ".." directly, instead of the raw undecoded URL levels 9-11 checked, so every dot or slash encoding trick from earlier levels is dead here. The allowlist is still just a substring check, though. Read a file that isn't part of it.
```
preetkaran20: This seems like hint too. can we make it a goal like we did in sql injection lab?

src/main/resources/i18n/messages.properties
preetkaran20: also can you please make sure that these payloads/challenges work for windows machine as well.

src/main/resources/i18n/messages.properties
pereiravp: Fixed across all 13 levels now, not just this one. Every challenge is self contained.

src/main/resources/i18n/messages.properties
pereiravp: Added a second challenge card on level 1, reads schema.sql from the SQL Injection module instead. Tried etc/passwd too, turns out this vulnerability cannot reach it. Explained in the reply below.

src/main/resources/i18n/messages.properties
pereiravp: Reworded as a goal, read secret.json and get it back. Mechanism is in the hints now.

src/main/resources/i18n/messages.properties
pereiravp: Confirmed it already does, tested on this Windows machine the whole time. Traversal here goes through Class.getResourceAsStream, not a raw filesystem path, so it stays inside the app's own classpath root regardless of OS. Tried reaching etc/passwd and win.ini equivalents through it, both fail, tried a dozen or so ../ depths against README.md and build.gradle too, same result. It can only reach files bundled inside the app itself, which is why the second challenge targets schema.sql instead.

src/main/resources/i18n/messages.properties
```diff
@@ -1065,6 +1065,90 @@ PERSISTENT_XSS_LEVEL7_PAYLOAD_DESCRIPTION=The same null-byte-prefixed script pay
...
+PATH_TRAVERSAL_LEVEL3_HINT2=Encode the dots too: %2e is the URL-encoding for ".". The raw URL then contains no literal ".." for the filter to see, but the decoded fileName parameter is a normal ../ traversal again.
+PATH_TRAVERSAL_LEVEL3_PAYLOAD_DESCRIPTION=Encoding every dot hides ".." from the raw-URL check entirely; the decoded fileName parameter still resolves to a real ../../ traversal.
+PATH_TRAVERSAL_LEVEL3_PAYLOAD_VALUE=%2e%2e/%2e%2e/application.properties
+
+PATH_TRAVERSAL_LEVEL4_CHALLENGE=Read the application's own application.properties file and get its exact contents, including its SMTP credentials, back in the response.
+PATH_TRAVERSAL_LEVEL4_HINT1=The check only rejects the raw substring "%2f" - it is still only inspecting undecoded text, just like the ".." check next to it.
+PATH_TRAVERSAL_LEVEL4_HINT2=Encoding the dots instead of the slash never touches "%2f" at all, so blocking that string changes nothing for it.
+PATH_TRAVERSAL_LEVEL4_PAYLOAD_DESCRIPTION=Neither the "%2f" check nor the ".." check inspects encoded dots, so a dot-encoded payload keeps working unchanged.
+PATH_TRAVERSAL_LEVEL4_PAYLOAD_VALUE=%2e%2e/%2e%2e/application.properties
```
preetkaran20: can you please use a payload that works for level 3 but doesn't work for level 4 and also level 5 so that different attack payloads are learnt by the students

src/main/resources/i18n/messages.properties
```diff
@@ -1065,6 +1065,90 @@ PERSISTENT_XSS_LEVEL7_PAYLOAD_DESCRIPTION=The same null-byte-prefixed script pay
...
+PATH_TRAVERSAL_LEVEL5_PAYLOAD_VALUE=%2e%2e/%2e%2e/application.properties
+
+PATH_TRAVERSAL_LEVEL6_CHALLENGE=Read a file named secret.json that lives in the same folder as the two files this level is meant to serve, and get its exact contents back in the response.
+PATH_TRAVERSAL_LEVEL6_HINT1=The check now runs on the already-decoded fileName string, so %2e%2e-style encoding tricks get decoded back into ".." and are caught immediately.
+PATH_TRAVERSAL_LEVEL6_HINT2=Blocking ".." only stops you leaving the folder - it never restricts which filename inside that same folder you're allowed to ask for. Is scripts/PathTraversal really only holding the two files the level shows you?
+PATH_TRAVERSAL_LEVEL6_PAYLOAD_DESCRIPTION=Blocking ".." prevents leaving the directory, but the level never validates fileName against an allowlist, so a sensitive file sitting right next to the intended ones is still just a filename away - no traversal needed.
+PATH_TRAVERSAL_LEVEL6_PAYLOAD_VALUE=secret.json
+
+PATH_TRAVERSAL_LEVEL7_CHALLENGE=Read the application's own application.properties file and get its exact contents, including its SMTP credentials, back in the response.
```
preetkaran20: you can ask for database credentials as well

src/main/resources/i18n/messages.properties
```diff
@@ -1065,6 +1065,90 @@ PERSISTENT_XSS_LEVEL7_PAYLOAD_DESCRIPTION=The same null-byte-prefixed script pay
...
+PATH_TRAVERSAL_LEVEL5_PAYLOAD_DESCRIPTION=Case-insensitive %2f matching only closes the slash-encoding loophole; a dot-encoding bypass is completely unaffected.
+PATH_TRAVERSAL_LEVEL5_PAYLOAD_VALUE=%2e%2e/%2e%2e/application.properties
+
+PATH_TRAVERSAL_LEVEL6_CHALLENGE=Read a file named secret.json that lives in the same folder as the two files this level is meant to serve, and get its exact contents back in the response.
+PATH_TRAVERSAL_LEVEL6_HINT1=The check now runs on the already-decoded fileName string, so %2e%2e-style encoding tricks get decoded back into ".." and are caught immediately.
+PATH_TRAVERSAL_LEVEL6_HINT2=Blocking ".." only stops you leaving the folder - it never restricts which filename inside that same folder you're allowed to ask for. Is scripts/PathTraversal really only holding the two files the level shows you?
+PATH_TRAVERSAL_LEVEL6_PAYLOAD_DESCRIPTION=Blocking ".." prevents leaving the directory, but the level never validates fileName against an allowlist, so a sensitive file sitting right next to the intended ones is still just a filename away - no traversal needed.
+PATH_TRAVERSAL_LEVEL6_PAYLOAD_VALUE=secret.json
```
preetkaran20, line 1109: can we make challenge of level 6 a bit different? it is simple with secret.json.

src/main/resources/i18n/messages.properties
pereiravp: Level 4 now uses a different payload (dot plus uppercase-slash encoding) that exploits the case-sensitive %2f check. Fails on level 5 once it lower-cases the URL, verified both ways.

src/main/resources/i18n/messages.properties
pereiravp: Added a second challenge card on level 7 targeting application-unsafe.properties, which has the actual DB credentials.

src/main/resources/i18n/messages.properties
pereiravp: Reworded as a goal, filename moved to a hint. Same as level 12 last round.

src/main/resources/i18n/messages.properties
```diff
@@ -1065,6 +1065,97 @@ PERSISTENT_XSS_LEVEL7_PAYLOAD_DESCRIPTION=The same null-byte-prefixed script pay
...
+PATH_TRAVERSAL_LEVEL3_PAYLOAD_VALUE=%2e%2e/%2e%2e/application.properties
+
+PATH_TRAVERSAL_LEVEL4_CHALLENGE=Read the application's own application.properties file and get its exact contents, including its SMTP credentials, back in the response - without relying on the exact same payload level 3 used.
+PATH_TRAVERSAL_LEVEL4_HINT1=The check rejects the raw substring "%2f", but the comparison is case sensitive and never lower-cases the URL first - it only ever looks for that one exact lowercase spelling.
+PATH_TRAVERSAL_LEVEL4_HINT2=Encode the slashes as well as the dots, but in uppercase: "%2F" is not the string "%2f", so it slides past this check even though it decodes to the exact same "/" character.
+PATH_TRAVERSAL_LEVEL4_PAYLOAD_DESCRIPTION=Both the dots and the slashes are percent-encoded here, with the slashes deliberately uppercase ("%2F") so the case-sensitive "%2f" check never matches, even though the URL still decodes to a real ../../ traversal. This exact payload stops working the moment the next level lower-cases the URL before checking.
+PATH_TRAVERSAL_LEVEL4_PAYLOAD_VALUE=%2e%2e%2F%2e%2e%2Fapplication.properties
```
preetkaran20: can we make them all 2f?

src/main/resources/i18n/messages.properties
```diff
@@ -1065,6 +1065,97 @@ PERSISTENT_XSS_LEVEL7_PAYLOAD_DESCRIPTION=The same null-byte-prefixed script pay
...
+
+PATH_TRAVERSAL_LEVEL5_CHALLENGE=Read the application's own application.properties file and get its exact contents, including its SMTP credentials, back in the response - level 4's uppercase-slash trick is dead here.
+PATH_TRAVERSAL_LEVEL5_HINT1=Lower-casing the URL before the "%2f" check closes the uppercase %2F loophole level 4 relied on - any percent-encoded slash, either case, is now caught.
+PATH_TRAVERSAL_LEVEL5_HINT2=The check still says nothing about encoded dots. If the slash is never encoded to begin with, there is no "%2f" of any case for this check to find.
+PATH_TRAVERSAL_LEVEL5_PAYLOAD_DESCRIPTION=Going back to a literal "/" and only percent-encoding the dots sidesteps the %2f check entirely, case-insensitive or not - it was never built to catch dot-encoding in the first place.
+PATH_TRAVERSAL_LEVEL5_PAYLOAD_VALUE=%2e%2e/%2e%2e/application.properties
```
preetkaran20, line 1102: payload is still with 2e across last 3 levels. can we make them distinct like level5 can be broken with %2e and we can break other levels without 2e and use some other payload?

src/main/resources/i18n/messages.properties
pereiravp: Done. Dots are uppercase now too (%2E%2E%2F), slash stays uppercase since that's what makes the bypass work. Tested again.

src/main/resources/i18n/messages.properties
pereiravp: Fixed. Level 10 now uses secret.json, no encoding needed, same gap as level 6. Level 11 keeps level 9's payload but targets application-unsafe.properties for the DB creds instead. Tested all three.

preetkaran20 (approved):
LGTM

preetkaran20:
Thanks a lot for the PR

Issue #495, opened 2026-02-10
JWT vulnerability Level 13 is really not right. Fix it.


preetkaran20:
**Describe the bug**
JWT Vulnerability level is buggy. It is not at all helping in exploiting the issue.

Issues:
1. It has 2 parameters in UI and they are not sure where used. We need to remove them
2. JS Code has hardcoded JWT which is very wrong and we need to remove that.
3. We should fix the backend code to read the cookie and JWK and then use JWK to find out JWT is valid or not and we know JWK is a customer provided value and that header can be manipulated and we should not just rely on that.
<img width="3556" height="1816" alt="Image" src="https://github.com/user-attachments/assets/094e1d55-4eb2-4f4b-ae4d-26b73e0739b6" />

preetkaran20:
sure, please go ahead.

pereiravp:
Hi @preetkaran20, this has had no activity since February (@subhamkumarr got the go ahead back then but nothing landed). Would like to pick it up if it is still open. Already spotted one concrete bug reading the code: line 680 in JWTVulnerability.java calls asymmetricAlgorithmKeyPair.get() right after logging isPresent(), with nothing gating the get() call if it is false.

pereiravp:
Confirmed the original complaints here look already fixed by #499 back in February. Found and fixed a smaller separate bug while checking, PR #766. Might be worth closing this one as resolved.

preetkaran20:
@pereiravp thank for looking into it, I will review the PR. Is it possible for you to pick jwt vulnerability challenge mode ?

pereiravp:
Sure, will pick it up.

PR #766, opened 2026-08-18
fix: remove debug log that can crash JWT levels 8, 13, 14

pereiravp:
## Summary

Related to #495. The original complaints there (hardcoded JWT in the frontend, backend not properly validating via JWK) look already resolved by #499 back in February. While reading through the level 13 code to confirm that, I found a smaller, separate bug still present, this PR fixes just that.

## The bug

Levels 8, 13, and 14 each start with:

```java
Optional<KeyPair> asymmetricAlgorithmKeyPair = jwtAlgorithmKMS.getAsymmetricAlgorithmKey("RS256");
LOGGER.info(asymmetricAlgorithmKeyPair.isPresent() + " " + asymmetricAlgorithmKeyPair.get());
```

`isPresent()` is checked but never gates the `.get()` call on the same line. `JWTAlgorithmKMS.loadAsymmetricAlgorithmKeys()` catches `KeyStoreException`, `NoSuchAlgorithmException`, `CertificateException`, `IOException`, and `UnrecoverableKeyException` and just logs them, so if keystore loading ever fails for any reason, `asymmetricAlgorithmKeyMap` stays empty and this `.get()` throws `NoSuchElementException`, failing the request with an unhandled 500 instead of the level's normal response.

The line has no purpose beyond debugging, so I removed it rather than guarding it. Left the other `.get()` calls in these methods alone, since they're part of the actual vulnerability logic (signing/verifying with the RS256 key), and rewriting how each of those degrades if the key is missing felt like a bigger change than this warrants.

## Test plan

- [x] `./gradlew test` all green
- [x] `./gradlew spotlessApply` no formatting diffs
- [x] Manually verified levels 8, 13, and 14 still fetch a valid token via `GET .../LEVEL_N?fetch=true` after the change

## Summary by CodeRabbit

* **Bug Fixes**
  * Removed unnecessary diagnostic messages from JWT vulnerability scenarios, resulting in cleaner application output.
  * JWT validation, token generation, and response behavior remain unchanged.

### DIFF
--- a/src/main/java/org/sasanlabs/service/vulnerability/jwt/JWTVulnerability.java
+++ b/src/main/java/org/sasanlabs/service/vulnerability/jwt/JWTVulnerability.java
@@ -104,7 +104,6 @@ public class JWTVulnerability {
         Optional<SymmetricAlgorithmKey> symmetricAlgorithmKey =
                 jwtAlgorithmKMS.getSymmetricAlgorithmKey(
                         JWTUtils.JWT_HMAC_SHA_256_ALGORITHM, KeyStrength.HIGH);
-        LOGGER.info(symmetricAlgorithmKey.isPresent() + " " + symmetricAlgorithmKey.get());
         String token = queryParams.get(JWT);
         if (token != null) {
             boolean isValid =
@@ -137,7 +136,6 @@ public class JWTVulnerability {
         Optional<SymmetricAlgorithmKey> symmetricAlgorithmKey =
                 jwtAlgorithmKMS.getSymmetricAlgorithmKey(
                         JWTUtils.JWT_HMAC_SHA_256_ALGORITHM, KeyStrength.HIGH);
-        LOGGER.info(symmetricAlgorithmKey.isPresent() + " " + symmetricAlgorithmKey.get());
         List<String> tokens = requestEntity.getHeaders().get("cookie");
         boolean isFetch = Boolean.valueOf(queryParams.get("fetch"));
         if (!isFetch) {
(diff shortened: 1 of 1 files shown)


src/main/java/org/sasanlabs/service/vulnerability/jwt/JWTVulnerability.java
```diff
@@ -457,8 +457,6 @@ private ResponseEntity<GenericVulnerabilityResponseBean<String>> getJWTResponseB
                     throws UnsupportedEncodingException, ServiceApplicationException {
         Optional<KeyPair> asymmetricAlgorithmKeyPair =
                 jwtAlgorithmKMS.getAsymmetricAlgorithmKey("RS256");
-        LOGGER.info(
```
preetkaran20, line 460: One question but eventually we will be doing get at the later step. I think we can log is present so that we can debug easily in case of failures

pereiravp:
Found 10 more of the same pattern while reading through the rest of this class for the JWT challenge mode work, same LOGGER.info(x.isPresent() + x.get()) issue but for symmetricAlgorithmKey across levels 1-16. Removed those too. All 14 unsecure levels tested fine after.

src/main/java/org/sasanlabs/service/vulnerability/jwt/JWTVulnerability.java
pereiravp: Good point. Kept isPresent logging on all 14, dropped the unsafe get() part. Tested fine.

preetkaran20 (approved):

preetkaran20:
@pereiravp thanks a lot for the PR. LGTM, merging it now.

PR #688, opened 2026-07-06
Add Challenge mode to JWT Vulnerability (#557)

Lalatenduswain:
## Description

Implements Challenge mode for the JWT vulnerability module, closing #557.

Following the pattern already merged for Open Redirect (`Http3xxStatusCodeBasedInjection`) and Clickjacking (#661), this adds `@ChallengeCard` annotations to every vulnerable JWT level and the supporting i18n keys, so each level surfaces a guided challenge with hints and an exploit payload.

## Changes

- **`JWTVulnerability.java`** — added `@ChallengeCard` (challenge text + 3 progressive hints + payload) to all 14 vulnerable levels: LEVEL_1–10 and LEVEL_13–16. (LEVEL_11 is commented out in the source and LEVEL_12 does not exist, so both are skipped.)
- **`i18n/messages.properties`** — added the `JWT_LEVEL_<n>_CHALLENGE`, `_HINT_1..3`, `_PAYLOAD_DESCRIPTION` and `_PAYLOAD_VALUE` keys. Each challenge is written to the specific weakness of that level (URL token leak, missing HttpOnly/Secure, weak-key brute force, null-byte bypass, `alg:none`, RS256/HS256 key confusion, JWK header injection / CVE-2018-0114, empty & unverified signatures, algorithm downgrade, etc.). LEVEL_6 reuses the existing `NONE_ALGORITHM_ATTACK_CURL_PAYLOAD` attack-vector payload.

## Testing

- `./gradlew spotlessApply compileJava` — passes (no formatting changes needed).
- `./gradlew test --tests "*JWTVulnerabilityTest*"` — passes.

Closes #557

## Summary by CodeRabbit

* **New Features**
  * Added detailed JWT challenge cards for multiple levels, including instructions, hints, and example payloads.
  * Expanded in-app guidance for common JWT weakness scenarios such as weak signing secrets, token tampering, algorithm handling, and header-based attacks.
  * Improved localization content so challenge information is displayed directly within the user experience.

PR #767, opened 2026-08-18
feat: add Challenge mode to JWT Vulnerability

### LINKED ISSUE #557: Add Challenge mode to JWT Vulnerability
updating [JWT](https://github.com/SasanLabs/VulnerableApp/tree/master/src/main/java/org/sasanlabs/service/vulnerability/jwt)

## What needs to be done

## Annotation details

It has three fields:

pereiravp:
## Summary

Closes #557. @preetkaran20 asked me to pick this up directly. #688 opened the same issue earlier, flagging that for visibility, happy to defer if you'd rather take that one forward.

Adds `@ChallengeCard` to all 14 UNSECURE levels of `JWTVulnerability`. Level 11 is dead code (not a live endpoint) and there's no SECURE level to skip.

## Approach

Every payload verified end to end against a running instance, including forging real tokens with a small Node script for HMAC signing, RSA keys, and JWK construction.

- **Levels 4, 14**: weak HMAC key ("password"), cracked and forged.
- **Level 5**: null byte in the signature truncates it before the check runs.
- **Levels 6, 7**: `alg: none` skips signature verification entirely (cookie and Authorization header).
- **Level 8**: RS256/HS256 key confusion, the RSA public key's bytes are used as the HMAC secret.
- **Levels 9, 13**: both trust a public key embedded in the token's own `jwk` header, so a self-signed token with your own key pair is accepted. Both call the same validator under the hood despite being framed as different attacks.
- **Level 10**: a token of exactly `..` splits to zero parts and hits an unconditional valid check.
- **Level 15**: only the part count is checked, the signature itself never is.
- **Level 16**: always verifies with HS256 regardless of the header's declared `alg`, so a real signature with a swapped `alg` label still passes.

No frontend changes needed.

## Test plan

- [x] `./gradlew compileJava`, `./gradlew test`, `./gradlew spotlessApply` all clean
- [x] All 14 payloads verified against a running instance, including negative/baseline checks
- [x] Confirmed `@ChallengeCard` is present on all 14 levels via `/VulnerableApp/allEndPointJson`, and absent on level 11

## Summary by CodeRabbit

* **New Features**
  * Added challenge content for JWT vulnerability levels 1–10 and 13–16.
  * Introduced prompts, hints, payload descriptions, and examples for common JWT security scenarios.
  * Added multiple challenge paths where applicable.
  * Expanded guidance on signed versus encrypted tokens, token exposure, cookie security, key retrieval, signature bypasses, and algorithm confusion.
  * Standardized token examples and clarified verification-related challenge guidance.


src/main/resources/i18n/messages.properties
```diff
@@ -1102,3 +1102,88 @@ UNION_BASED_SQL_INJECTION_VULNERABILITY_LEVEL2_HINT1=Your id value is placed bet
 UNION_BASED_SQL_INJECTION_VULNERABILITY_LEVEL2_HINT2=Close the quote, write your UNION SELECT as normal against an id that doesn't exist, then comment out the rest of the original query - including the app's own trailing quote - so the combined statement still parses.
 UNION_BASED_SQL_INJECTION_VULNERABILITY_LEVEL2_PAYLOAD_DESCRIPTION=The leading quote closes the app's string literal, the UNION SELECT filters idor_users down to the ADMIN role, and the trailing -- comments out the app's own closing quote so the combined query stays syntactically valid.
 UNION_BASED_SQL_INJECTION_VULNERABILITY_LEVEL2_PAYLOAD_VALUE=-1' UNION SELECT id, username, salary FROM idor_users WHERE role='ADMIN' --
+
+# JWT Challenge Mode
+JWT_LEVEL1_CHALLENGE=Recover the JWT this level leaks through its own request URL.
```
preetkaran20: I think we can define challenge as something like:

"Is this a secure way of pass JWT token from UI to Backend and if not, please list flaws?" or something like that.

src/main/resources/i18n/messages.properties
```diff
@@ -1102,3 +1102,88 @@ UNION_BASED_SQL_INJECTION_VULNERABILITY_LEVEL2_HINT1=Your id value is placed bet
...
+JWT_LEVEL1_HINT1=The token travels as a URL query parameter, so it lands in browser history, server access logs, and any Referer header sent to third-party resources on the page.
+JWT_LEVEL1_HINT2=A JWT is only Base64URL encoded, not encrypted, so anyone who gets hold of the URL can decode every claim in plain text.
+JWT_LEVEL1_PAYLOAD_DESCRIPTION=Base64URL-decoding the payload segment of the token exposes its claims in full, proving the token is fully readable wherever the URL ends up.
+JWT_LEVEL1_PAYLOAD_VALUE=echo 'eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIiwiaWF0IjoxNTE2MjM5MDIyfQ' | base64 -d
+
+JWT_LEVEL2_CHALLENGE=Confirm this level's JWT cookie is exposed to JavaScript running on the page, since it is set without HttpOnly or Secure.
```
preetkaran20: This doesn't sounds like a challenge. I think we can make it more like a challenge for user here. May be like "List the flaws in the JWT storage at client side" or something like that.

src/main/resources/i18n/messages.properties
```diff
@@ -1102,3 +1102,88 @@ UNION_BASED_SQL_INJECTION_VULNERABILITY_LEVEL2_HINT1=Your id value is placed bet
...
+JWT_LEVEL4_PAYLOAD_VALUE=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiJhdHRhY2tlciIsImFkbWluIjp0cnVlfQ.-d0dKlq4zHY09jm50snO-JOwUuaBj8IxyGix5xFXa8s
+
+JWT_LEVEL5_CHALLENGE=Get this level to accept a token with extra bytes appended after its real signature.
+JWT_LEVEL5_HINT1=The signature part is scanned for a URL-encoded null byte (%00); if one turns up, everything after it is silently dropped before the signature is checked.
+JWT_LEVEL5_HINT2=Take any already-valid token for this level and append %00 followed by anything you like to the end of its signature - the check truncates it back to the untouched original bytes it expects.
+JWT_LEVEL5_PAYLOAD_DESCRIPTION=Appending %00 and arbitrary extra bytes to a validly-signed token's signature still validates, since the check strips everything from the null byte onward before comparing.
+JWT_LEVEL5_PAYLOAD_VALUE=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIiwiaWF0IjoxNTE2MjM5MDIyfQ.MZiW2KkIRI6GhKsu16Me7-3IpS4nBw1W47CW67QAqS0%00extra
+
+JWT_LEVEL6_CHALLENGE=Get an unsigned token carrying your own claims accepted by declaring the algorithm as none.
```
preetkaran20: This challenge text contains the hint as well. please fix this.

src/main/resources/i18n/messages.properties
```diff
@@ -1102,3 +1102,88 @@ UNION_BASED_SQL_INJECTION_VULNERABILITY_LEVEL2_HINT1=Your id value is placed bet
...
+JWT_LEVEL4_PAYLOAD_DESCRIPTION=Signing a forged payload with the cracked key "password" produces a token the server verifies as fully valid.
+JWT_LEVEL4_PAYLOAD_VALUE=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiJhdHRhY2tlciIsImFkbWluIjp0cnVlfQ.-d0dKlq4zHY09jm50snO-JOwUuaBj8IxyGix5xFXa8s
+
+JWT_LEVEL5_CHALLENGE=Get this level to accept a token with extra bytes appended after its real signature.
+JWT_LEVEL5_HINT1=The signature part is scanned for a URL-encoded null byte (%00); if one turns up, everything after it is silently dropped before the signature is checked.
+JWT_LEVEL5_HINT2=Take any already-valid token for this level and append %00 followed by anything you like to the end of its signature - the check truncates it back to the untouched original bytes it expects.
+JWT_LEVEL5_PAYLOAD_DESCRIPTION=Appending %00 and arbitrary extra bytes to a validly-signed token's signature still validates, since the check strips everything from the null byte onward before comparing.
+JWT_LEVEL5_PAYLOAD_VALUE=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIiwiaWF0IjoxNTE2MjM5MDIyfQ.MZiW2KkIRI6GhKsu16Me7-3IpS4nBw1W47CW67QAqS0%00extra
+
```
preetkaran20, line 1177: one more thing, all the levels are vulnerable with secure flag in cookie so we can have 2 challenges in each level. one for  newer issues with the implementation and one for secure flag missing

src/main/resources/i18n/messages.properties
```diff
@@ -1102,3 +1102,88 @@ UNION_BASED_SQL_INJECTION_VULNERABILITY_LEVEL2_HINT1=Your id value is placed bet
...
+JWT_LEVEL6_PAYLOAD_DESCRIPTION=A token with alg set to none and an empty signature is accepted as valid regardless of its payload.
+JWT_LEVEL6_PAYLOAD_VALUE=eyJhbGciOiJub25lIiwidHlwIjoiSldUIn0.eyJzdWIiOiJhdHRhY2tlciIsImFkbWluIjp0cnVlfQ.
+
+JWT_LEVEL7_CHALLENGE=Get an unsigned token carrying your own claims accepted through the Authorization header, by declaring the algorithm as none.
+JWT_LEVEL7_HINT1=This endpoint reads the token from the Authorization header instead of a cookie, but checks it the exact same way.
+JWT_LEVEL7_HINT2=A none-algorithm token with an empty signature is accepted here too, just delivered as the Authorization header value instead of a cookie.
+JWT_LEVEL7_PAYLOAD_DESCRIPTION=Sending a none-algorithm token in the Authorization header is accepted as valid, the same check as the cookie-based version of this bypass.
+JWT_LEVEL7_PAYLOAD_VALUE=Authorization: eyJhbGciOiJub25lIiwidHlwIjoiSldUIn0.eyJzdWIiOiJhdHRhY2tlciIsImFkbWluIjp0cnVlfQ.
+
+JWT_LEVEL8_CHALLENGE=Forge a token accepted as valid by signing it with the server's own RSA public key, treated as an HMAC secret.
```
preetkaran20: can you please fix this challenge in a more descriptive way. this is key confusion attack, please try exploiting it.

preetkaran20 (commented):
@pereiravp I think challenges can we refined like we did in sql injection vulnerability,

src/main/resources/i18n/messages.properties
pereiravp: Reworded as an open question, now asks to assess it and list the flaws.

src/main/resources/i18n/messages.properties
pereiravp: Reworded to ask for the flaws, not state the exploit.

src/main/resources/i18n/messages.properties
pereiravp: Fixed, mechanism moved to hints only.

src/main/resources/i18n/messages.properties
pereiravp: Added a second challenge card to every cookie-based level (4,5,6,8,9,10,13,14,15,16) for the missing Secure flag.

src/main/resources/i18n/messages.properties
pereiravp: Named it as a key confusion attack, kept it goal-oriented.

pereiravp:
Reworded every level to state the goal, not the technique, same as SQL Injection. Mechanism stays in the hints only now.

src/main/resources/i18n/messages.properties
```diff
@@ -1133,3 +1133,138 @@ UNION_BASED_SQL_INJECTION_VULNERABILITY_LEVEL2_HINT1=Your id value is placed bet
...
 UNION_BASED_SQL_INJECTION_VULNERABILITY_LEVEL2_PAYLOAD_DESCRIPTION=The leading quote closes the app's string literal, the UNION SELECT filters idor_users down to the ADMIN role, and the trailing -- comments out the app's own closing quote so the combined query stays syntactically valid.
 UNION_BASED_SQL_INJECTION_VULNERABILITY_LEVEL2_PAYLOAD_VALUE=-1' UNION SELECT id, username, salary FROM idor_users WHERE role='ADMIN' --
+
+# JWT Challenge Mode
+JWT_LEVEL1_CHALLENGE=Is this a secure way to pass a JWT from the UI to the backend? If not, list every flaw with it and prove your point by recovering the token yourself.
+JWT_LEVEL1_HINT1=The token travels as a URL query parameter, so it lands in browser history, server access logs, and any Referer header sent to third-party resources on the page.
+JWT_LEVEL1_HINT2=A JWT is only Base64URL encoded, not encrypted, so anyone who gets hold of the URL can decode every claim in plain text.
```
preetkaran20: I think better to mention it as JWS instead of JWT or in brackets as JWS. JWT = (JWS + JWE)

src/main/resources/i18n/messages.properties
```diff
@@ -1133,3 +1133,138 @@ UNION_BASED_SQL_INJECTION_VULNERABILITY_LEVEL2_HINT1=Your id value is placed bet
...
+JWT_LEVEL7_PAYLOAD_DESCRIPTION=Sending a none-algorithm token in the Authorization header is accepted as valid, the same check as the cookie-based version of this bypass.
+JWT_LEVEL7_PAYLOAD_VALUE=Authorization: eyJhbGciOiJub25lIiwidHlwIjoiSldUIn0.eyJzdWIiOiJhdHRhY2tlciIsImFkbWluIjp0cnVlfQ.
+
+JWT_LEVEL8_CHALLENGE=This level is vulnerable to a JWT algorithm confusion (RS256-to-HS256 key confusion) attack. Exploit it to forge a token accepted as valid.
+JWT_LEVEL8_HINT1=This level issues RS256-signed tokens, but the check decides how to verify a token based on the alg value the token itself declares, not on what the server actually used to sign it.
+JWT_LEVEL8_HINT2=Declare alg as HS256 instead, and the check falls back to HMAC verification, using the RSA public key's raw encoded bytes as the HMAC secret - and that public key is meant to be public, not a secret.
+JWT_LEVEL8_PAYLOAD_DESCRIPTION=HMAC-SHA256-signing a forged token with the DER-encoded bytes of the server's own RSA public key produces a signature the check accepts.
+JWT_LEVEL8_PAYLOAD_VALUE=jwt_tool <token> -X k -pk public_crt.pem
```
preetkaran20: https://github.com/SasanLabs/VulnerableApp/blob/master/src/main/resources/sasanlabs.p12 is the key that is used and its password is changeIt which is default password.

src/main/resources/i18n/messages.properties
```diff
@@ -1133,3 +1133,138 @@ UNION_BASED_SQL_INJECTION_VULNERABILITY_LEVEL2_HINT1=Your id value is placed bet
...
+JWT_LEVEL15_PAYLOAD_DESCRIPTION=A three-part token whose signature segment is arbitrary garbage is accepted as fully valid, since the signature itself is never checked.
+JWT_LEVEL15_PAYLOAD_VALUE=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiJhdHRhY2tlciIsImFkbWluIjp0cnVlfQ.garbage-not-a-real-signature
+JWT_LEVEL15_CHALLENGE_2=Confirm this level's JWT cookie would still be sent over a plain, unencrypted connection.
+JWT_LEVEL15_CHALLENGE2_HINT1=HttpOnly stops JavaScript from reading the cookie, but it does nothing to protect it in transit.
+JWT_LEVEL15_CHALLENGE2_HINT2=Without Secure, the browser attaches this same cookie to a plain http request to the same host, not just https.
+JWT_LEVEL15_CHALLENGE2_PAYLOAD_DESCRIPTION=The Set-Cookie response for this level never includes the Secure attribute, confirming the cookie would be replayed over an unencrypted connection to this host.
+JWT_LEVEL15_CHALLENGE2_PAYLOAD_VALUE=curl -v "http://localhost:9090/VulnerableApp/JWTVulnerability/LEVEL_15?fetch=true" 2>&1 | grep -i set-cookie
+
+JWT_LEVEL16_CHALLENGE=Get a token accepted as valid even though its header lies about which algorithm signed it.
```
preetkaran20, line 1261: I am not sure if this is true. can you please go through it again once.

preetkaran20: I think there is a bug in this level implementation.

preetkaran20: Can you please either fix it or comment it out and file a bug to add this vulnerability.

src/main/resources/i18n/messages.properties
pereiravp: Good catch. Reworded to JWS, noted JWT covers JWS and JWE.

src/main/resources/i18n/messages.properties
pereiravp: Done. Payload now extracts the key from sasanlabs.p12 (password changeIt) with openssl. Tested it forges a valid token.

src/main/resources/i18n/messages.properties
pereiravp: Traced customHMACValidator, it always checks HS256 with the server's key regardless of the header's alg, so the challenge is correct, tested the payload and it validates. The bug was the old comment claiming HS384/HS512 support that was never implemented. Fixed the comment.

preetkaran20 (approved):
LGTM. Thanks a lot for the PR @pereiravp

PR #761, opened 2026-08-15
feat: add Challenge mode to Error-Based and Union-Based SQL Injection

pereiravp:
## Summary

Part of #561. Adds `@ChallengeCard` to `ErrorBasedSQLInjectionVulnerability` and `UnionBasedSQLInjectionVulnerability`, the two SQL injection types Blind SQL Injection's challenge mode PR left uncovered.

Edit: redesigned per [@preetkaran20's review](https://github.com/SasanLabs/VulnerableApp/pull/761#pullrequestreview-4945163621). All four levels now extract `admin_sqli`'s real credentials instead of generic bypasses, matching the Blind SQL Injection pattern.

## Scope

UNSECURE levels only, same precedent as Blind SQL Injection:

- Error-Based: levels 1-4 get cards, level 5 (parameterized) doesn't.
- Union-Based: levels 1-2 get cards, levels 3-9 (parameterized/JPA/Criteria/Spring Data) don't.

## Approach

- **Levels 1-2**: CAST-error and UNION SELECT payloads pull `admin_sqli`'s password/username directly, verified against a running instance.
- **Levels 3-4**: quotes are stripped and `id` is INT-typed, so no data-access bypass is possible. Challenge targets the verbose SQL error instead (CWE-209), and level 4 shows a `PreparedStatement` used without `setString` gives no protection.

Every challenge is self-contained, mechanism details live in the hints only.

## Test plan

- [x] `./gradlew compileJava`, `./gradlew test`, `./gradlew spotlessApply` all clean
- [x] All 6 payloads verified against a running instance
- [x] Confirmed `@ChallengeCard` is present on exactly the 6 UNSECURE levels via `/VulnerableApp/allEndPointJson`

## Summary by CodeRabbit

* **New Features**
  * Added interactive challenge cards for error-based SQL injection levels 1–4.
  * Added challenge cards for UNION-based SQL injection levels 1–2.
  * Included localized challenge descriptions, hints, and example payload guidance for each challenge.

PR #760, opened 2026-08-15
feat: add Challenge mode to XXE

pereiravp:
## Summary

Closes #563. Adds `@ChallengeCard` metadata to all 5 levels of `XXEVulnerability`, following the pattern from the merged XSS (#720) and Path Traversal challenge mode work.

## Approach

Verified against a locally running instance, including a local HTTP listener standing in for an attacker's server to confirm out-of-band exfiltration actually fires.

- **Level 1**: direct general entity reads an arbitrary local file. `book.name` is capped at 30 chars, so the hint steers toward a short target file.
- **Level 2**: general entities are blocked, but parameter entities aren't. The classic three-entity chain still works through an external DTD file, confirmed against a local listener.
- **Level 3**: both disabled, the level 2 payload parses harmlessly.
- **Level 4**: any DOCTYPE at all is rejected outright.
- **Level 5**: namespace-awareness plus explicit XInclude disabling closes the one gap that would otherwise let `<xi:include>` slip through without a DOCTYPE.

No frontend changes needed.

## Test plan

- [x] `./gradlew compileJava`, `./gradlew test`, `./gradlew spotlessApply` all clean
- [x] All 5 levels verified against a running instance
- [x] Level 2's out-of-band payload confirmed against a local HTTP listener
- [x] `@ChallengeCard` data confirmed via `/VulnerableApp/allEndPointJson` for every level

## Summary by CodeRabbit

* **New Features**
  * Added five progressive XXE security challenges covering file disclosure, out-of-band data exfiltration, external-entity protection, DOCTYPE rejection, and XInclude protection.
  * Each challenge includes localized descriptions, two hints, and example payload information to guide exploration.
  * Added challenge details to XXE attack-vector exercises, creating a structured learning path for exploring XML security protections.

PR #709, opened 2026-07-23
feat: add challenge mode to LDAP Injection (#568), File Upload (#559), and SSRF (#565)

Mubashir78:
## Summary

Adds challenge card annotations to three vulnerability modules, following the existing pattern established in BlindSQLInjectionVulnerability and ErrorBasedSQLInjectionVulnerability.

### LDAP Injection (#568)
- `LDAPInjectionVulnerability.java`: Added `@ChallengeCard` annotations to levels 1-5 (level 6 is SECURE)
- 25 message keys for challenges, hints, payloads

### File Upload (#559)
- `UnrestrictedFileUpload.java`: Added `@ChallengeCard` annotations to levels 1-9 (level 10 is SECURE)
- 45 message keys for challenges, hints, payloads

### SSRF (#565)
- `SSRFVulnerability.java`: Added `@ChallengeCard` annotations to levels 1-4 (level 5 is SECURE)
- 20 message keys for challenges, hints, payloads

### messages.properties
- Added 116 new message keys across all three modules

## Related Issues
- Closes #568
- Closes #559
- Closes #565

## Summary by CodeRabbit

- **New Features**
  - Added Challenge Mode content for File Upload vulnerabilities across levels 1–9.
  - Added challenge instructions, hints, and payload guidance for LDAP Injection levels 1–5.
  - Added Challenge Mode guidance for SSRF levels 1–4.
  - Introduced localized text supporting the new challenge cards, including descriptions, hints, and payload details.

PR #673, opened 2026-06-20
Add Challenge mode to Authentication Vulnerability

luks-santos:
### Add Challenge mode to Authentication Vulnerability

This PR brings the Authentication module into Challenge mode by adding `@ChallengeCard`annotations to each unsecure leve. Each card exposes a challenge objective, progressive hints, and a gated payload through new i18n message keys.

### What was implemented

- `@ChallengeCard` per unsecure level (1–8 and 10), each with:
    - The challenge objective (challengeText).
    - A payload describing the attack method, not the literal answer.

- Level 10 made solvable. The endpoint already existed but had no seed user, so it could never be completed. Added the `admin_lowcost` user hashed with BCrypt cost factor 4 and extended the algorithm enum in `schema.sql` with `BCRYPT_LOW_ITERATION`. The card teaches that a low work factor makes offline cracking feasible (e.g., `hashcat -m 3200`).

Closes #569 

## Release Notes

* **New Features**
  * Added Challenge Cards for nine authentication vulnerability levels (1–8, 10), featuring challenge prompts, progressive hints, and pre-configured payload examples for interactive learning.
  * Extended authentication method support to include low-iteration bcrypt credentials.

PR #774, opened 2026-08-21
feat: add Challenge mode to Cache Poisoning (#764)

### LINKED ISSUE #764: Add Challenge mode to Cache Poisoning Vulnerability
Updating [Cache Poisoning](https://github.com/SasanLabs/VulnerableApp/tree/master/src/main/java/org/sasanlabs/service/vulnerability/cachePoisoning) to support Challenge Mode, following the same pattern as #556 to #569. Those fourteen issues were all opened on 11 April 2026, the same day as the design issue #555, which they all sit under as sub-issues, and none of them covers Cache Poisoning. The module itself arrived later, in #534, merged on 30 April 2026. It has no challenge-mode issue of its own, and none of its levels carries a card today. See the [Design Doc](https://github.com/SasanLabs/VulnerableApp/wiki/Design-doc-for-Challenge-introduction) for full context.

## What needs to be done

Add `@ChallengeCard` to the four unsecure level methods in `CachePoisoningVulnerability`, and add the matching keys to `src/main/resources/i18n/messages.properties`. Whether level 5 should get one too is a question for you, at the end of the next section. The change is additive: the existing `@AttackVector` and `@VulnerableAppRequestMapping` annotations stay untouched.

## Scope

The class has five levels. Four take the default `UNSECURE` variant and level 5 is the secure reference:

| Level | Variant | `@AttackVector` descriptions | Cache key built by |
|---|---|---|---|
| 1 | UNSECURE | `CACHE_POISONING_LEVEL_1_ROUTE_ONLY_CACHE_KEY`, `CACHE_POISONING_LEVEL_1_XSS` | `buildRouteOnlyCacheKey` |
| 2 | UNSECURE | `CACHE_POISONING_LEVEL_2_NAIVE_FILTER`, `CACHE_POISONING_LEVEL_2_XSS` | `buildRouteOnlyCacheKey` |
| 3 | UNSECURE | `CACHE_POISONING_LEVEL_3_UNKEYED_FORWARDED_HOST`, `CACHE_POISONING_LEVEL_3_JS_INJECTION` | `buildRouteAndBannerCacheKey` |
| 4 | UNSECURE | `CACHE_POISONING_LEVEL_4_PUBLIC_CACHE_PERSONALIZED_CONTENT` | `buildRouteOnlyCacheKey` |
| 5 | SECURE | `CACHE_POISONING_LEVEL_5_SECURE_CACHE_POLICY` | `buildPrivateResponseKey` |

The four unsecure levels teach different mistakes, so the cards would not repeat each other:

- Level 1: `buildRouteOnlyCacheKey` returns `request.getRequestURI()` and nothing else, so the `banner` query parameter never reaches the key.
- Level 2: the same key, plus `applyNaiveBannerFilter`, which removes only `<script>` tags and the `javascript:` scheme. The filter looks like a fix without being one, which is the lesson.
- Level 3: the key is route plus banner, but `X-Forwarded-Host` is read through `resolveUntrustedForwardedHost` and never keyed.
- Level 4: the route-only key again, while the body is personalized from the `demo_user` cookie and the response is marked `public` in `Cache-Control`.

**One thing I would like your call on: should level 5 get a card too?** The tree is not settled on this. At `443c5f4`, sixteen classes carry `@ChallengeCard`. Between them they declare 29 methods marked `Variant.SECURE`, and four of those do carry a card. All four arrived with #720, merged on 11 August. The other thirteen classes hold 24 secure methods and none of them has a card, including the two SQL Injection classes that arrived with #761 on 16 August, whose eight secure methods are all uncarded. Of the three challenge-mode PRs opened on 15 August, #759 and #760 put a card on their secure levels and #758 does not. My default is four cards on levels 1 to 4, and I will add a fifth on the secure level if you would rather follow #720.

Numbers measured on `443c5f4`. To reproduce them from a clone that has fetched `master`:

```sh
R=443c5f4499ef8ef18b8fa9b0d7d78356b96214a4
V=src/main/java/org/sasanlabs/service/vulnerability

# 16 classes carry @ChallengeCard
git grep -l "@ChallengeCard" $R -- "$V" | wc -l

# 29 SECURE methods inside those 16 classes
git grep -l "@ChallengeCard" $R -- "$V" | sed "s|^$R:||" \
  | xargs -I{} git grep -c "variant = Variant.SECURE" $R -- {} \
  | awk -F: '{s+=$NF} END {print s}'

# 35 SECURE methods across every vulnerability class
git grep -c "variant = Variant.SECURE" $R -- "$V" | awk -F: '{s+=$NF} END {print s}'
```

I did not rely on the text search alone. `variant` defaults to `UNSECURE` when a method omits it, so anything touching the unsecure side has to read the resolved annotation rather than the source text. I read them at runtime with `getAnnotationsByType`, the way `EndPointsInformationProvider` builds its responses, and both methods give the same 16, 29 and 35, on a checkout where `./gradlew clean build` passes, with 447 tests and no failures.

## Notes on the message keys

In `BlindSQLInjectionVulnerability` and `LDAPInjectionVulnerability`, every card carries one `challengeText`, two or three `Hint` entries, and a `Payload` with a description and a value, and I would follow that shape. Counted per card rather than from totals, the hints run 2, 3, 3 in the first and 2, 2, 3, 3 in the second. Those two modules reference 17 and 22 distinct keys.

One detail that changes the count here: `getPayload` in `EndPointsInformationProvider` resolves a payload value from `src/main/resources/attackvectors/*.properties` first and only falls back to the message bundle, and `attackvectors/CachePoisoningPayload.properties` already carries eight keys covering all five levels. Reusing those, the way `IDORVulnerability` reuses `IDOR_PAYLOAD_LEVEL_1` to `LEVEL_4`, puts four cards at roughly 16 to 20 new keys in `messages.properties`, four to five per card. Say the word if you would rather have payload text written specifically for the challenges instead.

Challenge keys are English only in practice. Of the 546 keys referenced by the existing cards, 537 are in `messages.properties` and the remaining 9 live in `attackvectors/`. None of the 546 appears in any of the ten locale bundles, so I would keep to that and not touch translations.

## Would like to work on this

I would like to pick this up if it is useful. I am a backend engineer working in Java and Spring Boot day to day, and the cache-key reasoning here is close to what I deal with professionally.

Happy to adjust the wording of the challenges and hints, since those are a teaching decision rather than a coding one, and you may want them pitched differently.

If it is useful, this can sit under #555 as a sub-issue alongside the other fourteen. I only have read access here, so that would have to come from your side.

vianbas:
Closes #764.

Adds a `@ChallengeCard` to all five levels of `CachePoisoningVulnerability` and the thirty message keys they reference. Two files, eighty seven lines, nothing removed. The existing `@AttackVector` and `@VulnerableAppRequestMapping` annotations are untouched.

Rebased onto master `c833c66` on 2 September. The rebase changed no content: at that point the branch carried twenty four keys, four cards and twelve hints, with the same names and values before and after, compared through the contents API. The fifth card below was added after that. `CachePoisoningVulnerability`, its attack vector text and its templates are byte identical between the old base and the new one, so the checks recorded below still describe this branch.

The second commit fixes a defect I found in my own level 3 payload while comparing this PR with #779. The value ended with a sentence of instruction, and level 3 joins the forwarded host into an asset address after only trimming it, so pasting the documented value produced a host containing a comma and ten spaces and the asset never loaded. The value is now the header on its own, and the sentence was dropped because the payload description already said it.

**Level 5 has a card too**

I raised that question on #764 and it is now answered. @preetkaran20 asked whether a card on a secure level is shown in the UI; it is, and I confirmed it in a browser on `XXEVulnerability` LEVEL_5 rather than from the code alone. @pereiravp established the practice earlier in #720 and #759, and #760 added more.

So level 5 now carries one, at eleven Java lines and six keys, which is what each of the other four measures. Its card asks the reader to confirm rather than to attack: every payload from the levels before it is accepted here and none survives. The response is `private, no-store` and is never stored, the banner comes back with its angle brackets encoded, an attacker forwarded host never reaches the asset address, and a request without a cookie sees the default user rather than another user's details.

**One goal per level rather than one description per level**

Each card asks for something the reader has to achieve, which is the direction #761 was revised in:

| Level | The card asks for |
|---|---|
| 1 | a visitor who sends no parameters at all to receive a banner the attacker chose |
| 2 | script to run in another visitor's browser even though the banner is filtered first |
| 3 | every visitor on the same banner to load a page asset from a host the attacker controls |
| 4 | another user's email address and last login IP, read without sending any cookie |

The hints describe what the level does rather than quoting its source. Across each set they cover the input the page reflects, what the cache key actually contains, and what the gap between the two allows. Level 2 is the exception: its middle hint is about getting past the filter, because there the key is unchanged from level 1 and the filter is the new thing.

**A note on the payload field**

On #764 I proposed reusing the existing `attackvectors/CachePoisoningPayload.properties` text as the cards' payload value, the way `IDORVulnerability` reuses its own. Opening the cards in the browser showed that it does not work here: the card renders that field inside a `<code>` block as plain text, and the cache poisoning entries are HTML walkthroughs, so their tags show up literally. `IDORVulnerability` gets away with it because its payload strings carry no markup.

The one hundred and thirty three `PAYLOAD_VALUE` entries in the bundle at `c833c66`, the revision this branch now builds on, run from one to four hundred and eighty four characters and none is a formatted HTML walkthrough, while the cache poisoning attack vector text runs from four hundred and seven characters up to two thousand five hundred and four. So these cards carry five short values of their own instead, between thirty four and a hundred and thirty characters, which keeps the bundle inside the same range it already had. The attack vector text is untouched and still serves the help panel.

**How I checked it**

Every goal above was reached against a running instance, twice, each time from a cleared cache. On one of the two passes the `X-Cache-Status` and `X-Cache-Key` headers were read rather than assumed; the other checked the outcome alone.

```sh
./gradlew clean build
java -jar build/libs/VulnerableApp-1.0.0.jar > app.log 2>&1 &
B=http://localhost:9090/VulnerableApp/CachePoisoning
do sleep 2; done

# level 1: poison, then ask with nothing at all
curl -s -o /dev/null -X POST "$B/clearCache?level=LEVEL_1"
curl -s -o /dev/null "$B/LEVEL_1?banner=POISONED-BY-ATTACKER"
curl -s "$B/LEVEL_1" | grep -c POISONED-BY-ATTACKER

# level 4: one user's page, fetched by someone sending no cookie
curl -s -o /dev/null -X POST "$B/clearCache?level=LEVEL_4"
curl -s -o /dev/null -H "Cookie: demo_user=alice" "$B/LEVEL_4"
curl -s "$B/LEVEL_4" | grep -c alice@vulnerableapp.local
```

Both print `1`. The loop is there because the jar returns before the application is serving. Levels 2 and 3 follow the same shape, level 2 with an event-handler attribute in the banner and level 3 with `X-Forwarded-Host` alongside a banner that both requests share.

Three controls, because a passing attack does not by itself prove the hint text is accurate:

- Level 2 hint 1 says the filter removes one tag and one scheme. A `<script>` banner is stripped, so the filter is real and the event-handler attribute is a bypass rather than an absent filter.
- Level 3 hint 1 says the banner is keyed and escaped, so it is not the way in. `X-Cache-Key` reads `/VulnerableApp/CachePoisoning/LEVEL_3|banner=shared-status`, and a banner containing a tag comes back as `&lt;img`. Keying alone would not save it, since an entry poisoned through the banner still reaches anyone asking for that same banner; the escaping is what stops it.
- Level 4 hint 2 says the cookie never reaches the key. A request carrying `demo_user=bob` gets a `HIT` on the entry alice created, which is stronger than the anonymous case on its own.

For level 3 the check was the victim's markup rather than the raw body: the `iframe src` and the link both read `https://attacker.example/assets/cache-poisoning-demo.js`, and the trusted default host appears nowhere in the response.

Level 5 was exercised with both an attacker forwarded host and a cookie set. It answers `private, no-store` with `X-Cache-Status: MISS` and does not reflect the host, so the secure reference still behaves as its own card-less description says.

All thirty keys were read back from `/VulnerableApp/allEndPointJson` to confirm each resolves to English text with no raw key surfacing, and the cards were opened in the application's own challenge mode to see how they render.

**The level 2 payload was run in a browser**

A banner of `` went through the level's own attacker control, the title was set to a known marker, and the victim control was clicked. The title came back changed with `X-Cache-Status: HIT`, so the handler ran in the page that received the cached response, not in the one that poisoned it. Twice, the second time from a cleared entry.

Worth knowing if you reproduce it: with the level's browser cache toggle left on, the victim request can be answered by the browser's own store and never reach the application, showing a stale `MISS` and no execution. Turning it off makes the shared cache the only one in play.

**What I did not do**

No tests were added. Eight challenge-mode PRs have merged, #661, #673, #709, #720, #759, #760, #761 and #767, and none of them added a test file. Seven touched only their vulnerability classes and message bundles; #673 also touched two Authentication SQL scripts. So this follows them. When this PR was opened I wrote that only two had merged, which was wrong at the time and not merely out of date. Whether cards deserve coverage of their own is a separate question and I did not settle it here.

The wording of the challenges and hints is a teaching decision rather than a coding one, so please rewrite anything that is pitched wrong. The keys and structure would not change.

`./gradlew clean build` passes with 448 tests, 0 failures, 2 skipped, which is the same count as master at `c833c66`, measured by building both, and re-run after the fifth card was added. The extra test against the earlier count is `MessageBundleDuplicateKeyTest`, which arrived on master with #776 and runs against these keys for the first time after the rebase.

## Summary by CodeRabbit

* **New Features**
  * Added challenge-mode content for Cache Poisoning levels 1–4, including localized instructions, progressive hints, payload guidance, and practical attack scenarios.
  * Added challenge content for Path Traversal levels 1–13, XXE levels 1–5, and JWT levels 1–16.
* **Documentation**
  * Expanded explanations covering cache behavior, filtering bypasses, host manipulation, personalized-data exposure, and related vulnerability techniques.
  * Clarified level-specific descriptions for XSS HTML tag injection attack vectors.


vianbas:
Two open PRs close #764: this one and #779 by @pereiravp. That is my doing for not spotting theirs sooner, so here is the short version rather than another wall of text.

On #764 I set out what actually differs between them, measured: this PR puts one card on each of levels 1 to 4, #779 puts one on each of the seven attack vectors those levels carry. Both shapes already exist on master, so it is a teaching preference and not a correctness question. Neither PR cards level 5.

You do not have to referee it. If the per attack vector shape is the one you want for challenge mode, say so and I will close this PR, and you have one to look at instead of two.

Two housekeeping notes so nothing here needs checking. This branch was rebased onto `c833c66` today and the rebase changed no content, the same twenty four keys with the same values. `./gradlew clean build` gives 448 tests and 0 failures, the same count as master. The body above was corrected today as well. Six numbers in it were wrong: three the rebase made stale, one my own payload fix made stale, and two that were already wrong the day I opened this PR.

vianbas:
Thanks @pereiravp, that reasoning holds. Each of levels 1 to 3 does carry two `@AttackVector` annotations, the cache poisoning primitive and the injection it enables, so a card per attack vector maps onto a distinction the code already makes. I checked both PRs against a running instance and every card in each works: the unkeyed field genuinely poisons the shared cache, confirmed from the `X-Cache-Key` and a `HIT` on the following request, and the secure level stays `private, no-store`.

So the choice between them is only the shape. #779 puts a card on each of the seven attack vectors across levels 1 to 4; this PR puts one per level and covers four of them, with three hints per card against two. Both compile, pass the suite, and behave correctly.

@preetkaran20 your call on which shape fits challenge mode. If you want #779's, say so and I will close this one and, if it is useful, send the three-hint laddering as a small follow-up on top of it. The level 5 question I raised earlier is still open either way.

preetkaran20:
@vianbas @pereiravp I think of each challenge card primarily as a way to teach students about a vulnerability. Because of that, I believe it’s perfectly fine and even better to have more challenge cards than attack vectors.

Attack vectors are primarily useful for scanners, whereas challenge cards are designed around the learning experience and the challenges we want to present to students.

My simple philosophy would be: **if a level contains 2 vulnerabilities, it should have at least 2 challenge cards and exactly 2 attack vectors.**

This gives us flexibility to create multiple learning challenges around the same vulnerability while keeping the attack vectors focused and meaningful for scanner testing.

pereiravp:
@preetkaran20 that lines up with what #779 already does, one card per attack vector rather than per level, so levels 1 to 3 get two cards each and level 4 gets one, seven total against seven attack vectors. Good to have the rule spelled out for future levels too.

vianbas:
#779 is merged, so this is closing as I said it would.

Congratulations @pereiravp, and thanks for the reasoning on #764 that settled the shape question. Your split maps onto the two attack vectors each of levels 1 to 3 already carries, which mine did not.

One piece of this PR is not covered by what merged: cache poisoning is now the only challenge mode module whose secure level has no card, and @preetkaran20 said on #764 that there is no harm in adding one. I have opened that as a small separate PR, rebuilt in the shape the module now uses rather than copied across from here.

Issue #747, opened 2026-08-15
[Community Testing] XXE


preetkaran20:
## Community Testing Request

We’d love community testing feedback for **XXE** in VulnerableApp.

Please test on **both UIs**:
- **Legacy UI** (`http://localhost:9090/VulnerableApp`)
- **Modern Docker-based UI** (`docker-compose` stack at `http://localhost`)

Please share:
- Reproduction steps you used
- Payloads/attack paths that worked (or failed)
- Observed behavior vs expected behavior
- Differences between legacy and modern UI behavior (if any)
- Ideas to improve challenge clarity, realism, or difficulty
- Any blockers/challenges during testing

Your findings will help improve the lab quality and learning experience.

sharankumar-k:
Hi @preetkaran20, I’d like to work on this issue. I’ll test the XXE vulnerability on both the Legacy UI and Modern Docker-based UI and share the reproduction steps and findings.

sharankumar-k:
## XXE Community Testing Results

I tested the legacy UI across XXE Levels 1–5.

### Legacy UI

#### Level 1 — General External Entity

- **Result:** XXE confirmed.
- **Endpoint:** `POST /VulnerableApp/XXEVulnerability/LEVEL_1`
- Tested an external entity referencing `file:///C:/Windows/win.ini`.
- The application resolved the entity and returned the contents of `win.ini` in the `others` field.
- This confirms local file disclosure through XXE.

#### Level 2 — General External Entity

- **Result:** Blocked for the tested payload.
- The request returned `isValid: true`, but the external entity was not expanded and `others` was empty.

I also tested a parameter-entity payload on Level 2. It returned: `{"content":null,"isValid":false}`

No file disclosure was demonstrated with that test.

#### Level 3 — General External Entity

- **Result:** Blocked for the tested payload.
- The external entity was not expanded and `others` remained empty.

#### Level 4 — DOCTYPE

- **Result:** Blocked.
- The XML containing the `DOCTYPE` declaration was rejected.
- Response: `{"content":null,"isValid":false}`

#### Level 5 — DOCTYPE

- **Result:** Blocked.
- The XML containing the `DOCTYPE` declaration was rejected.
- Response: `{"content":null,"isValid":false}`

### Expected vs Observed

For the attack paths tested on the legacy UI:

- Level 1 allowed the XXE file-disclosure payload.
- The tested Level 2 and Level 3 external-entity payloads were not expanded.
[...]

PR #775, opened 2026-08-21
Fix #770: Use standard Deprecation and Sunset headers

### LINKED ISSUE #770: GET /scanner sends Deprecation: true, which RFC 9745 does not allow, and sends no Sunset header for the September 2027 removal
**Describe the bug**

`GET /VulnerableApp/scanner` sends `Deprecation: true`. RFC 9745 makes that field's value a Date, so `true` is not something the field can carry, and a client parsing it the way the RFC describes gets a type it cannot use. The same endpoint sends no `Sunset` header, so the September 2027 removal date exists only inside a javadoc comment and nothing on the wire states it.

Both of these are mine, from #733, and both are small. They are here as one ticket because they are the two halves of the same signal.

**The header value: RFC 9745 wants a date, and this follows the expired draft instead**

RFC 9745 section 2.1: *Deprecation is an Item Structured Header Field; its value MUST be a Date as per Section 3.3.7 of [RFC9651]*. A Date is `@` followed by a count of seconds, for example `Deprecation: @1688169599`.

`true` is not one. Under RFC 9651 a Date is `sf-date = "@" sf-integer`, while `true` matches `sf-token = ( ALPHA / "*" ) *( tchar / ":" / "/" )`, so a conforming parser reads a Token. It is not a Boolean either, since those are written `?1` and `?0`.

`Deprecation = IMF-fixdate / "true"` was the syntax in `draft-dalal-deprecation-header-03`. That draft expired, the work moved to the HTTPAPI working group, and it landed as RFC 9745 in March 2025. I proposed `true` on #725 before writing any code and then built what I proposed, which is how it reached master.

That it has to be a date at all is independent of how we read September 2027. Section 2.1 also says the value *conveys the deprecation date, which may be in the future (the resource in context will be deprecated at that date) or in the past (the resource in context was deprecated at that date)*. The reading decides which of the two it is, not whether it is a date, so `true` is wrong either way.

**The missing header: this one your answer already settled**

On #733 I asked which of two readings of September 2027 you meant, and you answered *Yes I mean first one*: the reading where September 2027 is when `/scanner` goes away. In the same comment I had written that the HTTP counterpart of that reading is a `Sunset` header, which the branch did not send. Your answer landed 93 seconds before the merge, so there was no room to add it there.

RFC 8594 section 3 defines `Sunset = HTTP-date`, so September 2027 is written `Sunset: Thu, 30 Sep 2027 23:59:59 GMT`.

RFC 9745 section 4 is the only constraint tying the two together: *The timestamp given in the Sunset HTTP header field MUST NOT be earlier than the one given in the Deprecation header field.* Both exact dates are your call.

**To Reproduce**

1. Build and start the application:

```
./gradlew build
```

2. Read the response headers of the deprecated endpoint:

```
curl -sI http://localhost:9090/VulnerableApp/scanner | grep -iE '^(Deprecation|Sunset|Link)'
```

```
Deprecation: true
Link: <http://localhost:9090/VulnerableApp/scanner/dast>; rel="successor-version"
```

There is no `Sunset` line. `grep -ric sunset src/` returns `0` for every file.

3. The successor endpoint carries no such headers, which is intended:

```
curl -sI http://localhost:9090/VulnerableApp/scanner/dast | grep -icE '^(Deprecation|Sunset)'
```

```
0
```

**Expected behavior**

`GET /scanner` sends a `Deprecation` value a conforming client can read, and a `Sunset` value stating when the endpoint goes away. With the deprecation dated at the #733 merge, that is:

```
Deprecation: @1786896221
Sunset: Thu, 30 Sep 2027 23:59:59 GMT
Link: <http://localhost:9090/VulnerableApp/scanner/dast>; rel="successor-version"
```

**Additional context**

`src/test/java/org/sasanlabs/controller/VulnerableAppRestControllerTest.java:88` asserts `header().string("Deprecation", "true")`, so the current value is pinned by a test and a fix has to change that assertion too. The javadoc on `getScannerRelatedInformation` names September 2027 in prose; once `Sunset` carries it, the two agree.

The `Link` header needs no change. `successor-version` is registered by RFC 5829 section 3.6 for a link to the successor version in the version history, which is how it is used here. RFC 9745 separately registers `rel="deprecation"` for human readable documentation about the deprecation, and this application has no such page, so leaving that one out looks right to me.

Worth settling rather than leaving: SasanLabs/VulnerableApp-php#20 and SasanLabs/VulnerableApp-jsp#17 ask for the same two endpoints in those apps and are both still open, so whichever form of this signal stands here is the one they have to match.

Yes. It is one constant beside the existing header names, two lines in `getScannerRelatedInformation`, the assertion that pins the current value, and a mention in the javadoc, which lists the headers the response carries. I would rather correct my own PR than leave it. Tell me the two dates you want and I will send it.

sharankumar-k:
## Description

Fixes #770.

- Replace the invalid `Deprecation: true` value with the RFC 9745 date format.
- Add the `Sunset` header for the September 2027 removal date.
- Update the existing controller test to verify both headers.
- Keep the existing `Link` successor-version header unchanged.

## Testing

- `gradle test` 
- `git diff --cached --check` 

## Summary by CodeRabbit

* **Enhancements**
  * Updated the deprecated `/scanner` endpoint to return standards-compliant deprecation and sunset timestamps.
  * Added a `Sunset` response header indicating removal is planned for September 2027.
  * Preserved the link to the recommended `/scanner/dast` replacement endpoint.
  * Confirmed that `/scanner/dast` and `/scanner/sast` responses do not include a sunset header.
* **Documentation**
  * Recommended `/scanner/dast` for new integrations and clarified the updated response headers.


vianbas:
I filed #770, and the defect it describes is mine, from #733. I said there that I would rather correct my own PR than leave it. You got there first and what you have is right, so I am not opening a competing PR.

I checked it rather than taking it on trust.

Both dates hold up. `@1786896221` is 2026-08-16T16:03:41Z, which is the instant #733 merged, so the deprecation is dated at the moment its replacement landed. 30 September 2027 really is a Thursday, which matters more than it sounds: an HTTP-date whose day name disagrees with the date is malformed, and this one agrees.

I also ran it. On the deprecated path the response now carries the timestamped `Deprecation`, the `Sunset`, and the unchanged `Link`, and both `/scanner/dast` and `/scanner/sast` carry none of the three.

One thing worth saying out loud rather than leaving implied. The part that decides the shape is already settled: on #733 you confirmed, @preetkaran20, that September 2027 is when `/scanner` goes away, which is what makes a `Sunset` in that month the right header to send. The two exact timestamps were never settled. Both came out of my issue, where I wrote that they are yours to choose, and #770 has no reply yet. Nothing structural turns on them; only the two literals would move.

One suggestion. The assertions pin the literals, so they cannot catch the property the RFC actually cares about. I moved `Sunset` fifteen days earlier than `Deprecation` in the controller and updated the assertion to match, which is what anyone changing a date would do, and the full build stayed green, all 447 tests, even though RFC 9745 section 4 says `Sunset` must not be earlier than `Deprecation`. Parsing the two and comparing them is a few lines, and it would have caught that.

Smaller one: the javadoc now reads as though all three headers name the successor, but only `Link` does.

sharankumar-k:
Thanks for the review. I’ve addressed the feedback, including the SAST no-header assertions. All checks are now passing.

vianbas:
Both hold up, and I ran them rather than taking the commits at their word. The ordering assertion turns the case I described red: with `Sunset` moved fifteen days earlier than `Deprecation`, the build now fails where it passed before. The javadoc reads right now too.

One narrow gap is left in the same place. `deprecation.substring(1)` never checks that the value starts with `@`. Take the `@` off `@1786896221` and update the literal assertion the way anyone would, and the build stays green: the substring eats the leading digit instead of the marker, what is left parses as December 1994, and 1994 is still before the sunset, so the ordering check is satisfied. To be fair to the guard, it does stop the literal original bug: put `true` back and the parse throws. The hole is narrower, a value that is all digits with the marker missing, which is still not a Date under section 2.1. One assertion before the parse, that the value starts with `@`, closes it.

The day name is safe, by the way. I tried a wrong one and `RFC_1123_DATE_TIME` throws, so that part needs nothing.

All of this is against your latest push, with the sast assertions in place.

preetkaran20 (approved):
LGTM. Thanks a lot for the PR

preetkaran20:
🎉 Thanks for contributing @sharankumar-k!
   https://docs.google.com/forms/d/e/1FAIpQLSfwWVdnULUhtfruA-DN328NwKnBGebaWg9U5y0xivLLxxoMog/viewform?usp=pp_url&entry.1414771947=sharankumar-k
   
👉 [#747 - [Community Testing] XXE](https://github.com/SasanLabs/VulnerableApp/issues/747)

preetkaran20:
@all-contributors please add @sharankumar-k for code

PR #776, opened 2026-08-21
fix: prevent duplicate i18n message keys

### LINKED ISSUE #769: XSS_DIRECT_INPUT_DIV_TAG is defined twice in all 11 message files, so one translation of it is dead in every language
**Describe the bug**

`XSS_DIRECT_INPUT_DIV_TAG` is defined twice in every one of the 11 files under `src/main/resources/i18n/`. `java.util.Properties` keeps the last definition and silently drops the first, with no warning, so the first translation can never be displayed in any language.

`messages.properties`, lines 50 and 53:

```properties
XSS_DIRECT_INPUT_DIV_TAG=Html Tag is directly injected into div tag.
XSS_DIRECT_INPUT_DIV_TAG=Url Parameters are added directly into div tag.
```

Searching the whole repository, and not only the Java sources, the key has exactly one consumer outside the message files themselves: `XSSWithHtmlTagInjection.java` line 35. What it serves is the second value.

The served text is not wrong. It matches what `getVulnerablePayloadLevel1` actually does, which is to put query parameter values into a div. The cost is that the dropped sentence was translated into all 11 files and none of those translations can ever appear, and that anyone editing the first definition would see no effect and get no error explaining why.

<details>
<summary>All 11 files, with the text that is dropped</summary>

| File | Lines | Dropped value |
|---|---|---|
| `messages.properties` | 50, 53 | `Html Tag is directly injected into div tag.` |
| `messages_en_US.properties` | 46, 49 | `Html Tag is directly injected into div tag.` |
| `messages_es.properties` | 46, 49 | `La etiqueta HTML se inyecta directamente en la etiqueda div.` |
| `messages_it.properties` | 46, 49 | `Il tag HTML viene iniettato direttamente nel tag div.` |
| `messages_sv.properties` | 44, 47 | `HTML-taggen injiceras direkt in i div-taggen.` |
| `messages_hi.properties` | 46, 49 | `Html टैग सीधे div टैग के अंदर एम्बेडेड होते हैं।` |
| `messages_bn_IN.properties` | 35, 38 | `HTML ট্যাগ সরাসরি div ট্যাগে ইনজেক্ট করা হয়।` |
| `messages_mr_IN.properties` | 35, 38 | `HTML टॅग थेट div टॅगमध्ये इंजेक्ट केला जातो.` |
| `messages_gu_IN.properties` | 44, 47 | `HTML ટેગ સીધી રીતે div ટેગમાં ઇન્જેક્ટ કરવામાં આવે છે.` (stored as `\u` escapes) |
| `messages_ko.properties` | 43, 46 | `HTML 태그가 div 태그에 직접 주입됩니다.` |
| `messages_zh_TW.properties` | 46, 49 | `Html 標籤被直接植入 div 標籤裡。` |

</details>

**To Reproduce**

1. Confirm the key is defined twice:

```
grep -n '^XSS_DIRECT_INPUT_DIV_TAG=' src/main/resources/i18n/messages.properties
```

```
50:XSS_DIRECT_INPUT_DIV_TAG=Html Tag is directly injected into div tag.
53:XSS_DIRECT_INPUT_DIV_TAG=Url Parameters are added directly into div tag.
```

2. Build and start the application:

```
./gradlew build
```

3. Count how often each of the two reaches the API:

```
curl -s http://localhost:9090/VulnerableApp/VulnerabilityDefinitions \
  | grep -o 'Html Tag is directly injected into div tag\.' | wc -l

curl -s http://localhost:9090/VulnerableApp/VulnerabilityDefinitions \
  | grep -o 'Url Parameters are added directly into div tag\.' | wc -l
```

```
0
1
```

The same holds per language: under `Accept-Language: es`, `sv` or `hi` the localized response serves only the second definition and never the first. For `es`, grepping the second value `Los parámetros de la URL se añaden directamente a la etiqueta div.` returns 1 while its dropped first value returns 0; `sv` and `hi` behave the same with their own strings.

**Expected behavior**

A key should be defined once per file. A second definition should not be able to remove a translated string without any signal.

**Additional context**

This is the clash predicted in #114 back in 2020: *"it is possible that message keys clash"*. It has now happened, silently, and it propagated into every translation. The scale is still small, 1 duplicate among 716 keys in `messages.properties`, which is probably why nobody noticed.

I am not proposing the key-prefix redesign from #114 here. That is a larger decision and it is yours to make. This ticket is the concrete instance plus cheap protection against the next one.

*Suggested fix, in two parts, where the second matters more than the first.*

1. Delete one of the two definitions in each of the 11 files. Which one to keep is your call. The second is the one in use today and it matches the code, so removing the first is the change with no difference in behaviour.

2. Fail the build on a duplicate key. `java.util.Properties` will not report one, but a subclass that overrides `put` sees every definition before the last one wins, which is how I measured this. That is a small test over the files in `src/main/resources/i18n/`, and it turns the next clash into a red build instead of a translation that quietly disappears.

If you want part 2, tell me which form you prefer, a plain JUnit test alongside the existing ones or a Gradle check task, and I will send it together with part 1.

*How I measured it.* Loaded each file with a `Properties` subclass that overrides `put` and records every definition rather than only the surviving one, using `Files.newBufferedReader` with the charset each file is actually stored in. Then confirmed the served value over HTTP per locale, as shown above.

Yes, for both parts, once you have said which form part 2 should take.

sharankumar-k:
## Summary

- Remove duplicate `XSS_DIRECT_INPUT_DIV_TAG` keys from all i18n message bundles.
- Add a regression test to prevent duplicate message keys.

## Testing

- `gradle test`
- Verified `XSS_DIRECT_INPUT_DIV_TAG` appears exactly once in every i18n bundle.

Closes #769

## Summary by CodeRabbit

* **Improvements**
  * Standardized HTML tag injection descriptions into three clear severity levels across supported languages.
  * Preserved existing translated messaging while improving consistency in security-related labels.

* **Bug Fixes**
  * Removed a redundant security-message entry from localized content.

* **Tests**
  * Added validation to detect duplicate localization keys and confirm that translation bundles are available before checks run.


vianbas:
I filed #769, so let me start there: that issue says I would send both parts. I am not doing that, and I will not open a competing PR. This one is yours.

I ran your check rather than only reading it.

Against master with the duplicates still in place it fails, and it names all eleven bundles and only `XSS_DIRECT_INPUT_DIV_TAG`. On your branch it passes. So it detects the thing it is meant to detect, which is the part of a check like this that is easy to leave unproven.

I also added two unrelated duplicate keys to two other bundles. It caught both and named the right file for each, then went green again once I reverted them. So it is not tied to this one key.

The surviving line is the right one to keep. `XSSWithHtmlTagInjection` reads the key through `@AttackVector`, and the value you kept is the one about url parameters going into the div, which is what `getVulnerablePayloadLevel1` actually does. Nothing user visible changes.

One thing I would change. The check can pass without checking anything. `Files.list` does not recurse, so if the bundles ever move a directory deeper, or the extension changes, the list comes back empty, `duplicates` stays empty, and the test is green. I reproduced that: with the bundles moved one level down and a real duplicate sitting in one of them, the build still succeeded. One assertion that the bundle list is not empty closes it.

Worth having either way, since `./gradlew clean build` runs this in CI, so it will hold the line for whoever adds the next key.

sharankumar-k:
Thanks for the detailed review and for testing the check against both the existing duplicates and additional duplicate keys.

I’ve added the bundle discovery assertion as suggested and verified the formatting and full build locally. The CI checks are now passing.

Really appreciate you taking the time to verify the behavior and point out that edge case!

src/main/resources/i18n/messages.properties
```diff
@@ -47,7 +47,6 @@ XSS_QUOTES_AND_WITH_HTML_ESCAPE_PLUS_FILTERING_ON_INPUT_SRC_ATTRIBUTE_IMG_TAG=Ur
 
 ## Html Tag Injection
 XSS_HTML_TAG_INJECTION=Html Tag based XSS attack.
-XSS_DIRECT_INPUT_DIV_TAG=Html Tag is directly injected into div tag.
```
preetkaran20, line 50: This Key is used at multiple places like it is used at XSS controller description as well as at the payload. can you please break it into 2. one for vulnerability description and one for payload. we need to do this for all the keys.

preetkaran20 (changes requested):

sharankumar-k:
Thanks for the feedback. I’ve separated the vulnerability description and payload into separate message keys and updated all the affected usages and localized bundles.

preetkaran20 (approved):
I relooked and i think i was wrong earlier. PR LGTM thanks a lot for the PR @sharankumar-k.

PR #779, opened 2026-08-30
feat: add Challenge mode to Cache Poisoning

### LINKED ISSUE #764: Add Challenge mode to Cache Poisoning Vulnerability

pereiravp:
## Summary
Closes #764.

Adds `@ChallengeCard` annotations to the unsecure Cache Poisoning levels (1-4), following the same pattern used for XXE (#760) and Path Traversal challenge mode. Purely additive — `@AttackVector` and `@VulnerableAppRequestMapping` are untouched, level 5 (secure) is left as-is.

Levels 1-3 each carry two `@ChallengeCard`s (matching their two `@AttackVector`s: the cache poisoning primitive and the resulting reflected XSS/JS injection). Level 4 gets a single card for the personalized-content-cached-as-public issue.

## Test plan
- [x] `./gradlew compileJava`

## Summary by CodeRabbit

* **New Features**
  * Added Cache Poisoning Challenge Mode with four progressive levels.
  * Introduced guided challenge descriptions, hints, payload details, and values.
  * Added alternative XSS and JavaScript-injection challenge paths for applicable levels.
  * Covers unkeyed parameters, headers, filtering bypasses, and cached personalized content.
* **Documentation**
  * Updated localized challenge text with clearer, more concise objectives.
  * Simplified the level-three JavaScript-injection payload description.


src/main/resources/i18n/messages.properties
```diff
@@ -711,6 +711,41 @@ CACHE_POISONING_LEVEL_3_UNKEYED_FORWARDED_HOST=This level includes the query par
...
 CACHE_POISONING_LEVEL_4_PUBLIC_CACHE_PERSONALIZED_CONTENT=This level demonstrates the danger of caching personalized content (driven by cookies) as public. The response body varies per user (username, email, last login IP) because it reads the <code>demo_user</code> cookie, but the cookie is an <b>unkeyed input</b> - it never participates in the cache key. The shared cache stores the response under the route alone with <code>Cache-Control: public</code>, so the first requester's PII is served to every subsequent visitor for the rest of the TTL.
 CACHE_POISONING_LEVEL_5_SECURE_CACHE_POLICY=This secure level demonstrates proper cache isolation. It ignores untrusted forwarding headers and marks personalized or sensitive responses as <code>private</code> and <code>no-store</code>, preventing shared caches from storing and reusing them across different users.
+CACHE_POISONING_LEVEL_1_CHALLENGE=Poison the shared cache through the unkeyed 'banner' parameter so a later request to the same route (without the parameter) returns your injected content.
```
preetkaran20, line 714: can you please make the language simple as well as challenge close ended? this applies to all the challenge texts.

preetkaran20: @vianbas can you please also review this PR?

pereiravp: simplified all the challenge texts to short, closed-ended goals and moved the how into the hints. pushed in 555a253.

preetkaran20 (approved):
Lgtm. Thanks a lot for the PR @pereiravp

pereiravp:
Thanks for being a good mentor @preetkaran20

vianbas:
One small follow-up on the level 3 JavaScript injection card, measured on master `2e068b7`. Not a defect in the code, and nothing here blocks anything; it is the goal text being harder to finish than it reads.

The card asks the reader to "Make the cached page load and run a script from a domain you control", and ships `X-Forwarded-Host: attacker-controlled.example` as the payload. Following that exactly, the cache is poisoned and the asset address does change, but nothing runs. I cleared the cache, sent the header once with a banner, then requested the same banner again, and the cached response carries this in its iframe:

```
https://attacker-controlled.example/assets/cache-poisoning-demo.js
```

`.example` is reserved by RFC 2606, so it never resolves and the iframe cannot load.

It is completable in principle. The second hint does say to host a script at the attacker controlled domain, so that part is signposted. What is not mentioned anywhere in the card is that `buildAssetUrl` hardcodes the `https` scheme, so the origin has to serve that exact path over HTTPS with a certificate the browser accepts inside an iframe. None of the five keys on this card mentions HTTPS or a certificate. That is a fair amount of unsignposted setup for a lab exercise.

What the level demonstrates reliably, with no setup at all, is that the asset address itself becomes attacker controlled, which is visible in the response and easy to check. Wording the goal that way would make it completable with the payload you already ship, and the JavaScript execution could stay in the hint as the consequence rather than the target.

Entirely your call, and I am happy to be told this is intentional.

PR #780, opened 2026-08-30
feat: add Challenge mode to Cryptographic Failures

### LINKED ISSUE #558: Add Challenge mode to Cryptography Failures
updating [Cryptography Failures](https://github.com/SasanLabs/VulnerableApp/tree/master/src/main/java/org/sasanlabs/service/vulnerability/cryptographicFailures)

## What needs to be done

## Annotation details

It has three fields:

pereiravp:
## Summary
Closes #558.

Adds `@ChallengeCard` annotations to unsecure Cryptographic Failures levels 1-10, following the same pattern used for XXE, Path Traversal and Cache Poisoning (#779). Purely additive — `@AttackVector` and `@VulnerableAppRequestMapping` are untouched, level 11 (secure Bcrypt) is left as-is.

Note: this issue had an assignee who was last active on it about 4 weeks ago with no PR opened since. I asked in the issue if it was still being worked on and got no response, so picked it up.

## Test plan
- [x] `./gradlew compileJava`

## Summary by CodeRabbit

- **New Features**
  - Added challenge cards for cryptographic-failure levels 1–10.
  - Each challenge includes instructions, two progressive hints, payload descriptions, and example values.
  - Added guidance for plaintext exposure, Base64 decoding, weak hashing, password cracking, and insecure AES key usage.

- **Documentation**
  - Clarified Level 1 instructions to retrieve the plaintext password from the database.
  - Refined Levels 2–10 wording to provide clearer guidance for decoding or recovering passwords, including level-specific hints.

### DIFF
--- a/src/main/resources/i18n/messages.properties
+++ b/src/main/resources/i18n/messages.properties
@@ -506,6 +506,56 @@ CRYPTOGRAPHIC_FAILURES_LM_HASHING=Password is hashed using LM algorithm which ha
 CRYPTOGRAPHIC_FAILURES_SHA256_HASHING=Password is hashed using SHA-256 without a salt, making it vulnerable to rainbow table and precomputed hash attacks. The user must crack the SHA-256 hash using online tools to find the original password.
 CRYPTOGRAPHIC_FAILURES_INSECURE_AES128=Password is encrypted with AES-128 but implemented with a weak key making it insecure to brute-force attacks. Encryption is a two-way function meaning that anyone with the key can decrypt the password. The user must crack the password using common password lists or key-recovery techniques.
 CRYPTOGRAPHIC_FAILURES_SECURE_BCRYPT=Password is hashed using Bcrypt with salt with a work factor (strength) of at least 4, making any type of brute-force or rainbow table attacks infeasible.
+CRYPTOGRAPHIC_FAILURES_LEVEL_1_CHALLENGE=Recover the password without any cracking at all, since the system exposes it directly.
+CRYPTOGRAPHIC_FAILURES_LEVEL_1_HINT_1=The challenge response already contains the real password in plaintext.
+CRYPTOGRAPHIC_FAILURES_LEVEL_1_HINT_2=No decoding or cracking is needed, just read it and submit it back through the 'password' parameter.
+CRYPTOGRAPHIC_FAILURES_LEVEL_1_PAYLOAD_DESCRIPTION=Plaintext password copied straight from the challenge response.
+CRYPTOGRAPHIC_FAILURES_LEVEL_1_PAYLOAD_VALUE=?password=<value shown in the CHALLENGE response>
+CRYPTOGRAPHIC_FAILURES_LEVEL_2_CHALLENGE=Decode the Base64 "encoded" value to recover the original password.
+CRYPTOGRAPHIC_FAILURES_LEVEL_2_HINT_1=Base64 is an encoding, not encryption, and is reversible with any standard decoder.
+CRYPTOGRAPHIC_FAILURES_LEVEL_2_HINT_2=Once decoded, submit the plaintext result through the 'password' parameter to confirm.
+CRYPTOGRAPHIC_FAILURES_LEVEL_2_PAYLOAD_DESCRIPTION=Base64-decoded password submitted back to verify the guess.
+CRYPTOGRAPHIC_FAILURES_LEVEL_2_PAYLOAD_VALUE=?password=<base64-decoded value>
+CRYPTOGRAPHIC_FAILURES_LEVEL_3_CHALLENGE=Reverse the Caesar cipher shift to recover the original password.
+CRYPTOGRAPHIC_FAILURES_LEVEL_3_HINT_1=A Caesar cipher only has 25 possible shifts, so it can be brute-forced entirely by hand.
+CRYPTOGRAPHIC_FAILURES_LEVEL_3_HINT_2=Try shifting each letter of the ciphertext backwards by a few positions and see when it turns into readable text.
+CRYPTOGRAPHIC_FAILURES_LEVEL_3_PAYLOAD_DESCRIPTION=Password recovered by reversing the Caesar cipher shift.
+CRYPTOGRAPHIC_FAILURES_LEVEL_3_PAYLOAD_VALUE=?password=<decrypted value>
+CRYPTOGRAPHIC_FAILURES_LEVEL_4_CHALLENGE=Figure out the custom obfuscation logic layered on top of encoding to recover the original password.
+CRYPTOGRAPHIC_FAILURES_LEVEL_4_HINT_1=The stored value looks encoded, look for a familiar transformation first.
+CRYPTOGRAPHIC_FAILURES_LEVEL_4_HINT_2=Try decoding it as Base64 first, then look for a second, simpler transformation applied underneath.
+CRYPTOGRAPHIC_FAILURES_LEVEL_4_PAYLOAD_DESCRIPTION=Password recovered by undoing the custom obfuscation logic.
+CRYPTOGRAPHIC_FAILURES_LEVEL_4_PAYLOAD_VALUE=?password=<recovered value>
+CRYPTOGRAPHIC_FAILURES_LEVEL_5_CHALLENGE=Crack the MD4 hash to recover the original password.
+CRYPTOGRAPHIC_FAILURES_LEVEL_5_HINT_1=MD4 has no salt and is trivially reversible with public rainbow tables.
+CRYPTOGRAPHIC_FAILURES_LEVEL_5_HINT_2=Paste the hash into an online hash-lookup tool, or crack it locally with hashcat (-m 900).
+CRYPTOGRAPHIC_FAILURES_LEVEL_5_PAYLOAD_DESCRIPTION=Password recovered by cracking the MD4 hash.
+CRYPTOGRAPHIC_FAILURES_LEVEL_5_PAYLOAD_VALUE=?password=<cracked value>
+CRYPTOGRAPHIC_FAILURES_LEVEL_6_CHALLENGE=Crack the MD5 hash to recover the original password.
+CRYPTOGRAPHIC_FAILURES_LEVEL_6_HINT_1=MD5 is unsalted and cryptographically broken, precomputed rainbow tables cover most common passwords.
+CRYPTOGRAPHIC_FAILURES_LEVEL_6_HINT_2=Paste the hash into an online hash-lookup tool, or crack it locally with hashcat (-m 0).
+CRYPTOGRAPHIC_FAILURES_LEVEL_6_PAYLOAD_DESCRIPTION=Password recovered by cracking the MD5 hash.
+CRYPTOGRAPHIC_FAILURES_LEVEL_6_PAYLOAD_VALUE=?password=<cracked value>
+CRYPTOGRAPHIC_FAILURES_LEVEL_7_CHALLENGE=Crack the SHA1 hash to recover the original password.
+CRYPTOGRAPHIC_FAILURES_LEVEL_7_HINT_1=SHA1 is deprecated for password storage and has no salt.
+CRYPTOGRAPHIC_FAILURES_LEVEL_7_HINT_2=Paste the hash into an online hash-lookup tool, or crack it locally with hashcat (-m 100).
+CRYPTOGRAPHIC_FAILURES_LEVEL_7_PAYLOAD_DESCRIPTION=Password recovered by cracking the SHA1 hash.
+CRYPTOGRAPHIC_FAILURES_LEVEL_7_PAYLOAD_VALUE=?password=<cracked value>
+CRYPTOGRAPHIC_FAILURES_LEVEL_8_CHALLENGE=Crack the LM hash to recover the original password.
+CRYPTOGRAPHIC_FAILURES_LEVEL_8_HINT_1=LM hashes are case-insensitive and split the password into two independent 7-character halves.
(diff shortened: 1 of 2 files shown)


src/main/resources/i18n/messages.properties
```diff
@@ -506,6 +506,56 @@ CRYPTOGRAPHIC_FAILURES_LM_HASHING=Password is hashed using LM algorithm which ha
...
 CRYPTOGRAPHIC_FAILURES_SECURE_BCRYPT=Password is hashed using Bcrypt with salt with a work factor (strength) of at least 4, making any type of brute-force or rainbow table attacks infeasible.
+CRYPTOGRAPHIC_FAILURES_LEVEL_1_CHALLENGE=Retrieve the plaintext password from the database, then submit it to the endpoint.
+CRYPTOGRAPHIC_FAILURES_LEVEL_1_HINT_1=The endpoint's own challenge text tells you to check the database when the password parameter is absent.
+CRYPTOGRAPHIC_FAILURES_LEVEL_1_HINT_2=No decoding or cracking is needed, just read the stored plaintext value and submit it back through the 'password' parameter.
+CRYPTOGRAPHIC_FAILURES_LEVEL_1_PAYLOAD_DESCRIPTION=Plaintext password read from the database and submitted to the endpoint.
+CRYPTOGRAPHIC_FAILURES_LEVEL_1_PAYLOAD_VALUE=?password=<value read from the database>
+CRYPTOGRAPHIC_FAILURES_LEVEL_2_CHALLENGE=Decode the Base64 "encoded" value to recover the original password.
```
preetkaran20, line 514: please don't tell that it is base64 encoded. let user figure it out. this can be the hint to the user.

src/main/resources/i18n/messages.properties
```diff
@@ -506,6 +506,56 @@ CRYPTOGRAPHIC_FAILURES_LM_HASHING=Password is hashed using LM algorithm which ha
...
+CRYPTOGRAPHIC_FAILURES_LEVEL_2_HINT_1=Base64 is an encoding, not encryption, and is reversible with any standard decoder.
+CRYPTOGRAPHIC_FAILURES_LEVEL_2_HINT_2=Once decoded, submit the plaintext result through the 'password' parameter to confirm.
+CRYPTOGRAPHIC_FAILURES_LEVEL_2_PAYLOAD_DESCRIPTION=Base64-decoded password submitted back to verify the guess.
+CRYPTOGRAPHIC_FAILURES_LEVEL_2_PAYLOAD_VALUE=?password=<base64-decoded value>
+CRYPTOGRAPHIC_FAILURES_LEVEL_3_CHALLENGE=Reverse the Caesar cipher shift to recover the original password.
```
preetkaran20, line 519: @pereiravp this level already has challenge text in UI. if you can mention the same exact same, it is better as we had put a lot of effort on it also then in future, we will remove the challenge text from the UI and only use challenge card.

preetkaran20 (commented):
can you please make the challenge text same as what is shown in UI as we had put a lot of effort on that in past so same text would be great. you can have hints to reach to the breaking it. 

Image

src/main/resources/i18n/messages.properties
pereiravp: removed the Base64 name from the challenge and payload text, only the hint mentions decoding now. pushed in bee086f.

src/main/resources/i18n/messages.properties
pereiravp: matched the level 1 challenge text to the exact wording already in the endpoint response. pushed in bee086f.

pereiravp:
good catch, that one applied beyond just level 1/2. matched all 10 levels' challenge text to the exact wording already in each endpoint's response (dropping only the dynamic hash/value and the surrounding chrome, since those are generated per request). pushed in bb496d3.

src/main/resources/i18n/messages.properties
preetkaran20: Can yoh please fix all other levels to match the text as well. We will then later on raised another PR to remove challenge text from UI and follow challenge card.

preetkaran20:
> good catch, that one applied beyond just level 1/2. matched all 10 levels' challenge text to the exact wording already in each endpoint's response (dropping only the dynamic hash/value and the surrounding chrome, since those are generated per request). pushed in [bb496d3](https://github.com/SasanLabs/VulnerableApp/commit/bb496d315920c5674f4273b379a8d7d9eea8ad23).

If you look at messages in UI it also shows the passwords too I think, can we do that ? Or do you think think we should not do it ?

src/main/resources/i18n/messages.properties
pereiravp: done, matched all 10 levels in bb496d3 (pushed right around the same time as this comment, see the message above).

preetkaran20:
> > good catch, that one applied beyond just level 1/2. matched all 10 levels' challenge text to the exact wording already in each endpoint's response (dropping only the dynamic hash/value and the surrounding chrome, since those are generated per request). pushed in [bb496d3](https://github.com/SasanLabs/VulnerableApp/commit/bb496d315920c5674f4273b379a8d7d9eea8ad23).
> 
> If you look at messages in UI it also shows the passwords too I think, can we do that ? Or do you think think we should not do it ?

I rethinked and it is ok to let user go and search database and find out the password. But payloads should tell extract password if attack vector payload is already doing that.

pereiravp:
makes sense, agreed on keeping level 1 as a database lookup. for the rest, reworded the payload descriptions to say the password is extracted from the value already shown in the challenge (decoded/cracked/etc depending on the level), instead of the vaguer "recovered" wording. pushed in fe33f9f.

src/main/resources/i18n/messages.properties
```diff
@@ -506,6 +506,56 @@ CRYPTOGRAPHIC_FAILURES_LM_HASHING=Password is hashed using LM algorithm which ha
 CRYPTOGRAPHIC_FAILURES_SHA256_HASHING=Password is hashed using SHA-256 without a salt, making it vulnerable to rainbow table and precomputed hash attacks. The user must crack the SHA-256 hash using online tools to find the original password.
 CRYPTOGRAPHIC_FAILURES_INSECURE_AES128=Password is encrypted with AES-128 but implemented with a weak key making it insecure to brute-force attacks. Encryption is a two-way function meaning that anyone with the key can decrypt the password. The user must crack the password using common password lists or key-recovery techniques.
 CRYPTOGRAPHIC_FAILURES_SECURE_BCRYPT=Password is hashed using Bcrypt with salt with a work factor (strength) of at least 4, making any type of brute-force or rainbow table attacks infeasible.
+CRYPTOGRAPHIC_FAILURES_LEVEL_1_CHALLENGE=The system stores passwords in plaintext. Check the database for the password to crack the challenge.
+CRYPTOGRAPHIC_FAILURES_LEVEL_1_HINT_1=The endpoint's own challenge text tells you to check the database when the password parameter is absent.
```
preetkaran20: Please don’t link to endpoints challenge text as we will remove it eventually.

src/main/resources/i18n/messages.properties
```diff
@@ -506,6 +506,56 @@ CRYPTOGRAPHIC_FAILURES_LM_HASHING=Password is hashed using LM algorithm which ha
...
+CRYPTOGRAPHIC_FAILURES_LEVEL_1_HINT_2=No decoding or cracking is needed, just read the stored plaintext value and submit it back through the 'password' parameter.
+CRYPTOGRAPHIC_FAILURES_LEVEL_1_PAYLOAD_DESCRIPTION=Plaintext password read from the database and submitted to the endpoint.
+CRYPTOGRAPHIC_FAILURES_LEVEL_1_PAYLOAD_VALUE=?password=<value read from the database>
+CRYPTOGRAPHIC_FAILURES_LEVEL_2_CHALLENGE=The system "encodes" passwords. Decode the stored value and enter the original password.
+CRYPTOGRAPHIC_FAILURES_LEVEL_2_HINT_1=This is an encoding, not encryption, and is reversible with any standard decoder once you identify the scheme.
+CRYPTOGRAPHIC_FAILURES_LEVEL_2_HINT_2=Once decoded, submit the plaintext result through the 'password' parameter to confirm.
```
preetkaran20, line 516: I think at this hint we should mention that it is base64 encoded.

src/main/resources/i18n/messages.properties
```diff
@@ -506,6 +506,56 @@ CRYPTOGRAPHIC_FAILURES_LM_HASHING=Password is hashed using LM algorithm which ha
...
+CRYPTOGRAPHIC_FAILURES_LEVEL_2_PAYLOAD_DESCRIPTION=Password extracted by decoding the value already shown in the challenge.
+CRYPTOGRAPHIC_FAILURES_LEVEL_2_PAYLOAD_VALUE=?password=<decoded value>
+CRYPTOGRAPHIC_FAILURES_LEVEL_3_CHALLENGE=A user's password is encrypted using an insecure cipher. Crack it and enter the original password.
+CRYPTOGRAPHIC_FAILURES_LEVEL_3_HINT_1=A Caesar cipher only has 25 possible shifts, so it can be brute-forced entirely by hand.
+CRYPTOGRAPHIC_FAILURES_LEVEL_3_HINT_2=Try shifting each letter of the ciphertext backwards by a few positions and see when it turns into readable text.
+CRYPTOGRAPHIC_FAILURES_LEVEL_3_PAYLOAD_DESCRIPTION=Password extracted by reversing the cipher shown in the challenge.
```
preetkaran20: Please don’t mention as shown in challenge ui as we will remove ui text.

src/main/resources/i18n/messages.properties
```diff
@@ -506,6 +506,56 @@ CRYPTOGRAPHIC_FAILURES_LM_HASHING=Password is hashed using LM algorithm which ha
...
+CRYPTOGRAPHIC_FAILURES_LEVEL_3_PAYLOAD_VALUE=?password=<decrypted value>
+CRYPTOGRAPHIC_FAILURES_LEVEL_4_CHALLENGE=A user's password is stored using custom logic. Crack it and enter the original password.
+CRYPTOGRAPHIC_FAILURES_LEVEL_4_HINT_1=The stored value looks encoded, look for a familiar transformation first.
+CRYPTOGRAPHIC_FAILURES_LEVEL_4_HINT_2=Try decoding it as Base64 first, then look for a second, simpler transformation applied underneath.
```
preetkaran20, line 526: Is this base 64 encoded ?

src/main/resources/i18n/messages.properties
```diff
@@ -506,6 +506,56 @@ CRYPTOGRAPHIC_FAILURES_LM_HASHING=Password is hashed using LM algorithm which ha
...
+CRYPTOGRAPHIC_FAILURES_LEVEL_4_PAYLOAD_DESCRIPTION=Password extracted by undoing the custom logic applied to the value shown in the challenge.
+CRYPTOGRAPHIC_FAILURES_LEVEL_4_PAYLOAD_VALUE=?password=<recovered value>
+CRYPTOGRAPHIC_FAILURES_LEVEL_5_CHALLENGE=A user's password is stored as an MD4 hash. Crack it and enter the original password.
```
preetkaran20: Please don’t mention the algo here and let user find it out. Let the algo be the hint

src/main/resources/i18n/messages.properties
```diff
@@ -506,6 +506,56 @@ CRYPTOGRAPHIC_FAILURES_LM_HASHING=Password is hashed using LM algorithm which ha
...
+CRYPTOGRAPHIC_FAILURES_LEVEL_5_HINT_1=MD4 has no salt and is trivially reversible with public rainbow tables.
+CRYPTOGRAPHIC_FAILURES_LEVEL_5_HINT_2=Paste the hash into an online hash-lookup tool, or crack it locally with hashcat (-m 900).
+CRYPTOGRAPHIC_FAILURES_LEVEL_5_PAYLOAD_DESCRIPTION=Password extracted by cracking the MD4 hash shown in the challenge.
+CRYPTOGRAPHIC_FAILURES_LEVEL_5_PAYLOAD_VALUE=?password=<cracked value>
+CRYPTOGRAPHIC_FAILURES_LEVEL_6_CHALLENGE=A user's password is stored as an MD5 hash. Crack it and enter the original password.
+CRYPTOGRAPHIC_FAILURES_LEVEL_6_HINT_1=MD5 is unsalted and cryptographically broken, precomputed rainbow tables cover most common passwords.
+CRYPTOGRAPHIC_FAILURES_LEVEL_6_HINT_2=Paste the hash into an online hash-lookup tool, or crack it locally with hashcat (-m 0).
+CRYPTOGRAPHIC_FAILURES_LEVEL_6_PAYLOAD_DESCRIPTION=Password extracted by cracking the MD5 hash shown in the challenge.
+CRYPTOGRAPHIC_FAILURES_LEVEL_6_PAYLOAD_VALUE=?password=<cracked value>
+CRYPTOGRAPHIC_FAILURES_LEVEL_7_CHALLENGE=A user's password is stored as a SHA1 hash. Crack it and enter the original password.
```
preetkaran20: Same here.

pereiravp:
pushed in cc78b31:
- moved the algorithm names (MD4, MD5, SHA1, LM, AES-128) out of the CHALLENGE text and into the hints for levels 5-10, matching the pattern you asked for on level 5/7
- level 2's hint now names Base64 directly instead of hedging with "identify the scheme"
- dropped the "shown in the challenge"/"endpoint's own challenge text" references from level 1's hint and all payload descriptions, so nothing points at UI text you're planning to remove
- double checked level 4: yes, it's Base64 encoded on top of the reversal, hint 2 already says that correctly so left it as is

preetkaran20 (approved):

PR #781, opened 2026-08-30
feat: add Insecure Deserialization vulnerability module

### LINKED ISSUE #528: Insecure Deserialization Vulnerability
Insecure deserialization occurs when untrusted data is deserialized by an application without proper validation or safeguards. Attackers can manipulate serialized objects to execute arbitrary code, escalate privileges, or bypass authentication. This vulnerability is particularly dangerous in Java applications that rely on `Serializable` objects or similar mechanisms.

This issue proposes adding an **Insecure Deserialization vulnerability module** to demonstrate unsafe deserialization patterns and their secure mitigations.

### Level 1 – Basic Unsafe Deserialization
The application deserializes user-controlled input without validation, allowing simple object manipulation or unexpected behavior.

### Level 2 – Arbitrary Code Execution
User-supplied serialized objects can trigger the execution of arbitrary code on the server during deserialization.

### Level 3 – Authentication Bypass / Privilege Escalation
Manipulated serialized objects can bypass authentication or elevate user privileges.

### Level 4 – Secure Implementation
Demonstrates secure deserialization practices such as:
- Avoiding deserialization of untrusted data
- Using validation or allowlists for acceptable classes
- Applying serialization filters (e.g., `ObjectInputFilter` in Java)
- Implementing integrity checks and safe parsing techniques

## Goal
Help users and security scanners understand how insecure deserialization can be exploited and teach secure deserialization practices in Java applications.

pereiravp:
## Summary
Closes #528.

New vulnerability module (CWE-502: Deserialization of Untrusted Data), built with Challenge Mode annotations from the start, following the module's proposed levels:

- **LEVEL_1** – Basic Unsafe Deserialization: the endpoint deserializes any Base64 payload with no type or integrity checks. Craft your own `UserPreferences` and the server rebuilds it as-is.
- **LEVEL_2** – Arbitrary Code Execution: a custom `readObject()` hook on `AuditLogGadget` runs attacker-controlled logic the instant the bytes are deserialized, before the app ever sees the result. No external gadget-chain library is used — the effect (appending to an in-memory audit trail) is observable without granting real code execution, but demonstrates exactly the mechanism real gadget chains (Commons Collections, etc.) exploit.
- **LEVEL_3** – Auth Bypass / Privilege Escalation: the endpoint trusts an `isAdmin` flag read straight off a deserialized `SessionToken`, no signature. Forge one with `isAdmin=true` and get admin access.
- **LEVEL_4** (secure) – The same `SessionToken` flow, but the payload is HMAC-signed server-side and the signature is verified *before* a single byte is deserialized; a class allowlist via `ObjectInputFilter` adds defense in depth. The exact forgery that works on LEVEL_3 is rejected here.

## Verification
Every level was exercised end-to-end against a running instance (`./gradlew bootRun`), not just unit tests: forged real Java-serialized payloads with a standalone `javac`/`java` script (matching class names/fields) and hit each endpoint over HTTP with curl, confirming levels 1-3 are exploitable as designed and level 4 correctly rejects the level-3 forgery via signature mismatch.

## Test plan
- [x] `./gradlew compileJava`
- [x] New unit tests (`InsecureDeserializationVulnerabilityTest`, 10 cases) pass
- [x] Manually exploited all 4 levels over HTTP against a running instance

## Summary by CodeRabbit

* **New Features**
  * Added a four-level insecure deserialization vulnerability challenge.
  * Added interactive preference, activity, and login flows with cookie-based payloads.
  * Added signed, class-restricted payload verification for the secure challenge level.
  * Added challenge pages, controls, result feedback, and styling for Levels 1–3.
  * Added support for detecting deserialized preference changes, audit activity, and administrator access.

* **Bug Fixes**
  * Improved handling of missing, malformed, tampered, oversized, and unsupported payloads.

* **Tests**
  * Added comprehensive coverage for challenge flows, validation, cookies, security checks, and failure scenarios.

### DIFF
--- a/src/main/java/org/sasanlabs/service/vulnerability/deserialization/AuditLogGadget.java
+++ b/src/main/java/org/sasanlabs/service/vulnerability/deserialization/AuditLogGadget.java
@@ -1,0 +1,44 @@
+package org.sasanlabs.service.vulnerability.deserialization;
+
+import java.io.IOException;
+import java.io.ObjectInputStream;
+import java.io.Serializable;
+import java.util.List;
+import java.util.concurrent.CopyOnWriteArrayList;
+
+/**
+ * VULNERABILITY (CWE-502): this class exists purely to prove that {@code readObject} runs
+ * attacker-controlled logic the instant the bytes are deserialized, before the caller gets any
+ * chance to inspect or validate the resulting object. In a real gadget chain this hook would be a
+ * class already on the classpath (e.g. Commons Collections) chosen by the attacker to reach a
+ * dangerous sink such as process execution; here it appends to an in-memory audit trail so the
+ * effect is observable without granting real code execution.
+ */
+public class AuditLogGadget implements Serializable {
+
+    private static final long serialVersionUID = 1L;
+
+    private static final List<String> AUDIT_TRAIL = new CopyOnWriteArrayList<>();
+
+    private String note;
+
+    public AuditLogGadget() {}
+
+    public AuditLogGadget(String note) {
+        this.note = note;
+    }
+
+    public String getNote() {
+        return note;
+    }
+
+    public static List<String> getAuditTrail() {
+        return AUDIT_TRAIL;
+    }
+
+    private void readObject(ObjectInputStream in) throws IOException, ClassNotFoundException {
+        in.defaultReadObject();
+        // This runs during deserialization itself, before the application ever sees the object.
+        AUDIT_TRAIL.add("CODE EXECUTED DURING DESERIALIZATION: " + note);
+    }
+}
--- a/src/main/java/org/sasanlabs/service/vulnerability/deserialization/AuditLogGadget.java
+++ b/src/main/java/org/sasanlabs/service/vulnerability/deserialization/AuditLogGadget.java
@@ -18,6 +18,8 @@ public class AuditLogGadget implements Serializable {
 
     private static final long serialVersionUID = 1L;
 
+    // Capped so repeated challenge submissions cannot grow this list without bound.
+    private static final int MAX_AUDIT_TRAIL_SIZE = 50;
     private static final List<String> AUDIT_TRAIL = new CopyOnWriteArrayList<>();
 
     private String note;
@@ -39,6 +41,9 @@ public class AuditLogGadget implements Serializable {
     private void readObject(ObjectInputStream in) throws IOException, ClassNotFoundException {
         in.defaultReadObject();
         // This runs during deserialization itself, before the application ever sees the object.
+        while (AUDIT_TRAIL.size() >= MAX_AUDIT_TRAIL_SIZE) {
+            AUDIT_TRAIL.remove(0);
+        }
         AUDIT_TRAIL.add("CODE EXECUTED DURING DESERIALIZATION: " + note);
     }
 }
--- a/src/main/java/org/sasanlabs/service/vulnerability/deserialization/DeserializationUtils.java
+++ b/src/main/java/org/sasanlabs/service/vulnerability/deserialization/DeserializationUtils.java
@@ -1,0 +1,137 @@
+package org.sasanlabs.service.vulnerability.deserialization;
+
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.InvalidClassException;
+import java.io.ObjectInputFilter;
+import java.io.ObjectInputStream;
+import java.io.ObjectOutputStream;
+import java.io.Serializable;
+import java.security.InvalidKeyException;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import java.security.SecureRandom;
+import java.util.Arrays;
+import java.util.Base64;
+import javax.crypto.Mac;
+import javax.crypto.spec.SecretKeySpec;
+import org.sasanlabs.internal.utility.exception.DeserializationException;
+
+/** Serialization helpers backing the Insecure Deserialization vulnerability levels. */
+public class DeserializationUtils {
+
+    private DeserializationUtils() {}
+
+    private static final String HMAC_ALGORITHM = "HmacSHA256";
+    private static final int SIGNATURE_LENGTH_BYTES = 32;
+
+    // Used only by LEVEL_4's secure implementation to sign/verify tokens it issues itself.
+    // Levels 1-3 never look at a secret at all, which is exactly the vulnerability.
+    private static final byte[] SERVER_SECRET = new byte[32];
+
+    static {
+        new SecureRandom().nextBytes(SERVER_SECRET);
+    }
+
+    public static String serializeToBase64(Serializable object) throws DeserializationException {
+        try (ByteArrayOutputStream byteStream = new ByteArrayOutputStream();
+                ObjectOutputStream objectStream = new ObjectOutputStream(byteStream)) {
+            objectStream.writeObject(object);
+            objectStream.flush();
+            return Base64.getEncoder().encodeToString(byteStream.toByteArray());
+        } catch (IOException e) {
+            throw new DeserializationException("Failed to serialize object", e);
+        }
+    }
+
+    /**
+     * INSECURE: deserializes attacker-controlled bytes with no signature check and no class
+     * allowlist. Any {@link Serializable} class on the classpath can be instantiated this way, and
+     * its {@code readObject} runs before the caller gets any chance to validate the result.
+     */
+    public static Object deserializeUnsafe(String base64Payload) throws DeserializationException {
+        byte[] raw = decodeBase64(base64Payload);
+        try (ObjectInputStream objectStream =
+                new ObjectInputStream(new ByteArrayInputStream(raw))) {
+            return objectStream.readObject();
+        } catch (IOException | ClassNotFoundException e) {
+            throw new DeserializationException("Failed to deserialize payload", e);
+        }
+    }
+
+    /**
+     * SECURE: signs the serialized bytes of {@code object} with an HMAC known only to the server,
+     * so that {@link #verifyAndDeserialize} can reject anything it did not issue before ever
+     * touching {@link ObjectInputStream}.
+     */
+    public static String signAndSerialize(Serializable object) throws DeserializationException {
+        try (ByteArrayOutputStream byteStream = new ByteArrayOutputStream();
+                ObjectOutputStream objectStream = new ObjectOutputStream(byteStream)) {
+            objectStream.writeObject(object);
+            objectStream.flush();
+            byte[] serializedBytes = byteStream.toByteArray();
+            byte[] signature = hmac(serializedBytes);
+
+            ByteArrayOutputStream combined = new ByteArrayOutputStream();
+            combined.write(signature);
+            combined.write(serializedBytes);
+            return Base64.getEncoder().encodeToString(combined.toByteArray());
+        } catch (IOException e) {
+            throw new DeserializationException("Failed to serialize object", e);
+        }
+    }
+
+    /**
+     * SECURE: verifies the HMAC signature over the payload's bytes before deserializing a single
+     * byte of it, then restricts the resolvable class to {@code allowedClass} as defense in depth.
+     * A tampered signature is rejected without ever reaching {@link ObjectInputStream}.
+     */
+    public static Object verifyAndDeserialize(String base64Payload, Class<?> allowedClass)
+            throws DeserializationException {
+        byte[] combined = decodeBase64(base64Payload);
+        if (combined.length <= SIGNATURE_LENGTH_BYTES) {
+            throw new DeserializationException("Payload too short to contain a signature");
+        }
+        byte[] signature = Arrays.copyOfRange(combined, 0, SIGNATURE_LENGTH_BYTES);
+        byte[] serializedBytes =
+                Arrays.copyOfRange(combined, SIGNATURE_LENGTH_BYTES, combined.length);
+
+        byte[] expectedSignature = hmac(serializedBytes);
+        if (!MessageDigest.isEqual(signature, expectedSignature)) {
+            throw new DeserializationException(
+                    "Signature mismatch: payload was tampered with or was never issued by this"
+                            + " server");
+        }
+
+        ObjectInputFilter filter =
+                ObjectInputFilter.Config.createFilter(allowedClass.getName() + ";!*");
+        try (ObjectInputStream objectStream =
+                new ObjectInputStream(new ByteArrayInputStream(serializedBytes))) {
+            objectStream.setObjectInputFilter(filter);
+            return objectStream.readObject();
+        } catch (InvalidClassException e) {
+            throw new DeserializationException("Rejected class not on the allowlist", e);
+        } catch (IOException | ClassNotFoundException e) {
+            throw new DeserializationException("Failed to deserialize payload", e);
+        }
+    }
+
+    private static byte[] hmac(byte[] data) throws DeserializationException {
+        try {
+            Mac mac = Mac.getInstance(HMAC_ALGORITHM);
+            mac.init(new SecretKeySpec(SERVER_SECRET, HMAC_ALGORITHM));
+            return mac.doFinal(data);
+        } catch (NoSuchAlgorithmException | InvalidKeyException e) {
+            throw new DeserializationException("Failed to compute HMAC signature", e);
+        }
+    }
+
+    private static byte[] decodeBase64(String base64Payload) throws DeserializationException {
+        try {
+            return Base64.getDecoder().decode(base64Payload);
+        } catch (IllegalArgumentException e) {
+            throw new DeserializationException("Payload is not valid Base64", e);
+        }
+    }
+}
--- a/src/main/java/org/sasanlabs/service/vulnerability/deserialization/DeserializationUtils.java
+++ b/src/main/java/org/sasanlabs/service/vulnerability/deserialization/DeserializationUtils.java
@@ -93,6 +93,9 @@ public class DeserializationUtils {
      */
     public static Object verifyAndDeserialize(String base64Payload, Class<?> allowedClass)
             throws DeserializationException {
+        if (base64Payload == null) {
+            throw new DeserializationException("Payload is required");
+        }
         if (base64Payload.length() > MAX_ENCODED_PAYLOAD_LENGTH) {
             throw new DeserializationException("Payload exceeds the maximum allowed size");
         }
--- a/src/main/java/org/sasanlabs/service/vulnerability/deserialization/DeserializationUtils.java
+++ b/src/main/java/org/sasanlabs/service/vulnerability/deserialization/DeserializationUtils.java
@@ -26,6 +26,10 @@ public class DeserializationUtils {
     private static final String HMAC_ALGORITHM = "HmacSHA256";
     private static final int SIGNATURE_LENGTH_BYTES = 32;
 
+    // Rejects grossly oversized input before it is even Base64-decoded, so LEVEL_4's secure path
+    // can't be used to burn memory/CPU on a payload that was always going to fail verification.
+    private static final int MAX_ENCODED_PAYLOAD_LENGTH = 16_384;
+
     // Used only by LEVEL_4's secure implementation to sign/verify tokens it issues itself.
     // Levels 1-3 never look at a secret at all, which is exactly the vulnerability.
     private static final byte[] SERVER_SECRET = new byte[32];
@@ -89,6 +93,9 @@ public class DeserializationUtils {
      */
     public static Object verifyAndDeserialize(String base64Payload, Class<?> allowedClass)
             throws DeserializationException {
+        if (base64Payload.length() > MAX_ENCODED_PAYLOAD_LENGTH) {
+            throw new DeserializationException("Payload exceeds the maximum allowed size");
+        }
         byte[] combined = decodeBase64(base64Payload);
         if (combined.length <= SIGNATURE_LENGTH_BYTES) {
             throw new DeserializationException("Payload too short to contain a signature");
@@ -105,7 +112,10 @@ public class DeserializationUtils {
         }
 
         ObjectInputFilter filter =
-                ObjectInputFilter.Config.createFilter(allowedClass.getName() + ";!*");
+                ObjectInputFilter.Config.createFilter(
+                        "maxdepth=5;maxarray=1000;maxrefs=100;maxbytes=8192;"
+                                + allowedClass.getName()
+                                + ";!*");
         try (ObjectInputStream objectStream =
                 new ObjectInputStream(new ByteArrayInputStream(serializedBytes))) {
             objectStream.setObjectInputFilter(filter);
--- a/src/main/java/org/sasanlabs/service/vulnerability/deserialization/InsecureDeserializationLoginController.java
+++ b/src/main/java/org/sasanlabs/service/vulnerability/deserialization/InsecureDeserializationLoginController.java
@@ -1,0 +1,132 @@
+package org.sasanlabs.service.vulnerability.deserialization;
+
+import javax.servlet.http.Cookie;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import org.sasanlabs.internal.utility.LevelConstants;
+import org.sasanlabs.internal.utility.exception.DeserializationException;
+import org.sasanlabs.service.vulnerability.bean.GenericVulnerabilityResponseBean;
+import org.springframework.context.annotation.Profile;
+import org.springframework.http.HttpStatus;
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.bind.annotation.PostMapping;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RequestParam;
+import org.springframework.web.bind.annotation.RestController;
+
+/**
+ * Issues the cookies that {@link InsecureDeserializationVulnerability} deserializes. Each endpoint
+ * here only ever writes values a normal user could pick from the UI (a theme, a section name, a
+ * username), then hands the browser back an HttpOnly cookie carrying the resulting object
+ * serialized and Base64-encoded. The vulnerable levels never check where that cookie came from,
+ * which is what lets a value edited in the browser's dev tools reach them unexamined.
+ *
+ * @author KSASAN preetkaran20@gmail.com
+ */
+@Profile("public")
+@RestController
+@RequestMapping("InsecureDeserialization")
+public class InsecureDeserializationLoginController {
+
+    @PostMapping(LevelConstants.LEVEL_1)
+    public ResponseEntity<GenericVulnerabilityResponseBean<String>> saveLevel1(
+            @RequestParam String theme,
+            @RequestParam(required = false, defaultValue = "false") boolean notificationsEnabled,
+            HttpServletRequest request,
+            HttpServletResponse response) {
+        if (!"light".equalsIgnoreCase(theme) && !"dark".equalsIgnoreCase(theme)) {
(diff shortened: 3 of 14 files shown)


src/main/java/org/sasanlabs/service/vulnerability/deserialization/InsecureDeserializationVulnerability.java
```diff
@@ -0,0 +1,285 @@
...
+                                + " — Craft your own serialized UserPreferences with"
+                                + " theme=\"HACKED\" and submit it via the 'payload' parameter!");
+            } catch (DeserializationException e) {
+                return ok(false, "Failed to build the sample payload: " + e.getMessage());
+            }
+        }
+
+        try {
+            Object result = DeserializationUtils.deserializeUnsafe(payload);
+            if (result instanceof UserPreferences preferences) {
```
preetkaran20, line 75: I think it would be worth thinking a bit more about the concrete use case here before implementation. I think is use case is importnatn that UI has options and that stores the serialized data but user should not be putting serialized data as that is really difficult for scanners and students to learn on why serialization is even needed.

preetkaran20: @pereiravp

pereiravp: added a real theme/notifications form for level 1 that calls the endpoint to get an actual serialized preferences token, same as the app would issue for a real feature. the exploit is now tampering with that real token (change theme to a value the dropdown never offers) instead of hand-building a UserPreferences payload from nothing. pushed in cdf1e20, let me know if this matches what you had in mind or if you want a different concrete use case.

src/main/java/org/sasanlabs/service/vulnerability/deserialization/AuditLogGadget.java
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    public static List<String> getAuditTrail() {
+        return AUDIT_TRAIL;
+    }
+
+    private void readObject(ObjectInputStream in) throws IOException, ClassNotFoundException {
+        in.defaultReadObject();
+        // This runs during deserialization itself, before the application ever sees the object.
+        AUDIT_TRAIL.add("CODE EXECUTED DURING DESERIALIZATION: " + note);
```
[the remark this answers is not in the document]
pereiravp: capped AUDIT_TRAIL at 50 entries (evicts oldest), pushed in 756d357.

src/main/java/org/sasanlabs/service/vulnerability/deserialization/InsecureDeserializationVulnerability.java
```diff
@@ -0,0 +1,311 @@
...
+                                + " — Try tampering with it (e.g. re-encoding a forged"
+                                + " isAdmin=true token like in LEVEL_3) and submit it via the"
+                                + " 'payload' parameter.");
+            } catch (DeserializationException e) {
+                return ok(false, "Failed to build the sample payload: " + e.getMessage());
+            }
+        }
+
+        try {
+            Object result = DeserializationUtils.verifyAndDeserialize(payload, SessionToken.class);
```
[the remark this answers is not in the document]
pereiravp: added a max encoded length check before decoding and configured the ObjectInputFilter with maxdepth/maxarray/maxrefs/maxbytes alongside the class allowlist. pushed in 756d357.

src/main/resources/i18n/messages.properties
```diff
@@ -1358,3 +1358,28 @@ JWT_LEVEL16_CHALLENGE2_HINT1=HttpOnly stops JavaScript from reading the cookie,
...
+INSECURE_DESERIALIZATION_LEVEL_4_SECURE=This secure level verifies an HMAC signature over the serialized bytes before deserializing a single one of them, and restricts the resolvable class to an allowlist, so a tampered or substituted payload is rejected before it can run any code.
+INSECURE_DESERIALIZATION_LEVEL_1_CHALLENGE=Get your own preferences token, then get the server to deserialize it back with a theme value the form never offered.
+INSECURE_DESERIALIZATION_LEVEL_1_HINT_1=The token is just a Base64-encoded, Java-serialized UserPreferences object with no signature, and the endpoint deserializes any one you send as-is.
+INSECURE_DESERIALIZATION_LEVEL_1_HINT_2=Take the token the app gave you, decode it, edit the theme field to something the dropdown never offers, re-encode it, and submit it through the 'payload' parameter.
+INSECURE_DESERIALIZATION_LEVEL_1_PAYLOAD_DESCRIPTION=Your own preferences token, decoded and re-encoded with theme changed to a value the UI never offers.
+INSECURE_DESERIALIZATION_LEVEL_1_PAYLOAD_VALUE=?payload=<base64 of your token with theme edited to "HACKED">
```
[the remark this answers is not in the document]
pereiravp: escaped the angle brackets in the payload placeholder values. pushed in 756d357.

src/main/resources/static/templates/InsecureDeserialization/LEVEL_1/InsecureDeserialization.js
```diff
@@ -0,0 +1,80 @@
...
+        url + "?" + params.toString(),
+        true
+      );
+    });
+}
+
+function appendResponseCallback(data) {
+  let resultDiv = document.getElementById("result");
+  if (data.isValid) {
+    resultDiv.innerHTML = "<strong>Result:</strong> " + data.content;
```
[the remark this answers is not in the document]
pereiravp: good catch, the theme/note fields can carry arbitrary attacker text through a tampered payload. rewrote the level 1 JS to build the DOM instead of using innerHTML on server content. pushed in 756d357.

src/main/java/org/sasanlabs/service/vulnerability/deserialization/DeserializationUtils.java
```diff
@@ -0,0 +1,147 @@
...
+    /**
+     * SECURE: verifies the HMAC signature over the payload's bytes before deserializing a single
+     * byte of it, then restricts the resolvable class to {@code allowedClass} as defense in depth.
+     * A tampered signature is rejected without ever reaching {@link ObjectInputStream}.
+     */
+    public static Object verifyAndDeserialize(String base64Payload, Class<?> allowedClass)
+            throws DeserializationException {
+        if (base64Payload.length() > MAX_ENCODED_PAYLOAD_LENGTH) {
+            throw new DeserializationException("Payload exceeds the maximum allowed size");
+        }
```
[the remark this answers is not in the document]
pereiravp: right, my own size check introduced that. added the null guard before it. pushed in 9d22832.

src/main/java/org/sasanlabs/service/vulnerability/deserialization/InsecureDeserializationVulnerability.java
preetkaran20: @pereiravp I think there is some misunderstanding. My thought is that  we should make the serialized object part of a normal application flow for all the levels.

My thought is to use a cookie-based flow consistently across the deserialization levels.

For example, we could have a normal Preferences UI:
* User selects a theme (e.g. Dark or Choosing larger font size)
* User clicks Save
* React sends the normal JSON preference to the backend
* Backend creates/stores the serialized Preferences object in a cookie
* On subsequent requests, the browser automatically sends the cookie back
* Backend deserializes the cookie and uses the resulting preferences when generating the response
* React reads the response and changes UI behaviour like making the theme as dark etc.

The important part is that the React UI does not need to read or manipulate the cookie. The browser handles that automatically. The cookie can also be HttpOnly, so the serialized value is not exposed to application JavaScript.

So we will not give any option to put Base64 encoded preferences in UI. It is part of cookie and user can manipulate it in the browser dev tools.

pereiravp: that's clearer, thanks. rebuilt it that way: added InsecureDeserializationLoginController (same shape as IDORLoginController) with a normal save/login action per level (save preferences, visit a section, log in with a username) that sets an HttpOnly cookie carrying the serialized object. The vulnerable GET endpoints now read that cookie via @CookieValue and deserialize it, same as before. No Base64 box in the UI any more for any level, the exploit is editing the cookie in dev tools. Level 4 uses the same shape but with the signed cookie. Verified end to end against the packaged jar (save -> cookie set -> tamper -> resubmit -> accepted). pushed in 4a58060.

src/main/resources/i18n/messages.properties
```diff
@@ -1358,3 +1358,28 @@ JWT_LEVEL16_CHALLENGE2_HINT1=HttpOnly stops JavaScript from reading the cookie,
...
+INSECURE_DESERIALIZATION_LEVEL_4_SECURE=This secure level verifies an HMAC signature over the serialized bytes before deserializing a single one of them, and restricts the resolvable class to an allowlist, so a tampered or substituted payload is rejected before it can run any code.
+INSECURE_DESERIALIZATION_LEVEL_1_CHALLENGE=Save your preferences, then edit the resulting cookie so the theme becomes a value the form never offers.
+INSECURE_DESERIALIZATION_LEVEL_1_HINT_1=The saved cookie is just a Base64-encoded, Java-serialized UserPreferences object with no signature.
+INSECURE_DESERIALIZATION_LEVEL_1_HINT_2=Open your browser's dev tools, decode the cookie, change theme to something the dropdown never offers, re-encode it, and save it back.
+INSECURE_DESERIALIZATION_LEVEL_1_PAYLOAD_DESCRIPTION=Your saved preferences cookie, decoded and re-encoded with theme changed to a value the UI never offers.
+INSECURE_DESERIALIZATION_LEVEL_1_PAYLOAD_VALUE=Cookie: preferences_level1=&lt;base64 of your cookie with theme edited to "HACKED"&gt;
```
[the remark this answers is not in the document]
pereiravp: good catch, fixed in bfc119e.

preetkaran20 (changes requested):
I looked at the PR and there are following concerns:
1. While saving preferences, I got 400s and at few other places. can you please have a look.
2. I am not very convinced with Challenge text in the UI. Challenge text is not needed in the UI as Challenge card should drive it
3. I am also not very convinced with Level1 challenge text. May be something like `The application stores your preferences in a serialized cookie. The UI only allows a limited set of preference values. Can you modify the stored preferences so that the application accepts a value that is not available through the UI?` 
4. I think the response from the level should be simple saved preference. we should not return ```json {
    "content": "Deserialized fine (theme='light'), but that is not the target value. Edit the cookie so theme becomes \"HACKED\".",
    "isValid": false
}``` as this is not about gamification of VulnerableApp. The goal is to be a simple application with vulnerabilities.
5. I am also confused by the fact that Level 3 and Level 4 introduce a login/password flow.
I don't think authentication is necessary to demonstrate insecure deserialization. Introducing login/passwords makes the flow more complicated and, in my opinion, distracts from the actual vulnerability.
Could we instead make the application revolve around one simple Settings/Preferences object that is serialized and stored client-side?
For example:
Settings
├── role = ANONYMOUS
├── theme = LIGHT
├── language = EN
└── notifications = true
The application could have one simple Settings UI. The server creates a legitimate Settings object, serializes it, and provides the serialized representation to the client. The client can then modify that representation, and the server later deserializes and trusts it.
This gives us a very simple and consistent mental model for the first few levels:
Settings object
      ↓
   serialize
      ↓
client receives serialized object
      ↓
client modifies it
      ↓
server deserializes it
      ↓
server trusts the resulting object
6. I also think that we can reorder level 2 as last level considering it teaches different thing from all the other levels.
7. Also UI should be simple, all the preferences/settings are saved by single button as that is simple and consistent. Ensure each level has its own cookies so one level doesn't interfere with another level's cookie.

preetkaran20:
@pereiravp Overall I liked the approach, a few things related to UI and levels. please let me know if any concerns.

src/main/resources/static/templates/InsecureDeserialization/LEVEL_1/InsecureDeserialization.js
```diff
@@ -0,0 +1,102 @@
...
+  let level = getCurrentVulnerabilityLevel();
+  let level1Controls = document.getElementById("level1Controls");
+  let level2Controls = document.getElementById("level2Controls");
+  let level34Controls = document.getElementById("level34Controls");
+
+  level1Controls.classList.add("hide-component");
+  level2Controls.classList.add("hide-component");
+  level34Controls.classList.add("hide-component");
+
+  if (level === "LEVEL_1") {
```
preetkaran20: please avoid these checks, you can use different templates for each level to make it easier to maintain.

src/main/resources/i18n/messages.properties
```diff
@@ -1358,3 +1358,28 @@ JWT_LEVEL16_CHALLENGE2_HINT1=HttpOnly stops JavaScript from reading the cookie,
...
+INSECURE_DESERIALIZATION_LEVEL_1_PAYLOAD_DESCRIPTION=Your saved preferences cookie, decoded and re-encoded with theme changed to a value the UI never offers.
+INSECURE_DESERIALIZATION_LEVEL_1_PAYLOAD_VALUE=Cookie: preferences_level1=<base64 of your cookie with theme edited to "HACKED">
+INSECURE_DESERIALIZATION_LEVEL_2_CHALLENGE=Visit a section above to get an activity cookie, then edit it so its class still runs code you did not intend during deserialization.
+INSECURE_DESERIALIZATION_LEVEL_2_HINT_1=The cookie's class has a custom readObject() method, code inside it runs during deserialization itself, not after.
+INSECURE_DESERIALIZATION_LEVEL_2_HINT_2=Decode the activity cookie, change its note to a value of your choosing, re-encode it, and check the audit trail the response reveals.
+INSECURE_DESERIALIZATION_LEVEL_2_PAYLOAD_DESCRIPTION=Your saved activity cookie, decoded and re-encoded with a note of your choosing.
+INSECURE_DESERIALIZATION_LEVEL_2_PAYLOAD_VALUE=Cookie: activity_level2=<base64 of your cookie with an attacker-chosen note>
```
preetkaran20: it would be best if you give the actual payload for users as it is easier for them to look how application behaviour changes with the payload.

src/main/java/org/sasanlabs/service/vulnerability/deserialization/SessionToken.java
```diff
@@ -0,0 +1,31 @@
+package org.sasanlabs.service.vulnerability.deserialization;
+
+import java.io.Serializable;
+
+/**
+ * Stand-in for a session/auth token carried as a serialized Java object. Levels 3 and 4 use it to
+ * demonstrate that trusting a deserialized object's fields for an authorization decision is only as
+ * safe as the integrity check guarding the bytes it came from.
+ */
+public class SessionToken implements Serializable {
```
preetkaran20: I think this is not needed. One single object should suffice all the levels.

src/main/java/org/sasanlabs/service/vulnerability/deserialization/InsecureDeserializationVulnerability.java
```diff
@@ -0,0 +1,257 @@
...
+            Object result =
+                    DeserializationUtils.verifyAndDeserialize(sessionCookie, SessionToken.class);
+            SessionToken token = (SessionToken) result;
+            return ok(
+                    true,
+                    "Signature valid, token accepted. User '"
+                            + token.getUsername()
+                            + "', isAdmin="
+                            + token.isAdmin()
+                            + ". Because the signature is verified before any bytes are"
```
preetkaran20: No need to mention this information to the user. Lets build it like a simple application with preference functionality.

src/main/java/org/sasanlabs/service/vulnerability/deserialization/InsecureDeserializationVulnerability.java
```diff
@@ -0,0 +1,257 @@
...
+    @VulnerableAppRequestMapping(
+            value = LevelConstants.LEVEL_4,
+            variant = Variant.SECURE,
+            htmlTemplate = "LEVEL_1/InsecureDeserialization")
+    public ResponseEntity<GenericVulnerabilityResponseBean<String>> getSecurePayloadLevel4(
+            @CookieValue(value = SESSION_COOKIE_LEVEL_4, required = false) String sessionCookie) {
+        if (sessionCookie == null || sessionCookie.isEmpty()) {
+            return ok(
+                    true,
+                    "SECURE CHALLENGE: Log in above to get a session cookie. It is HMAC-signed"
```
preetkaran20: This text is not needed as challenge card is enough for telling the goal of the vulnerability.

src/main/java/org/sasanlabs/service/vulnerability/deserialization/InsecureDeserializationVulnerability.java
```diff
@@ -0,0 +1,257 @@
...
+                            + " isAdmin value that cookie deserializes to, with no signature.");
+        }
+
+        try {
+            Object result = DeserializationUtils.deserializeUnsafe(sessionCookie);
+            if (result instanceof SessionToken token) {
+                if (token.isAdmin()) {
+                    return ok(
+                            true,
+                            "Correct! A session cookie for user '"
```
preetkaran20: This looks like gamification. please ignore this way. let user see changes in UI or something else instead of us telling that they have passed the level.

preetkaran20: Our use case if Scanners and we want vulnerable app to be as close to a genuine production application as we can.

src/main/resources/static/templates/InsecureDeserialization/LEVEL_1/InsecureDeserialization.js
pereiravp: done, each level has its own template and js now. level 1 has the preferences form, level 2 the section picker, level 3 and 4 the login one. no more level checks in the js. pushed in 3da31b4

src/main/resources/i18n/messages.properties
pereiravp: done, levels 1 to 3 now have real base64 cookie values you can paste straight into dev tools and see the app behaviour change. tested each one against the running app

src/main/java/org/sasanlabs/service/vulnerability/deserialization/SessionToken.java
pereiravp: removed. UserPreferences is the only object now, level 1 uses theme/notifications and levels 3 and 4 use username/admin on it

src/main/java/org/sasanlabs/service/vulnerability/deserialization/InsecureDeserializationVulnerability.java
pereiravp: removed, level 4 now just shows the signed in profile like level 3 does and returns the plain error when the signature fails

src/main/java/org/sasanlabs/service/vulnerability/deserialization/InsecureDeserializationVulnerability.java
pereiravp: removed, it just tells the user to log in now. the challenge card already carries the goal

src/main/java/org/sasanlabs/service/vulnerability/deserialization/InsecureDeserializationVulnerability.java
pereiravp: agreed, dropped all the correct/access granted commentary. responses now show what the app did with the cookie (theme applied, audit log entry, admin panel data for an admin). the ui only uses the isValid flag it already had to mark the level

src/main/java/org/sasanlabs/service/vulnerability/deserialization/AuditLogGadget.java
```diff
@@ -0,0 +1,49 @@
...
+        return AUDIT_TRAIL;
+    }
+
+    private void readObject(ObjectInputStream in) throws IOException, ClassNotFoundException {
+        in.defaultReadObject();
+        // This runs during deserialization itself, before the application ever sees the object.
+        while (AUDIT_TRAIL.size() >= MAX_AUDIT_TRAIL_SIZE) {
+            AUDIT_TRAIL.remove(0);
+        }
+        AUDIT_TRAIL.add(note);
```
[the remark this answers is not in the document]
pereiravp: good catch, wrapped the eviction and the add in one synchronized block so concurrent deserializations can't pass the size check together or hit remove(0) on an empty list. pushed in 33aae3f

src/main/java/org/sasanlabs/service/vulnerability/deserialization/DeserializationUtils.java
```diff
@@ -0,0 +1,150 @@
...
+    private static final byte[] SERVER_SECRET = new byte[32];
+
+    static {
+        new SecureRandom().nextBytes(SERVER_SECRET);
+    }
+
+    public static String serializeToBase64(Serializable object) throws DeserializationException {
+        try (ByteArrayOutputStream byteStream = new ByteArrayOutputStream();
+                ObjectOutputStream objectStream = new ObjectOutputStream(byteStream)) {
+            objectStream.writeObject(object);
```
preetkaran20, line 44: due to java object serialization, it is very difficult to modify the base64 encoded text. it doesn't convert into json printable object. I think we can use Jackson json library or some other library that can create the printable json object and base64 encode it so that user can modify it.

preetkaran20: Image

src/main/java/org/sasanlabs/service/vulnerability/deserialization/InsecureDeserializationLoginController.java
```diff
@@ -0,0 +1,134 @@
...
+ * username), then hands the browser back an HttpOnly cookie carrying the resulting object
+ * serialized and Base64-encoded. The vulnerable levels never check where that cookie came from,
+ * which is what lets a value edited in the browser's dev tools reach them unexamined.
+ *
+ * @author KSASAN preetkaran20@gmail.com
+ */
+@Profile("public")
+@RestController
+@RequestMapping("InsecureDeserialization")
+public class InsecureDeserializationLoginController {
```
preetkaran20, line 29: why is this named as LoginController?

src/main/java/org/sasanlabs/service/vulnerability/deserialization/InsecureDeserializationLoginController.java
```diff
@@ -0,0 +1,134 @@
...
+                    token,
+                    request,
+                    response);
+            return ok("Preferences saved.");
+        } catch (DeserializationException e) {
+            return badRequest("Failed to save preferences: " + e.getMessage());
+        }
+    }
+
+    @PostMapping(LevelConstants.LEVEL_2)
```
preetkaran20, line 56: make this as last level i.e. level 4 as this is different from all the other levels.

preetkaran20: so level 3 becomes level 2 and Lebel 4 becomes level 3

src/main/java/org/sasanlabs/service/vulnerability/deserialization/InsecureDeserializationVulnerability.java
```diff
@@ -0,0 +1,204 @@
...
+                @ChallengeCard.Hint(order = 2, text = "INSECURE_DESERIALIZATION_LEVEL_3_HINT_2")
+            },
+            payload =
+                    @ChallengeCard.Payload(
+                            description = "INSECURE_DESERIALIZATION_LEVEL_3_PAYLOAD_DESCRIPTION",
+                            value = "INSECURE_DESERIALIZATION_LEVEL_3_PAYLOAD_VALUE"))
+    @VulnerableAppRequestMapping(
+            value = LevelConstants.LEVEL_3,
+            htmlTemplate = "LEVEL_3/InsecureDeserialization")
+    public ResponseEntity<GenericVulnerabilityResponseBean<String>> getVulnerablePayloadLevel3(
```
preetkaran20, line 148: @pereiravp I’d suggest removing the username field as well. Since there’s no real authentication here, asking for a username doesn’t add much value.

Instead, we can simulate an authenticated user (e.g., Adam) and show “Welcome Adam” with their current role/tasks. The challenge can then be to modify the client-side session/cookie so Adam is interpreted as an admin and gets access to admin-only tasks.

This would make the insecure deserialization scenario much more intuitive and keep the focus on the vulnerability rather than the login/authentication flow.

It is ok that the UI has nothing actionable and scanners and users understand that cookies can be manipulated.

src/main/java/org/sasanlabs/service/vulnerability/deserialization/InsecureDeserializationVulnerability.java
```diff
@@ -0,0 +1,204 @@
...
+        } catch (DeserializationException e) {
+            return ok(false, "Deserialization failed: " + e.getMessage());
+        }
+    }
+
+    // Level 3: Authentication bypass / privilege escalation via a forged token (CWE-502)
+    @AttackVector(
+            vulnerabilityExposed = VulnerabilityType.INSECURE_DESERIALIZATION,
+            description = "INSECURE_DESERIALIZATION_LEVEL_3_PRIVILEGE_ESCALATION")
+    @ChallengeCard(
```
preetkaran20, line 135: This is not just deserialization issue, this is also an authorization flow. please add the attack vector for that too.

src/main/java/org/sasanlabs/service/vulnerability/deserialization/InsecureDeserializationVulnerability.java
```diff
@@ -0,0 +1,204 @@
...
+                    false,
+                    "Deserialized an unexpected type: "
+                            + (result == null ? "null" : result.getClass().getName()));
+        } catch (DeserializationException e) {
+            return ok(false, "Deserialization failed: " + e.getMessage());
+        }
+    }
+
+    // Level 2: Arbitrary code execution during deserialization (CWE-502)
+    @AttackVector(
```
preetkaran20, line 93: this could be a RCE too. please add another attack vector annotation if you think that is right,

src/main/java/org/sasanlabs/service/vulnerability/deserialization/InsecureDeserializationVulnerability.java
```diff
@@ -0,0 +1,204 @@
...
+        if (preferencesCookie == null || preferencesCookie.isEmpty()) {
+            return ok(false, "Save your preferences above to continue.");
+        }
+
+        try {
+            Object result = DeserializationUtils.deserializeUnsafe(preferencesCookie);
+            if (result instanceof UserPreferences preferences) {
+                boolean unsupportedTheme =
+                        preferences.getTheme() != null
+                                && !"light".equalsIgnoreCase(preferences.getTheme())
```
preetkaran20, line 73: please create string constants

src/main/java/org/sasanlabs/service/vulnerability/deserialization/InsecureDeserializationVulnerability.java
```diff
@@ -0,0 +1,204 @@
...
+            htmlTemplate = "LEVEL_1/InsecureDeserialization")
+    public ResponseEntity<GenericVulnerabilityResponseBean<String>> getVulnerablePayloadLevel1(
+            @CookieValue(value = PREFERENCES_COOKIE_LEVEL_1, required = false)
+                    String preferencesCookie) {
+        if (preferencesCookie == null || preferencesCookie.isEmpty()) {
+            return ok(false, "Save your preferences above to continue.");
+        }
+
+        try {
+            Object result = DeserializationUtils.deserializeUnsafe(preferencesCookie);
```
preetkaran20, line 69: using Jackson will help against type checks like instance of too.

Issue #403, opened 2022-10-01
Add secured 2-3 secured implementations for OpenRedirect


preetkaran20:
**Is your feature request related to a problem? Please describe.**
As main target for Vulnerable App is scanners so we need to find out false positives by Scanners such that they can improve on there scan rules. So in order to find out the false positives, we need to add more secured implementations.

**Describe the solution you'd like**
Add 2-3 more functions in https://github.com/SasanLabs/VulnerableApp/tree/master/src/main/java/org/sasanlabs/service/vulnerability/openRedirect which are secure.

vianbas:
@preetkaran20 thanks for the pointer on #728. Before claiming this one I went looking for what had already been done, and the answer changes the shape of the issue, so I would rather check with you than start. I would like to pick this up once we agree on what it should cover.

**This issue has been partly satisfied without anything linking back to it**

When this was filed in October 2022 the class had 8 levels. `LEVEL_8` already validated against `WHITELISTED_URLS` back then; it simply carried no `Variant.SECURE` marker. Two PRs have touched that picture since:

| PR | Merged | What it did | New secure implementation? |
|---|---|---|---|
| #589 | 2026-04-19 | Added `variant = Variant.SECURE` to `LEVEL_8`, a one-line annotation change with no behaviour change. Its body says `Closes #312` | No |
| #619 | 2026-05-18 | Added `LEVEL_9`, `LEVEL_10` and `LEVEL_11`, the last a secure redirect | Yes, one |

So of the "2-3" this issue asks for, one has landed. Neither PR references this issue, and its timeline carries only label changes and one cross-reference, so from the outside it still reads as untouched.

**Both secure levels use the same check**

```java
LEVEL_8   getURLRedirectionResponseEntity(urlToRedirect, WHITELISTED_URLS::contains)
LEVEL_11  if (WHITELISTED_URLS.contains(urlToRedirect))
```

`WHITELISTED_URLS` is `{"/", "/VulnerableApp/"}`, so both are an exact string match against a two-element set. There are two secure levels but one defensive technique, written twice.

They also disagree on what a rejected redirect returns. `LEVEL_8` answers `200 OK` with an empty body, by way of the shared `getURLRedirectionResponseEntity` helper; `LEVEL_11` answers `403` with `Redirect blocked: untrusted redirect target`. I do not know which you intend as the reference behaviour, so I have not assumed either.

**What I think would serve the stated goal**

[...]

Issue #255, opened 2020-11-16
Java jar running is not working because of Path Api issue with Jar


preetkaran20:
Error logs:
```
org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'UnrestrictedFileUpload' defined in URL [jar:file:/Users/ksinghsasa/Learning%20Stuff/Web%20Application%20Proxy%20Tools/Payloads/SasanLabs/VulnerableApp/build/libs/VulnerableApp-1.0.0.jar!/BOOT-INF/classes!/org/sasanlabs/service/vulnerability/fileupload/UnrestrictedFileUpload.class]: Instantiation of bean failed; nested exception is org.springframework.beans.BeanInstantiationException: Failed to instantiate [org.sasanlabs.service.vulnerability.fileupload.UnrestrictedFileUpload]: Constructor threw exception; nested exception is java.nio.file.FileSystemNotFoundException
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.instantiateBean(AbstractAutowireCapableBeanFactory.java:1320)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.createBeanInstance(AbstractAutowireCapableBeanFactory.java:1214)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.doCreateBean(AbstractAutowireCapableBeanFactory.java:557)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.createBean(AbstractAutowireCapableBeanFactory.java:517)
	at org.springframework.beans.factory.support.AbstractBeanFactory.lambda$doGetBean$0(AbstractBeanFactory.java:323)
	at org.springframework.beans.factory.support.DefaultSingletonBeanRegistry.getSingleton(DefaultSingletonBeanRegistry.java:222)
	at org.springframework.beans.factory.support.AbstractBeanFactory.doGetBean(AbstractBeanFactory.java:321)
	at org.springframework.beans.factory.support.AbstractBeanFactory.getBean(AbstractBeanFactory.java:202)
	at org.springframework.beans.factory.support.DefaultListableBeanFactory.preInstantiateSingletons(DefaultListableBeanFactory.java:879)
	at org.springframework.context.support.AbstractApplicationContext.finishBeanFactoryInitialization(AbstractApplicationContext.java:878)
	at org.springframework.context.support.AbstractApplicationContext.refresh(AbstractApplicationContext.java:550)
[...]

PR #783, opened 2026-09-02
feat: add two secure FileUpload implementations

### LINKED ISSUE #401: Add secured 2-3 secured implementations for FileUpload
As main target for Vulnerable App is scanners so we need to find out false positives by Scanners such that they can improve on there scan rules. So in order to find out the false positives, we need to add more secured implementations.

**Describe the solution you'd like**
Add 2-3 more functions in https://github.com/SasanLabs/VulnerableApp/tree/master/src/main/java/org/sasanlabs/service/vulnerability/fileupload which are secure.

pereiravp:
Closes #401.

FileUpload currently has one secure level (level 10), which checks that the file name ends in `.png` or `.jpeg`. That check never looks past the name: a file named `shell.php.png` still clears it, since it ends with `.png`.

**Level 11** reads the file's own bytes and checks them against the actual PNG and JPEG binary signatures. The stored name is built only from the extension the server detects that way, never from anything the client sent, so a renamed non-image file is rejected and a real image can never end up stored under an attacker-chosen extension like `.html`.

**Level 12** carries the same signature check and server-generated name as level 11, plus an upload size cap of 512 KB. That cap has to sit below Spring Boot's own multipart default of 1 MB, or a request over it would never reach the method at all, it would be rejected by the framework with a 500 before any of this code ran.

Both are additive. Nothing in levels 1 to 10 changed.

How I checked it: unit tests cover a real PNG being accepted on level 11, a PHP payload renamed to `.png` being rejected, a real PNG uploaded with a name of `../../shell.html` still ending up stored as `.png`, level 12 accepting a small JPEG and rejecting one over its own cap, and a non-image upload being rejected on both levels. I also ran the built jar and posted the same cases with curl: the `.html`-named PNG landed on disk as plain `.png`, and a 700 KB JPEG, over level 12's cap but under Spring's own 1 MB limit, came back rejected by this code rather than by the framework.

```sh
./gradlew clean build
java -jar build/libs/VulnerableApp-1.0.0.jar &
B=http://localhost:9090/VulnerableApp/UnrestrictedFileUpload
do sleep 2; done

# level 11: real PNG bytes, name ends in .html, stored as .png anyway
curl -s -F "file=@real.png;filename=../../shell.html;type=image/png" "$B/LEVEL_11"

# level 12: ~700 KB JPEG, over the 512 KB cap but under Spring's 1 MB limit
curl -s -F "file=@mid.jpeg;filename=mid.jpeg;type=image/jpeg" "$B/LEVEL_12"
```

`./gradlew clean build` passes with 454 tests, 0 failures, on master `c833c66`.

Update: an earlier revision of level 11 in this PR kept the client-supplied extension for the stored name even after validating the signature, which CodeRabbit correctly flagged as still allowing a signature-valid PNG to be stored as `.html`. Fixed in the second commit, which is also where level 12's distinct control (the size cap) was added, since after that fix the two levels would otherwise have ended up doing the same thing.

## Summary by CodeRabbit

* **New Features**
  * Added PNG and JPEG validation based on file content for applicable upload levels.
  * Added server-generated filenames with detected image extensions, preventing client-provided names and path traversal from affecting stored files.
  * Added a 512 KB upload-size limit for the stricter validation level.
  * Invalid image content and oversized uploads are rejected.

* **Documentation**
  * Updated upload-validation messages to describe signature checks, generated filenames, and the size limit.

### DIFF
--- a/src/main/java/org/sasanlabs/service/vulnerability/fileupload/UnrestrictedFileUpload.java
+++ b/src/main/java/org/sasanlabs/service/vulnerability/fileupload/UnrestrictedFileUpload.java
@@ -11,6 +11,7 @@ import java.nio.file.Paths;
 import java.nio.file.StandardCopyOption;
 import java.util.Date;
 import java.util.Random;
+import java.util.UUID;
 import java.util.function.Supplier;
 import java.util.regex.Pattern;
 import org.apache.commons.text.StringEscapeUtils;
@@ -625,7 +626,10 @@ public class UnrestrictedFileUpload {
     // with .png, regardless of what bytes are actually inside it. This level reads the file's
     // own binary signature instead, and the stored name is built from that detected type rather
     // than from anything the client sent, so a renamed non-image file is rejected and a real
-    // image can never be stored under an attacker-chosen extension like .html.
+    // image can never be stored under an attacker-chosen extension like .html. The name itself
+    // is a UUID rather than RANDOM.nextInt(): genericFileUploadUtility overwrites on a name
+    // collision, and a 32-bit int is small enough that one becomes likely after tens of
+    // thousands of uploads, which a reference secure level should not carry.
     @AttackVector(
             vulnerabilityExposed = {
                 VulnerabilityType.UNRESTRICTED_FILE_UPLOAD,
@@ -643,7 +647,7 @@ public class UnrestrictedFileUpload {
         byte[] content = file.getBytes();
         String extension = detectImageExtension(content);
         Supplier<Boolean> validator = () -> extension != null;
-        String storedFileName = RANDOM.nextInt() + (extension == null ? "" : extension);
+        String storedFileName = UUID.randomUUID() + (extension == null ? "" : extension);
         return genericFileUploadUtility(root, storedFileName, validator, file, true, false);
     }
 
@@ -670,7 +674,7 @@ public class UnrestrictedFileUpload {
         String extension = detectImageExtension(content);
         Supplier<Boolean> validator =
                 () -> extension != null && content.length <= MAX_UPLOAD_SIZE_BYTES;
-        String storedFileName = RANDOM.nextInt() + (extension == null ? "" : extension);
+        String storedFileName = UUID.randomUUID() + (extension == null ? "" : extension);
         return genericFileUploadUtility(root, storedFileName, validator, file, true, false);
     }
 }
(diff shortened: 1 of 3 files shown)


vianbas:
Reviewed this by running it rather than reading it, since the claims are behavioural. Disclosure first: I have #793 open, which also adds a secure level, and I commented on #403 about the same family of issues. Different module, nothing competing, but worth saying.

**Everything the description claims, holds.** I built the branch at `23c67cc`, ran the jar, and uploaded against it. A positive control first, so a row of rejections means something: `LEVEL_1` accepted a raw PHP file and stored it under the name I sent.

| Upload | Level 11 | Level 12 |
|---|---|---|
| genuine 1x1 PNG | accepted, stored as a UUID with `.png` | accepted |
| PHP renamed `.png` | rejected | rejected |
| genuine PNG sent as `../../shell.html` | accepted, stored as a UUID with `.png` | accepted |
| empty file | rejected | rejected |
| JPEG about 400 KB | accepted | accepted |
| JPEG about 700 KB | accepted, no cap at this level | rejected |

The traversal case is the one I most wanted to break and could not. The stored name never carries anything the client sent, and `genericFileUploadUtility` uses only the name you pass it, so there is nothing left of the original to exploit.

**One thing worth knowing, and I do not think it is a defect.** The check is a signature check, which is exactly what the description says it is, so this is a limit rather than a gap between claim and behaviour. Four files passed it that are not decodable images: an eight byte PNG signature with nothing after it, the three byte JPEG marker with nothing after it, a PNG signature followed by PHP source, and the 700 KB file I made by putting the JPEG marker in front of half a megabyte of the letter A. Only the genuine PNG decodes through `ImageIO`.

Nothing about that is exploitable through the name, because the name is yours rather than the caller's. Whether the stored bytes matter at all depends on how the upload is later served, and that is the part I could not test: running from the jar puts uploads in a temp directory, which is #255, so I never got a served file back to inspect its response headers. Someone running from an exploded build could answer that quickly.

If you did want to close it, `ImageIO.read` returning non null is a one line addition, and it would also reject the 700 KB case at level 11 as a side effect. Your call whether that is worth it for a level whose point is the signature check.

Build is green on my side too: 471 tests, 0 failures.

pereiravp:
thanks for actually running it and the traversal probe, appreciated. agree it's not a defect given the level is specifically about the signature check, not full decodability, so leaving it as is for now. good catch on #255 too, makes sense that's what blocked you from checking response headers.

preetkaran20 (approved):
LGTM. Thanks a lot for the PR. Merging it now.

PR #784, opened 2026-09-02
feat: add two secure PathTraversal implementations

### LINKED ISSUE #404: Add secured 2-3 secured implementations for PathTraversal Vulnerability
As main target for Vulnerable App is scanners so we need to find out false positives by Scanners such that they can improve on there scan rules. So in order to find out the false positives, we need to add more secured implementations.

**Describe the solution you'd like**
Add 2-3 more functions in https://github.com/SasanLabs/VulnerableApp/tree/master/src/main/java/org/sasanlabs/service/vulnerability/pathTraversal which are secure.

pereiravp:
Closes #404.

PathTraversal has one secure level (level 13), which checks the whole `fileName` parameter against two hardcoded strings. On the issue thread, the maintainer asked for a mix of strategies rather than a second name list, and named rejection of traversal attempts specifically as the other kind worth having.

**Level 14** resolves the requested path instead of matching a name against a list. `fileName` gets appended to the resource's base directory, run through `Paths.normalize()` so every `..` segment collapses, and the request only proceeds if the normalized result still starts inside that base directory. A NUL byte makes the path itself invalid rather than truncating it, so `InvalidPathException` is caught and treated as a rejection.

**Level 15** takes a third angle: the whole parameter has to match `^[A-Za-z0-9_-]+\.[A-Za-z0-9]{2,5}$`. There is no path separator or NUL byte in that character set, so nothing in it can be shaped into a traversal sequence.

Neither depends on knowing every legitimate file name ahead of time the way level 13 does. Both are additive; nothing in levels 1 to 13 changed.

How I checked it: unit tests cover both levels rejecting a deep `/etc/passwd` traversal, a `../secret.json` escape one directory up, and a NUL-byte payload, plus level 14 accepting `OwaspAppInfo.json` even though it is never referenced anywhere in that level's code. I also ran the built jar and hit both endpoints directly to see the actual responses.

```sh
./gradlew clean build
java -jar build/libs/VulnerableApp-1.0.0.jar &
B=http://localhost:9090/VulnerableApp/PathTraversal
do sleep 2; done

# level 14: accepted even though never hardcoded
curl -s "$B/LEVEL_14?fileName=OwaspAppInfo.json"

# level 14: one directory up, rejected
curl -s "$B/LEVEL_14?fileName=../secret.json"

# level 15: traversal sequence, rejected
curl -s "$B/LEVEL_15?fileName=../../../../../../etc/passwd"
```

`./gradlew clean build` passes with 502 tests, 0 failures, on master `c833c66`.

## Summary by CodeRabbit

- **New Features**
  - Added Path Traversal challenge levels 14 and 15.
  - Level 14 validates normalized paths and blocks requests that leave the permitted resource directory.
  - Level 15 accepts only filenames matching a restricted alphanumeric, underscore, and hyphen pattern with a single extension.
  - Added localized challenge descriptions, hints, and payload information for both levels.

- **Bug Fixes**
  - Invalid, null, traversal, and null-byte path inputs are now rejected safely.

### DIFF
--- a/src/main/java/org/sasanlabs/service/vulnerability/pathTraversal/PathTraversalVulnerability.java
+++ b/src/main/java/org/sasanlabs/service/vulnerability/pathTraversal/PathTraversalVulnerability.java
@@ -52,9 +52,10 @@ public class PathTraversalVulnerability {
 
     private static final String BASE_RESOURCE_PATH = "/scripts/PathTraversal/";
 
-    // Letters, digits, dots, underscores and hyphens only, with no path separator of any kind
-    // in the character set at all, so there is nothing here for a traversal sequence to be
-    // built out of, encoded or not.
+    // Allows file names made of letters, digits, underscores or hyphens, followed by a single
+    // dot and a 2-5 character alphanumeric extension. No path separator or dot sequence is in
+    // that character set, so there is nothing here for a traversal segment to be built out of,
+    // encoded or not.
     private static final Pattern SAFE_FILE_NAME_PATTERN =
             Pattern.compile("^[A-Za-z0-9_-]+\\.[A-Za-z0-9]{2,5}$");
 
--- a/src/main/java/org/sasanlabs/service/vulnerability/pathTraversal/PathTraversalVulnerability.java
+++ b/src/main/java/org/sasanlabs/service/vulnerability/pathTraversal/PathTraversalVulnerability.java
@@ -6,10 +6,13 @@ import java.io.BufferedReader;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
+import java.nio.file.InvalidPathException;
+import java.nio.file.Paths;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 import java.util.function.Supplier;
+import java.util.regex.Pattern;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.sasanlabs.internal.utility.LevelConstants;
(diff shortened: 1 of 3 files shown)


src/main/java/org/sasanlabs/service/vulnerability/pathTraversal/PathTraversalVulnerability.java
```diff
@@ -46,11 +49,19 @@ public class PathTraversalVulnerability {
 
     private static final String URL_PARAM_KEY = "fileName";
 
+    private static final String BASE_RESOURCE_PATH = "/scripts/PathTraversal/";
+
+    // Letters, digits, dots, underscores and hyphens only, with no path separator of any kind
+    // in the character set at all, so there is nothing here for a traversal sequence to be
+    // built out of, encoded or not.
+    private static final Pattern SAFE_FILE_NAME_PATTERN =
```
preetkaran20, line 59: please add something like this allows filenames containing letters, digits, _ or -, followed by a single . and a 2–5 character alphanumeric extension.

src/main/java/org/sasanlabs/service/vulnerability/pathTraversal/PathTraversalVulnerability.java
```diff
@@ -499,4 +510,72 @@ public ResponseEntity<GenericVulnerabilityResponseBean<String>> getVulnerablePay
...
+                () -> {
+                    if (fileName == null) {
+                        return false;
+                    }
+                    try {
+                        String normalizedBase =
+                                Paths.get(BASE_RESOURCE_PATH).normalize().toString();
+                        String normalizedTarget =
+                                Paths.get(BASE_RESOURCE_PATH + fileName).normalize().toString();
+                        return normalizedTarget.startsWith(normalizedBase);
```
preetkaran20, line 548: can you please compare paths instead of strings like targetPath.startsWith(basePath).

preetkaran20 (commented):
Thank you for the PR. can you please fix a small cosmetic change

src/main/java/org/sasanlabs/service/vulnerability/pathTraversal/PathTraversalVulnerability.java
pereiravp: fixed, comparing Path objects with startsWith now instead of the normalized strings.

src/main/java/org/sasanlabs/service/vulnerability/pathTraversal/PathTraversalVulnerability.java
pereiravp: done, reworded to state what it actually allows.

preetkaran20 (approved):

PR #788, opened 2026-09-07
feat: add a challenge card to the secure cache poisoning level

vianbas:
Cache poisoning is the only challenge mode module whose secure level carries no card. This adds one.

@preetkaran20 asked on #764 whether a card on a secure level is shown in the UI and said there is no harm in adding one if it is. It is: on `XXEVulnerability` LEVEL_5 the card renders in full, with its text and both the Reveal hint and Show Payload controls. That question is what this PR follows up.

**Built to match the module rather than copied from my closed PR.** The goal is short and close ended in the style @pereiravp landed in #779, and the card carries two hints and a payload to match the other seven cards in the class and the ten carded secure levels elsewhere in the tree. The card asks the reader to confirm rather than to attack, which is the same pattern as `XXE_LEVEL5_CHALLENGE` and `PATH_TRAVERSAL_LEVEL13_CHALLENGE`.

Ten Java lines and five keys, in the default bundle only, which is what all nine challenge mode PRs merged so far have done.

**Verified against a running instance, not just built.** Each claim the card makes was checked, with a positive control so a passing result means something:

| Claim in the card | Result |
|---|---|
| the response is never stored | six consecutive requests to level 5 all answer `MISS` |
| control, so `MISS` is not vacuous | level 1 answers `HIT` on five of six |
| the forwarded host is ignored | `X-Forwarded-Host: attacker-controlled.example` never reaches the asset address |
| the asset host comes from configuration | the configured `static.vulnerableapp.local` appears instead |
| the banner comes back escaped | an `img` payload returns as `&lt;img`, not as an executable tag |

`./gradlew spotlessCheck build` passes: 471 tests, 0 failures, 2 skipped, and `MessageBundleDuplicateKeyTest` passes with the new keys.

The wording is a teaching decision rather than a coding one, so please rewrite anything pitched wrong. The keys and structure would not change.

## Summary by CodeRabbit

- **Documentation**
  - Added two hints to the Level 5 cache-poisoning challenge.
  - Added a concrete banner XSS payload example to clarify how to verify that no payload survives.
  - Clarified that responses are never cached and that the asset host is determined by configuration rather than the request.

### DIFF
--- a/src/main/resources/i18n/messages.properties
+++ b/src/main/resources/i18n/messages.properties
@@ -749,8 +749,10 @@ CACHE_POISONING_LEVEL_4_HINT_2=Set the demo_user cookie to a specific value to p
 CACHE_POISONING_LEVEL_4_PAYLOAD_DESCRIPTION=Cookie value used to poison the shared, publicly-cached personalized response.
 CACHE_POISONING_LEVEL_4_PAYLOAD_VALUE=Cookie: demo_user=admin
 CACHE_POISONING_LEVEL_5_CHALLENGE=Confirm there's no way to make this level serve your content to another visitor - it never stores the response and ignores the forwarded host.
-CACHE_POISONING_LEVEL_5_PAYLOAD_DESCRIPTION=
-CACHE_POISONING_LEVEL_5_PAYLOAD_VALUE=
+CACHE_POISONING_LEVEL_5_HINT_1=Every response reports the cache lookup next to the key. On the earlier levels it turns to a hit once the entry exists; watch whether it ever does here.
+CACHE_POISONING_LEVEL_5_HINT_2=Replay each earlier payload in turn. The banner comes back escaped, the asset host comes from configuration rather than the request, and the response is never stored.
+CACHE_POISONING_LEVEL_5_PAYLOAD_DESCRIPTION=Any payload from the earlier levels, replayed here to confirm that none of them survives.
+CACHE_POISONING_LEVEL_5_PAYLOAD_VALUE=?banner=<img src=x onerror=alert(document.domain)>
 
 #IDOR ChallengeCard Labels
 IDOR_VULNERABILITY_LEVEL_1_CHALLENGE=Can you bypass the authorization check and access another user's profile?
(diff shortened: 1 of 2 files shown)


src/main/resources/i18n/messages.properties
```diff
@@ -748,6 +748,11 @@ CACHE_POISONING_LEVEL_4_HINT_1=The response is built from the demo_user cookie,
 CACHE_POISONING_LEVEL_4_HINT_2=Set the demo_user cookie to a specific value to poison the route, then request the same route again with a different (or no) cookie and check whether you still get the first user's data.
 CACHE_POISONING_LEVEL_4_PAYLOAD_DESCRIPTION=Cookie value used to poison the shared, publicly-cached personalized response.
 CACHE_POISONING_LEVEL_4_PAYLOAD_VALUE=Cookie: demo_user=admin
+CACHE_POISONING_LEVEL_5_CHALLENGE=Confirm there's no way to make this level serve your content to another visitor - it never stores the response and ignores the forwarded host.
+CACHE_POISONING_LEVEL_5_HINT_1=Every response reports the cache lookup next to the key. On the earlier levels it turns to a hit once the entry exists; watch whether it ever does here.
```
preetkaran20: I think hints and payload are not needed for secure level considering there isn't anything user can do with hints or payloads.

vianbas: Done, hints dropped. Two things you may want to know before deciding whether this is the shape you want.

`hints` is optional on the annotation and now renders as an empty list, so that part is clean. `payload` is not optional, it has no default, so the card still has to name a description and a value. I pointed both at empty values, and `getPayload` falls back to its own "Payload is not applicable." string. That reads sensibly for a secure level, but it is the code choosing the wording rather than the card saying nothing.

The other thing is that this makes the card the odd one out. On master today 38 levels are `Variant.SECURE` and 10 of them carry a card, in `XXEVulnerability`, `PathTraversalVulnerability`, `XSSWithHtmlTagInjection`, `XSSInImgTagAttribute` and `PersistentXSSInHTMLTagVulnerability`. Every one of those 10 has two hints and a payload. So after this change cache poisoning level 5 is the only carded secure level without them.

I have no strong view on which is right, and I can see the argument that a level with nothing to exploit has nothing to hint at. If you want the convention to change, the other 10 are where it would show. If you would rather this one match them instead, say so and I will put the two hints back.

Build is green either way: 471 tests, 0 failures, 2 skipped, spotless clean.

preetkaran20: oh is it. then please add the hints back. Thanks for letting me know.

vianbas: Hints are back, along with the payload text, so the card now matches the other ten carded secure levels rather than being the exception. Reverted rather than rewritten, so the wording is exactly what you saw first.

`./gradlew spotlessCheck build` passes: 471 tests, 0 failures, 2 skipped, and the duplicate key check is green. Read back over HTTP, level 5 serves one card with its two hints, the payload description and the payload value.

The branch is ten commits behind master now. I tested the merge and it is clean, so I have left it alone rather than rebasing for its own sake. Say the word if you would rather it were current.

preetkaran20 (approved):

preetkaran20:
Lgtm. @vianbas thanks a lot for the PR. Merging it now.

PR #791, opened 2026-09-09
fix: restore nine i18n keys broken by line continuations (#782)

### LINKED ISSUE #782: Nine keys exist only in a translation bundle, so three translations are never served and four values stop mid-sentence
**Describe the bug**

A key that exists only in a translation bundle is never read. Every lookup starts from a key name written in the Java code, and those names are resolved against `messages.properties`. If a translation file spells a key differently, Spring finds nothing there and falls back to the default bundle, so the translated text sits in the repository and never reaches a user.

There are 9 such keys, in 4 of the 11 bundles, measured on `8a8d049`:

| File | Keys | Line | Orphaned key |
|---|---:|---:|---|
| `messages_hi.properties` | 97 | 12 | `BILITY` |
| | | 56 | `किसी` |
| | | 58 | `फ़िशिंग` |
| | | 60 | `3xx` |
| | | 126 | `XML` |
| `messages_es.properties` | 96 | 60 | `<br/><br/>` |
| | | 240 | `Links` |
| `messages_zh_TW.properties` | 93 | 12 | `BILITY` |
| `messages_bn_IN.properties` | 95 | 19 | `PERSISTENT_XSS_HTML_TAG_URL_PARAM_DIRECTLY_INJECTED_IN_DIV_TAG_CASE_INSENSITIVEIF_TAGS_ARE_PRESENT_BEFORE_NULL_BYTE` |

The other 7 bundles have none. `messages.properties` holds 938 keys.

They arise two ways, and both are visible over HTTP.

**One: a key name that lost its beginning.**

`messages_zh_TW.properties:11` and the line under it read:

```properties
# XSS based Injections
BILITY=<the translated XSS description>
```

`messages_hi.properties:12` is the same, under the same section comment written in Hindi. In both files the line above is a comment, not a continuation, so `XSS_VULNERA` is simply absent. `XSS_VULNERABILITY` is present in 9 of the 11 bundles and missing from exactly these two. It is the `descriptionLabel` of `XSSWithHtmlTagInjection`, `XSSInImgTagAttribute` and `PersistentXSSInHTMLTagVulnerability`, so it is what three cards display.

`messages_bn_IN.properties:19` is the same shape in the middle of a name: it lacks the `REPLACING_IMG_AND_INPUT_TAG_` segment that `messages.properties:33` has.

**Two: a continuation that lost its backslash.**

The remaining 6 are the tail of a multi-line value. When the trailing `\` is missing, the value ends early and the rest of the sentence is read as a new key. `XXE_VULNERABILITY` in Hindi ends at `messages_hi.properties:125`, which carries no trailing backslash, so the remainder on line 126 is read as the key `XML`.

Between them the 6 cut four values short: `OPEN_REDIRECTION_VULNERABILITY_3XX_BASED` in Hindi, where three orphans split one value into four pieces, the same key in Spanish, `XXE_VULNERABILITY` in Hindi, and `SQL_INJECTION_VULNERABILITY` in Spanish. The last of those is the `descriptionLabel` of three controllers, `UnionBasedSQLInjectionVulnerability`, `BlindSQLInjectionVulnerability` and `ErrorBasedSQLInjectionVulnerability`.

**To Reproduce**

```bash
./gradlew bootJar -x test
V=http://localhost:9090/VulnerableApp/allEndPointJson
for L in en hi zh-TW es; do
  printf '%-6s ' "$L"
  curl -s -H "Accept-Language: $L" "$V" \
    | python3 -c 'import sys,json;d=json.load(sys.stdin);print([len(e["Description"]) for e in d if e["Name"]=="XSSWithHtmlTagInjection"][0])'
done
```

Description length in characters, from a build of `8a8d049`:

| Vulnerability | en | hi | zh-TW | es |
|---|---:|---:|---:|---:|
| `XSSWithHtmlTagInjection` | 1199 | 1199 | 1199 | 1372 |
| `XSSInImgTagAttribute` | 1199 | 1199 | 1199 | 1372 |
| `PersistentXSSInHTMLTagVulnerability` | 1199 | 1199 | 1199 | 1372 |
| `XXEVulnerability` | 1155 | 38 | 864 | 1309 |
| `Http3xxStatusCodeBasedInjection` | 1965 | 192 | 1417 | 754 |
| `UnionBasedSQLInjectionVulnerability` | 1454 | 1416 | 1028 | 726 |

For the three XSS rows, hi and zh-TW are not merely the same length as English, they are the same string. Spanish differs, because `messages_es.properties` spells the key correctly. The Hindi text stored under `BILITY` appears in no response. I looked for it across 22 values of the Accept-Language header against both endpoints, 44 responses in all, searching the whole body rather than one field. The Spanish description, which is served, turns up only in the responses for Spanish locales.

`XXEVulnerability` is the control that shows locale selection itself works. It changes in all three languages. What Hindi returns is the whole served value:

```
XML बाहरी इकाई हमला  के लिए एक विधि है
```

38 characters against 1155 in English, which is where `messages_hi.properties:125` stops. The same pairing holds for `Http3xxStatusCodeBasedInjection`: 192 for Hindi against 1965, matching the three orphans at lines 56, 58 and 60, and 754 for Spanish, matching the one at line 60 of that file. The last row is the Spanish `Links` at line 240, and since that label serves three SQL controllers, the same truncated paragraph is what all three of them show.

A shorter translation is not by itself a defect, since a translation may be more concise. What makes these two defects is that each short value pairs with an orphaned key on the line where the value stops.

**Expected behavior**

A request asking for hi or for zh-TW should return the Hindi and Traditional Chinese XSS description that the repository already contains, and the Hindi XXE description should not break off in the middle of its first sentence.

**Additional context**

*How old, and fixed by hand once already.* None of the 9 is recent. `messages_hi.properties` arrived in #439 on 28 July 2023 carrying **15** keys that `messages.properties` did not have. #440 landed 14 minutes later, twenty lines changed in that one file and nothing else, and removed 10 of them. The other 5 are the ones still there today, `BILITY` at the same line 12 it occupied then. The Spanish, Chinese and Bengali ones are likewise present in the commit before #732 landed, and #732 did not touch `messages_zh_TW.properties` or `messages_bn_IN.properties` at all.

That is the same shape as the continuation defect. #326 put the space before the backslash in nine keys by hand in 2021, and #729 found the pattern back across the bundles added since. A by-hand pass fixes the instances and not the class.

*Relationship to the two checks already in the build.* `MessageBundleContinuationTest` from #732 inspects the seam between two physical lines of one entry, so it never sees a value that ended early. `MessageBundleDuplicateKeyTest` from #776 compares a file against itself, so it never compares a file against `messages.properties`. On `8a8d049` both are green while the 9 sit in the tree. Both are one file away from finding them.

*Suggested check.* Every key in a translation bundle must exist in `messages.properties`. Run against `8a8d049` that rule reports these 9 and nothing else. The reverse direction is deliberately not checked: translations are incomplete by a wide margin, from 38 to 107 keys against 938, and that is a normal state for a volunteer translation, not a defect.

*Effect on the pull requests that are open now.* Eight of them touch a translation bundle, so I applied the rule to each at its own head. Five stay green. Three do not, and all three already conflict with master. Two of those are red only because their base predates the key they add, which is in `messages.properties` today, so a rebase clears them. The third adds three keys to `messages_en_US.properties` that exist in no other bundle, which is the defect this issue is about. I can name them if that is useful.

*One trap in the fix.* Restoring a missing backslash has to restore ` \` and not `\`. I tried the plain version first and `MessageBundleContinuationTest` went red on four seams in `messages_hi.properties`, since the Hindi lines end on a letter. Repairing the four truncated values therefore has to put the space before the backslash, which is what #326 and #732 both settled on.

*Which bundle answers.* Rather than assume the fallback, I replaced `XSS_VULNERABILITY` in `messages.properties` with a marker string, and a different marker in `messages_es.properties`, then rebuilt. hi, zh-TW and de all return the `messages.properties` marker while es returns its own, so the fallback for a locale whose bundle lacks the key does land on the default bundle. That holds with the JVM default locale forced to Spanish as well as left alone, so the numbers above do not depend on the host.

*How I measured it.* All 11 bundles were loaded with `java.util.Properties` and each key set compared against the default bundle. The same measurement was written a second time as an independent parser in Python; the two agree on every file and every key name. Both were then checked against the running application, since a key set difference only matters if it changes what is served.

Yes. The text fix is small and I would send the check with it, as a third test beside the two already there. If you would rather have the text alone, that is easy to split out.

vianbas:
Fixes #782.

Nine keys existed only in a translation bundle. Nothing reads them, because no code asks for a key that the default bundle does not define, so three translated descriptions were never served. The same breaks truncated four other values, which stopped mid-sentence.

**Three causes, not one.** That is why this touches four files in slightly different ways.

| File | Line | What was wrong |
|---|---:|---|
| `messages_hi.properties` | 12 | key read `BILITY`, the `XSS_VULNERA` prefix was absent |
| `messages_zh_TW.properties` | 12 | same |
| `messages_bn_IN.properties` | 19 | key was missing `REPLACING_IMG_AND_INPUT_TAG_` |
| `messages_hi.properties` | 55, 57, 59, 125 | trailing continuation backslash missing |
| `messages_es.properties` | 239 | backslash followed by a space, so the line does not continue |
| `messages_es.properties` | 59 | a blank line inside a continued value ended it early |

The last three shapes each turned the next line of prose into a key of its own, which is where keys like `3xx`, `XML`, `Links` and `<br/><br/>` came from.

**Verified two independent ways** that agree: a parser written against the `java.util.Properties` rules, and `java.util.Properties` itself loading the files. Orphan keys go from nine to zero.

The three descriptions that were never served now resolve. The four truncated values are whole again, and land close to the default bundle's length for the same key rather than stopping mid-word:

| Key | before | after | default |
|---|---:|---:|---:|
| `OPEN_REDIRECTION_VULNERABILITY_3XX_BASED` in `hi` | 192 | 1998 | 1965 |
| `XXE_VULNERABILITY` in `hi` | 38 | 1123 | 1155 |
| `OPEN_REDIRECTION_VULNERABILITY_3XX_BASED` in `es` | 754 | 2142 | 1965 |
| `SQL_INJECTION_VULNERABILITY` in `es` | 726 | 1612 | 1454 |

All four now end with the same closing markup the default ends with, translated, instead of ending mid-word.

No translated wording was written or changed. The edits are the key prefixes and the continuation characters only, so the text is the translators' own, just rejoined.

`./gradlew spotlessCheck build` passes with 471 tests, 0 failures, 2 skipped. `MessageBundleContinuationTest` and `MessageBundleDuplicateKeyTest` both pass.

While checking I also looked for the two shapes this class of bug can take without producing an orphan key. Both turned up and neither is a defect, so nothing here changes them: `messages_sv.properties` has a blank line inside a continued value at lines 34 and 259, and `messages.properties` has a backslash followed by a space at line 612. In all three the value was already complete, ending in a full stop at a length comparable to its counterpart.

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected security-related localization keys and formatting in Bengali, Hindi, and Traditional Chinese resources.
  - Fixed Hindi line continuations so translated vulnerability messages display correctly.
  - Removed an unintended text fragment from Hindi and Traditional Chinese security messages.

- **Style**
  - Removed trailing whitespace from the Spanish localization resource.

- **Localization**
  - Preserved existing translated message text while improving consistency across supported languages.


vianbas:
Good catch, and it was a real consequence of this PR rather than a false positive, so thank you.

The split that produced the `BILITY` key had left the other half of the key name inside the value, on line 21, just before the final list item. While the key was `BILITY` nothing ever read that value, so the fragment was invisible. Restoring the key is what made it visible, which means this PR would have shipped `XSS_VULNERA` into the Hindi and Chinese XSS descriptions.

Fixed in `a3f715e`. Both values now end the way the default bundle does, with the closing anchor followed directly by the last list item, verified by loading each file with `java.util.Properties` and comparing.

I also checked the other five values this PR repairs for the same shape, since if one key name was displaced into its own value the others could be too. Scanning each loaded value for a stray run of capitals, and ignoring the deliberate template references such as the one in `XSS_IMAGE_TAG_INJECTION`, all five come back clean. Orphan keys remain at zero and the build still passes with 471 tests.

For completeness I widened the same scan to every value in all eleven bundles, not just the ones this PR touches, in case a key name had been displaced into a value anywhere else. Nothing of this kind turned up. The scan does report a few uppercase runs that are deliberate, such as the cookie name in `SESSION_MANAGEMENT_LEVEL1_HINT_2` and the constant named in `PATH_TRAVERSAL_LEVEL13_HINT1`, and those are correct as they stand.

preetkaran20:
> Good catch, and it was a real consequence of this PR rather than a false positive, so thank you.
> 
> The split that produced the `BILITY` key had left the other half of the key name inside the value, on line 21, just before the final list item. While the key was `BILITY` nothing ever read that value, so the fragment was invisible. Restoring the key is what made it visible, which means this PR would have shipped `XSS_VULNERA` into the Hindi and Chinese XSS descriptions.
> 
> Fixed in `a3f715e`. Both values now end the way the default bundle does, with the closing anchor followed directly by the last list item, verified by loading each file with `java.util.Properties` and comparing.
> 
> I also checked the other five values this PR repairs for the same shape, since if one key name was displaced into its own value the others could be too. Scanning each loaded value for a stray run of capitals, and ignoring the deliberate template references such as the one in `XSS_IMAGE_TAG_INJECTION`, all five come back clean. Orphan keys remain at zero and the build still passes with 471 tests.
> 
> For completeness I widened the same scan to every value in all eleven bundles, not just the ones this PR touches, in case a key name had been displaced into a value anywhere else. Nothing of this kind turned up. The scan does report a few uppercase runs that are deliberate, such as the cookie name in `SESSION_MANAGEMENT_LEVEL1_HINT_2` and the constant named in `PATH_TRAVERSAL_LEVEL13_HINT1`, and those are correct as they stand.

@vianbas can you please make comments bit consize as it is very difficult to read so much text,

preetkaran20 (approved):

PR #619, opened 2026-05-18
Pull Request: Add Phishing-Aware Open Redirect Levels (LEVEL_9, LEVEL_10, LEVEL_11)

StevenTaing2:
## Summary
This PR extends the `Http3xxStatusCodeBasedInjection` vulnerability class with three new difficulty levels demonstrating how open redirect vulnerabilities can be escalated into phishing and credential harvesting attacks, and how such attacks can be mitigated.

- **LEVEL_9** — A baseline implementation using an unvalidated redirect, illustrating how a user can be redirected from a trusted domain to a fake login page which can steal credentials.
- **LEVEL_10** — A partial mitigation of LEVEL_9, where the user is shown a warning page which requires user interaction prior to redirection.
- **LEVEL_11** — A secure redirect implementation that prevents external redirection and details to the user why the redirect attempt was blocked.

## Changes
| Area | Detail |
|---|---|
| `Http3xxStatusCodeBasedInjection.java` | Added LEVEL_9, LEVEL_10, LEVEL_11 endpoint handlers |
| `messages_en_US.properties` | Added i18n strings for the three new levels |
| `static/phishing/fake-login.html` | Simulated phishing login page used as redirect target in LEVEL_9/10 |
| `templates/Phishing/fake-login.html` | Template variant of the fake login page |
| `LEVEL_9/` UI assets | HTML, CSS, and JS for the vulnerable redirect with warning flow |
| `LEVEL_10/` UI assets | HTML, CSS, and JS for the improved interstitial-style redirect |
| `LEVEL_11/` UI assets | HTML, CSS, and JS for the secure redirect and blocked-page interactive UI |
| `Http3xxStatusCodeBasedInjectionTest.java` | Extended unit tests covering all three new levels |

## Test Plan
- [ ] Run `./gradlew test` — all existing and new unit tests pass
- [ ] Start the application and navigate to the Open Redirect vulnerability section
- [ ] Confirm LEVEL_9 redirects to the fake login page
- [ ] Confirm LEVEL_10 interstitial page renders correctly and the redirect works
- [ ] Confirm LEVEL_11 blocks a known phishing URL and displays the interactive blocked page
- [ ] Confirm LEVEL_11 allows a safe redirect URL through normally
- [ ] Verify `fake-login.html` renders correctly as a credential harvesting demonstration

## Summary by CodeRabbit

* **New Features**
[...]

PR #793, opened 2026-09-09
feat: add a secure open redirect level that never takes a URL (#403)

vianbas:
Towards #403.

That issue asks for two or three more secure implementations in the open redirect package. One of them has effectively landed already: #619 added `LEVEL_11`, which is secure. This adds another, and deliberately not a third variant of what is already there.

**Why not another allowlist.** `LEVEL_8` and `LEVEL_11` are both marked secure, but they run the same check, `WHITELISTED_URLS.contains(urlToRedirect)`. They differ only in what happens on failure, a quiet 200 against a 403. So the class demonstrates one secure technique twice. Since the issue exists to give scanners false positive material, a second technique seemed more useful than a third variant of the first.

**What LEVEL_12 does.** The `returnTo` parameter carries a short destination name rather than a URL, and the application maps that name to a destination it already holds. Nothing the caller sends ever reaches the `Location` header, so there is no value to smuggle past a filter. That makes it secure by construction rather than by a check that has to be argued about, which is the property that makes it useful as a false positive: the usual redirect payloads have nothing to act on.

Unknown names are answered with 400 and no `Location` header.

**Verified against a running instance, with a control.** A test that everything rejects proves nothing, so the same harness was pointed at `LEVEL_1`, which is fully open, and it did redirect to an external host. `LEVEL_12` was then given nineteen inputs: the scheme forms, protocol relative, backslash and mixed slash forms, `javascript:`, userinfo before the host, single and double encoded forms, a CRLF attempt, a null byte, case variants, leading and trailing whitespace, a duplicated query parameter, and the two URLs the existing allowlist does accept.

Every `Location` header it emitted was one of its own two mapped destinations. No input produced an attacker chosen target.

One result is worth stating rather than hiding. Sending the name `home`, a null byte, and then a URL answers 302 to `/VulnerableApp/`, because the null byte truncates the parameter and what remains matches a known name. That is not a bypass, since the destination still comes from the map rather than from the request, and it is a fair illustration of the point: even when input handling does something surprising, the worst case here is that a valid name matched.

`./gradlew spotlessCheck build` passes with 474 tests, 0 failures, 2 skipped.

**On scope.** I asked four questions on #403 in August about how far this should go and did not want to sit on the issue indefinitely, so I have taken the narrowest reading. If you would rather it looked different, this is cheap to redirect.

I also considered a host based check that parses the value and compares the host against an allowed set. It is a real pattern and would be good material, but a level marked secure that a scanner is right to flag is worse for your ground truth than no level at all, so I would rather do that one separately and with proper adversarial testing than bundle it here.

## Summary by CodeRabbit

- **New Features**
  - Added a secure redirect demonstration using approved destination names instead of arbitrary URLs.
  - Added an interactive test page with examples of allowed destination names and blocked URL inputs.
  - Added guidance explaining the secure redirect policy.
  - Approved destinations redirect successfully, while unknown names and direct URL payloads are rejected with a bad-request response and no redirect.
  - Added validation covering malicious URL formats, path traversal, and other unsafe inputs.

### DIFF
--- a/src/main/java/org/sasanlabs/service/vulnerability/openRedirect/Http3xxStatusCodeBasedInjection.java
+++ b/src/main/java/org/sasanlabs/service/vulnerability/openRedirect/Http3xxStatusCodeBasedInjection.java
@@ -6,7 +6,10 @@ import java.net.MalformedURLException;
 import java.net.URL;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashMap;
 import java.util.HashSet;
+import java.util.Map;
 import java.util.Set;
 import java.util.function.Function;
 import org.sasanlabs.internal.utility.FrameworkConstants;
@@ -58,6 +61,19 @@ public class Http3xxStatusCodeBasedInjection {
     private static final String RETURN_TO = "returnTo";
     private static final Set<String> WHITELISTED_URLS =
             new HashSet<>(Arrays.asList("/", "/VulnerableApp/"));
+
+    /**
+     * Destinations this level is willing to redirect to, keyed by a short name. The request never
+     * carries a URL, only one of these keys, so there is no attacker controlled value to validate.
+     */
+    private static final Map<String, String> REDIRECT_DESTINATIONS =
+            Collections.unmodifiableMap(
+                    new HashMap<String, String>() {
+                        {
+                            put("home", "/VulnerableApp/");
+                            put("root", "/");
+                        }
+                    });
 
     private ResponseEntity<?> getURLRedirectionResponseEntity(
             String urlToRedirect, Function<String, Boolean> validator) {
@@ -381,5 +397,27 @@ public class Http3xxStatusCodeBasedInjection {
 
         return ResponseEntity.status(HttpStatus.FORBIDDEN)
                 .body("Redirect blocked: untrusted redirect target");
+    }
+
+    @AttackVector(
+            vulnerabilityExposed = {VulnerabilityType.OPEN_REDIRECT_3XX_STATUS_CODE},
+            description = "OPEN_REDIRECT_PHISHING_SECURE_INDIRECT_REFERENCE")
+    @VulnerableAppRequestMapping(
+            value = LevelConstants.LEVEL_12,
+            variant = Variant.SECURE,
+            htmlTemplate = "LEVEL_12/Http3xxStatusCodeBasedInjection")
+    public ResponseEntity<String> getVulnerablePayloadLevel12(
+            @RequestParam(RETURN_TO) String destinationKey) {
+
+        String destination = REDIRECT_DESTINATIONS.get(destinationKey);
+
+        if (destination == null) {
+            return ResponseEntity.status(HttpStatus.BAD_REQUEST)
+                    .body("Redirect blocked: unknown destination");
+        }
+
+        return ResponseEntity.status(HttpStatus.FOUND)
+                .header(LOCATION_HEADER_KEY, destination)
+                .build();
     }
 }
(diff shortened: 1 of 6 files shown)


igorzelaya-io (approved):
Reviewed current head 37db32d against #403 and the existing LEVEL_8/LEVEL_11 behavior. The indirect-reference design keeps caller-controlled values out of the Location header, rejects unknown keys without a redirect, and adds adversarial coverage for URL-shaped inputs. I also ran Http3xxStatusCodeBasedInjectionTest locally; it passed. Looks good to me.

src/main/java/org/sasanlabs/service/vulnerability/openRedirect/Http3xxStatusCodeBasedInjection.java
```diff
@@ -382,4 +398,26 @@ public ResponseEntity<String> getVulnerablePayloadLevel11(
...
+    @VulnerableAppRequestMapping(
+            value = LevelConstants.LEVEL_12,
+            variant = Variant.SECURE,
+            htmlTemplate = "LEVEL_12/Http3xxStatusCodeBasedInjection")
+    public ResponseEntity<String> getVulnerablePayloadLevel12(
+            @RequestParam(RETURN_TO) String destinationKey) {
+
+        String destination = REDIRECT_DESTINATIONS.get(destinationKey);
+
+        if (destination == null) {
```
preetkaran20: I think better to use containskey function instead of null comparison.

preetkaran20:
Can you please share the UI/Ux for the level

src/main/resources/static/templates/Http3xxStatusCodeBasedInjection/LEVEL_12/Http3xxStatusCodeBasedInjection.js
```diff
@@ -0,0 +1,54 @@
...
+  resultBox.style.display = "block";
+}
+
+function testLevel12Redirect(value) {
+  const encoded = encodeURIComponent(value);
+  const url =
+    "/VulnerableApp/Http3xxStatusCodeBasedInjection/LEVEL_12?returnTo=" +
+    encoded;
+
+  fetch(url, {
```
preetkaran20: Please use frameworks method instead of directly calling backend. Look at other levels JavaScript.

src/main/resources/static/templates/Http3xxStatusCodeBasedInjection/LEVEL_12/Http3xxStatusCodeBasedInjection.js
```diff
@@ -0,0 +1,54 @@
...
+    method: "GET",
+    redirect: "follow",
+  })
+    .then(function (response) {
+      if (response.redirected) {
```
preetkaran20: I think if we are returning 301 or 302 it should auto redirect, can you please look at other levels ?

src/main/java/org/sasanlabs/service/vulnerability/openRedirect/Http3xxStatusCodeBasedInjection.java
vianbas: Done in `6bd86e4`:

```java
if (!REDIRECT_DESTINATIONS.containsKey(destinationKey)) {
```

The destination is then read with `get`.

src/main/resources/static/templates/Http3xxStatusCodeBasedInjection/LEVEL_12/Http3xxStatusCodeBasedInjection.js
vianbas: Done in `6bd86e4`. The links are now built with `getUrlForVulnerabilityLevel()`, as in `LEVEL_1`, `LEVEL_9` and `LEVEL_10`, and the direct `fetch` is gone.

I did not use `doGetAjaxCall`. `XMLHttpRequest` follows a 302 without moving the page, and the framework's response handler shows an alert on a 400.

src/main/resources/static/templates/Http3xxStatusCodeBasedInjection/LEVEL_12/Http3xxStatusCodeBasedInjection.js
vianbas: Done in `6bd86e4`. Each link opens in a new tab, as in `LEVEL_1` and `LEVEL_9`, so the browser follows the 302 itself. `home` lands on `/VulnerableApp/`, and an unknown name shows the server's 400 message.

preetkaran20 (approved):

preetkaran20:
@vianbas Thanks a lot for the PR

